PR #86, opened 2021-10-28
Add a real Malloy Runtime API

christopherswenson:
[Draft]

- Also add a demo CLI utility that shows how it can be used


packages/malloy/src/malloy.ts
```diff
@@ -691,6 +691,11 @@ export class Runtime {
     });
   }
 
+  // used internally by the testing framwork.
+  public _makeModelFromModelDef(_modelDef: ModelDef): ModelRuntimeRequest {
```
lloydtabb, line 695: @christopherswenson can you implement this?  Uncommenting out the calls in the describe section of  expr.spec.ts  should make this work.

packages/malloy/src/malloy.ts
```diff
@@ -785,6 +790,10 @@ export class ModelRuntimeRequest extends RuntimeRequest<Model> {
     });
   }
 
+  _makeQueryFromQueryDef(_query: InternalQuery): PreparedQueryRuntimeRequest {
```
lloydtabb, line 793: @christopherswenson and implement this one too?  

thanks.

PR #87, opened 2021-11-01
move our packages to use "@malloy-lang" scope (which we own on npm)

bporterfield:
There is a concern that perhaps someone would think we have published npm packages (we have not) and if some nefarious actor grabs a package on npm with a name we're using, some other developer might intend to install a malloy package and accidentally install the bad package.

This change resolves that concern by scoping the packages to `@malloy-lang/{package-name}` - an organization name in npm that we own - with the exception of the VScode extension. VSCode extensions have chosen (possibly poorly?) to utilize `package.json` as the manifest file for an extension - but they do not support all possible values for the `name` property and as such [do not support scoped names](https://github.com/microsoft/vscode-vsce/issues/186).

While it is not intended to be installed via npm, we have reserved the `malloy-vscode` name on npm by [publishing the vs-code package to it](https://www.npmjs.com/package/malloy-vscode).

### DIFF
--- a/docs/_scripts/build_docs/index.ts
+++ b/docs/_scripts/build_docs/index.ts
@@ -13,8 +13,8 @@
 
 import path from "path";
 import fs from "fs";
-import { Malloy } from "malloy";
-import { BigQueryConnection } from "malloy-db-bigquery";
+import { Malloy } from "@malloy-lang/malloy";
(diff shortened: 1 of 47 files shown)


christopherswenson:
Looks great 👍

christopherswenson (approved):

PR #119, opened 2021-12-28
Initial cut as making postgres testing work

lloydtabb:


lloydtabb:
I'm going to merge this as it shouldn't effect anything at this point and will let Mac users test their postgres connection.

PR #121, opened 2021-12-31
Make a runtime that can handle multiple dialects.

lloydtabb:
Make a test first.

### DIFF
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -33,6 +33,7 @@ import {
   NamedStructDefs,
   Connection,
 } from "@malloy-lang/malloy";
+import { parseTableName } from "@malloy-lang/malloy/src/malloy";
 
 export interface BigQueryManagerOptions {
   credentials?: {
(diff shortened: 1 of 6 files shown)


lloydtabb:
@christopherswenson this test hangs, not sure why.  Maybe I'm using things wrongly.  Can you take a look?

lloydtabb:
Ok, I think I understand the problem.  @mtoy-googly-moogly need your advice here....

We are getting stuck in an infinite translator loop because the code that gets a structdef from a table name is returning a different name then you asked for...

`explore 'bigquery:malloytest.flights' | reduce c is count()`

Will hang because the structdef returned has table who's `name: 'malloytest.flights'` and `connection: bigquery`.  Since the parser didn't get what it asked for, the translate loop goes again and attempts to translate.  Because there is no error but it still isn't returning what is asked for we loop infinitely.

I'm not sure how we want to fix this.  We can either make the parser know about connections or the table fetcher always return the full URL to the table including how we asked for it.  The backend code will still have to know how to remove the connection name.  

Alternatively, we could add the connection space to the language...

`explore table('malloytest.flights', 'bigquery') | reduce c is count()`

There is something nice about making the connection not part of the URL...

Thoughts?

mtoy-googly-moogly:
Hard to comment on the solution because I don't understand how the proposed fix would prevent the problem, but then I haven't sat in the debugger. Let me know if you need me to do that.

I'm pretty in love with "we have a URI for table names", in general, so I'd like to know more before I agree to throw that in the toilet.

lloydtabb:
Ok.  URLs work.  We never modify the URL and the connections expect URLs when referencing tables.  

Before today, the translator would break a URL into a connection and a tableName (really a table path).  Now the translator only cares about the connection name.  The database schema interface expects URLs and returns them modified.

lloydtabb:
Resolution:

*tableURL* - path to the table optionally including the connection  
*tablePath* - the part following the colon in a table URL

Names in structDefs are tableURLs.  StructDef.source.tablePath is the path part and is used when generating code.  

tableURLs are used for cacheKeys (which is the source of this problem) and structDef indentifiers.

I needed to add a notion of defaultProject to BigQuery connections to make tests work reliably.  We probably need a notion of default dataset too.  All this needs to be repackaged into connection parameters somehow..

PR #122, opened 2022-01-01
Beta Language

mtoy-googly-moogly:
The beta language branch is now roughly in sync with main. Here is the TODO before this could land

- [ ] re-write all sample models
- [ ] Get existing tests to pass
-  [ ] Replace all the tests in lang which were deleted
- [x] Restore the IDE integration, which was gutted and disabled.
- [ ] Update docs


packages/malloy-db-test/src/expr.spec.ts
```diff
@@ -586,29 +586,6 @@ expressionModels.forEach((expressionModel, databaseName) => {
     expect(result.data.path(0, "flight_count").value).toBe(199726);
     expect(result.data.path(0, "popular_name").value).toBe("Isabella");
   });
-
-  // Issue: #151
-  it(`unknonwn dialect  - ${databaseName}`, async () => {
```
lloydtabb: Issue #151 @mtoy-googly-moogly

lloydtabb: Not sure how to link this for you in the bug...

PR #125, opened 2022-01-03
Change the Run buttons to Run | Render

lloydtabb:
Experimental, but I think we should just have two lenses over queries Run and Render.   Since we are a data tool, the Run button should show the raw data.  The Render button passes these results to the renderer.  I think this will make Malloy more comprehensible.


lloydtabb:
I want to live with the buttons like this for a while and see how well it demos...

PR #126, opened 2022-01-04
We needed to pay attention to how the field was generated,

lloydtabb:
not its current state.


lloydtabb:
@christopherswenson , API change here.  Please see of the function name makes sense to you.

lloydtabb:
We weren't rendering the third component of a bar chart if it was a measure.  This fixes that bug.

christopherswenson:
Added some changes. Removed `isDimensional` and `isMeasureLike` entirely for now. They were not implemented correctly, and were confusing. We can add them back if we need them.

PR #144, opened 2022-01-11
Fix refined queries which contain field expressions

christopherswenson:
@mtoy-googly-moogly fixed an issue where no fields would show up when a query was refined, e.g.

```
query: flights -> airport_dashboard {? origin.code: 'SJC' }
```

However, if the query contained any field definitions rather than field references, the resulting query would not correctly re-define those fields. 

This PR should fix that.

Two sample models (`iowa.malloy` and `flights_beta.malloy`) are now compiling again with this change.


mtoy-googly-moogly (approved):

mtoy-googly-moogly:
nice, thanks

PR #188, opened 2022-01-17
Quick set of edits to the quick start page

mtoy-googly-moogly:
I am not a good "reviewer for mistakes" person because my brain shuts off after about 15 minutes. "Hey michael, whatever this boring thing you are doing is, let me know when you are done and I will power back up".

So here are 15 minutes of changes I think the quick start page needs. Some are typos and some are edits for clarity.

Made the a PR because I don't know what the right thing to do is when I have doc changes.


mtoy-googly-moogly:
Also, editing the document seems to produce a tree which doesn't build, so HELP!

PR #194, opened 2022-01-18
fix ga_sessions error

anikaks:
half of fix for https://github.com/looker-open-source/malloy/issues/193


anikaks:
ack don't use this

mtoy-googly-moogly:
Shouldn't the old code with the "is" also have complied and worked fine?

anikaks:
The malloy code is valid but my understanding is that the docs do not currently like named queries like that to be runnable

PR #206, opened 2022-01-19
Joins joins joins!

mtoy-googly-moogly:
Still thinking about curating, and not presenting the algebraic totality of all join types. 

#205 is the discussion about syntax.


mtoy-googly-moogly:
* filtered cross joins are broken
* some references in existing code i did not fix, search for `join:` and fix them
* `ON` joins throw because there is a "check for primary key" not a "check if i am joined with known relationship"

expecting @lloydtabb to fix all these by time i get back to my keyboard, i'm off to shred

PR #211, opened 2022-01-20
Docs build should exit with code 1 if any snippet had errors

christopherswenson:


christopherswenson:
The last commit is actually a great example of this working as intended:
Screen Shot 2022-01-20 at 10 21 21 AM

christopherswenson:
I fixed that in another branch. Will merge in and we should see the build succeed this time.

PR #221, opened 2022-01-21
avoid stack overflow when field def references its name

### LINKED ISSUE #195: rename creates a stack overflow.
```
explore: words_base is table('malloy-data.wordle.words'){
  where: length(word) = 5 and  word ~ r'^[a-z]....$'

  join: numbers is table('malloy-data.malloytest.numbers'){dimension: a is 'a'; primary_key: a} on a
  dimension: a is 'a'

  rename: lower_word is word
  dimension: word is UPPER(word)

  query: words_and_position is {
    group_by: word is UPPER(word)
    nest: letters is {
      where: substr(word, numbers.num, 1) != ''
      order_by: 2
      group_by: [
        letter is substr(word, numbers.num, 1)
        position is numbers.num

      ]
    }
  }
}
```

mtoy-googly-moogly:
#195 brought this to light, if you define a field, and reference that field in the definition, it makes a stack overflow.

i fixed that in the first couple of commits here  ...

But this fix is wrong because

```
...
   group_by: distance is distance / 100
```

is legal because the expression is in a different namespace than the one being created ...

however that also would be a problem for queries and joins and explores ... kind of need a more general solution


mtoy-googly-moogly:
I am kicking this down the hill a little because there is a scary refactor, and I want to collect myself more before I do that.

1) `dimension: x is x+1` is illegal
2) `group_by: x is x+1` is legal

This fix pretty narrowly does that and tests for those two things.

PR #223, opened 2022-01-22
Update Join Docs.

lloydtabb:


lloydtabb:
I added some notes to the doc about things I think we should cover.  I have a hard time with reference docs.  @anikaks can you take the lead on this.  @mtoy-googly-moogly can you advise/review?

PR #248, opened 2022-01-25
quickstart edits

anikaks:


anikaks:
@lloydtabb my cross join example doesn't work; let's look at it tomorrow with fresh brains

anikaks:
also--I used a hypothetical ecommerce schema rather than our actual one for the joins examples to have an orders table in between order_items and users; can try to come up with a different one (or point to order_facts) if we feel this will confuse folks more than help them

anikaks:
@lloydtabb could you take a look tomorrow or whenever you have a chance? Need to get the cross join working still

anikaks:
redid / pushed these changes separately.

Issue #255, opened 2022-01-25
expand support for postgres datatypes


anikaks:
boolean type for sure not currently supported. @lloydtabb is on this but documenting for posterity.

jonathanswenson:
Added some types here: #259 

but there are several others that I didn’t know what to do with that are referenced in the pg_catalog schema (referenced in #251) 

```
bytea
ARRAY (in caps for some reason)
pg_ndistinct
pg_mcv_list
pg_dependencies
interval
inet
pg_node_tree
bytea
pg_lsn
anyarray
regproc
regtype
```

wrote some simple explores that let me play with them that might be useful (the “missing” type is a comment at the end of the line): 
Any commented out line was NOT working as of #259 but the others were at least possible to query / load the schema for a preview. 
```
explore: pg_tables is table('pg:pg_catalog.pg_tables') {}
explore: pg_authid is table('pg:pg_catalog.pg_authid') {}
-- explore: pg_statistic is table('pg:pg_catalog.pg_statistic') {} // array
--  explore: pg_type is table('pg:pg_catalog.pg_type') {} // regproc
--  explore: pg_foreign_table is table('pg:pg_catalog.pg_foreign_table') {} // array
--  explore: pg_statistic_ext_data is table('pg:pg_catalog.pg_statistic_ext_data') {} // pg_ndistinct
--  explore: pg_user_mapping is table('pg:pg_catalog.pg_user_mapping') {} // array
--  explore: pg_subscription is table('pg:pg_catalog.pg_subscription') {} // array
--  explore: pg_attribute is table('pg:pg_catalog.pg_attribute') {} // array
--  explore: pg_proc is table('pg:pg_catalog.pg_proc') {} // regproc
--  explore: pg_class is table('pg:pg_catalog.pg_class') {} // array
--  explore: pg_attrdef is table('pg:pg_catalog.pg_attrdef') {} // pg_node_tree
--  explore: pg_constraint is table('pg:pg_catalog.pg_constraint') {} // regproc
 explore: pg_inherits is table('pg:pg_catalog.pg_inherits') {} // empty for me
--  explore: pg_index is table('pg:pg_catalog.pg_index') {} // array
--  explore: pg_operator is table('pg:pg_catalog.pg_operator') {} // regproc
 explore: pg_opfamily is table('pg:pg_catalog.pg_opfamily') {}
 explore: pg_opclass is table('pg:pg_catalog.pg_opclass') {}
[...]

Issue #251, opened 2022-01-25
Path not found inventory_id (vscode)


swuecho:
```malloy
query: table('db_name.public.table_name') -> {
  project: [ inventory_id ]
  where: asn_id=11371
}
```

It's great to have compile time check, but the error msg seems not right. I double checked that the 'inventory_id' exists.

anikaks:
Hi @swuecho - your syntax looks right to me at first look. Just to check, have you tried any other fields here, and did they work? has anything else successfully run for you yet?

swuecho:
```
query: table('titan.pg_catalog.pg_tables') -> {
  project: [ tablename ]
  where: tableowner='titan'
}
```
```sh
 select * from titan.pg_catalog.pg_tables where tableowner = 'titan';
```

I tried above malloy query. no luck too. The sql works fine. I can not run anything with postgres yet.

anikaks:
Thanks @swuecho -- what is the datatype  of that field in postgres?

another thing to try - can you try making it an explore and see if it comes up in the schema view? using the example you provided, that would look like:
```malloy
explore: table_name is table('db_name.public.table_name') {}
```
 you can find the schema viewer in the malloy tab on the left. You may need to click into the open file to see it.
<img width="940" alt="Screen Shot 2022-01-24 at 7 45 00 PM" src="https://user-images.githubusercontent.com/7178946/150907182-317f77d3-4aac-4664-9712-695ca2a52aa3.png">

swuecho:
![image](https://user-images.githubusercontent.com/666683/150909982-5633d659-034a-432a-9d82-f66f557026a3.png)

❯ node -v
v17.2.0

anikaks:
@swuecho I'm also hitting a bump querying the pg_catalog tables; looking into this.

In the mean time, was the other query you were trying to run also against pg_catalog? I am wondering if that was a different issue (and hopefully one we can solve quickly!)

swuecho:
If I left the db prefix, `unknown postgres type boolean malloy` will reported. 

and seems wherever the problem is, the error message is always on the first line.

anikaks:
The boolean type is currently not supported in postgres; we're working to get that addressed asap. Lots of booleans in the `pg_catalog` tables which seems to be the cause of our trouble here.

swuecho:
> The boolean type is currently not supported in postgres; we're working to get that addressed asap. Lots of booleans in the `pg_catalog` tables which seems to be the cause of our trouble here.

[...]

PR #259, opened 2022-01-25
Add some postgres types to allow for querying a few of the pg_catalog tables

jonathanswenson:
Fixes part of #251 / #255 

There are two gotchas here:
+ Like the original author of this type system, I didn not figure out how timestamps work in postgres -- I actually think its impossible
+ Something is causing the information schema function to reutrn `”char”` instead of `char` (like all the rest of the types).

Notably in psql running a similar command as is done [here](https://github.com/looker-open-source/malloy/blob/main/packages/malloy-db-postgres/src/postgres_connection.ts#L142-L144)  it returns the `”char”`
```
test=# SELECT table_schema, table_name, column_name, data_type FROM information_schema.columns WHERE table_schema = 'pg_catalog' and table_name = 'pg_attribute'  order by data_type  limit 20;
 table_schema |  table_name  |  column_name   | data_type 
--------------+--------------+----------------+-----------
 pg_catalog   | pg_attribute | attgenerated   | "char"
 pg_catalog   | pg_attribute | attidentity    | "char"
...
 pg_catalog   | pg_attribute | attacl         | ARRAY
...
 pg_catalog   | pg_attribute | attmissingval  | anyarray
 pg_catalog   | pg_attribute | attnotnull     | boolean
```

There are a variety of pg_catalog tables that are now possible to use, but many of the others are inaccessible due to lack of support for:
```
bytea
ARRAY (in caps for some reason)
pg_ndistinct
pg_mcv_list
pg_dependencies
interval
inet
pg_node_tree
bytea
pg_lsn
anyarray
regproc
regtype
```

Unsure if some of these should ever be supported as some of them are internal, sensitive columns used for things like query optimization statistics.

### DIFF
--- a/packages/malloy-db-postgres/src/postgres_connection.ts
+++ b/packages/malloy-db-postgres/src/postgres_connection.ts
@@ -32,6 +32,14 @@ const postgresToMalloyTypes: { [key: string]: AtomicFieldType } = {
   bigint: "number",
   "double precision": "number",
   "timestamp without time zone": "timestamp", // maybe not right
+  "timestamp with time zone": "timestamp", // also maybe not right?
+  boolean: "boolean",
+  smallint: "number",
+  "\"char\"": "string", // This has to be in quotes because the DB sends it back like this from the information_schema query
+  numeric: "number",
+  real: "number",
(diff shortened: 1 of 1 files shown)


jonathanswenson:
closing in favor of #262

PR #265, opened 2022-01-25
Add a checkbox for whether you want to use a service account key

christopherswenson:
@anikaks Is this helpful? Let me know if you want to chat about other ideas here

Screen Shot 2022-01-25 at 9 32 53 AM


anikaks:
hmm so my thinking with https://github.com/looker-open-source/malloy/issues/252 was that it's not obvious you can (and prob should) leave everything blank if you've set up your BQ connection with oauth, somehow I feel like this isn't clarifying it. what about something like "Connect with service account (unless checked, will attempt to connect with oauth)" or something? IT's just not obvious how you choose what's going on here.

anikaks:
one other thought--should we have a hover "?" or something that "name" is whatever you want it to be and is just how you'll reference this connection in models?

anikaks:
@christopherswenson any thoughts on the above changes so we can clear out this PR? :)

bporterfield:
closing b/c of inactivity

PR #269, opened 2022-01-25
When saving connections, update workspace config only if already set. Always update user config.

christopherswenson:
Fixes https://github.com/looker-open-source/malloy/issues/256

In the future we can give the user more options here. But for now, this is the least likely to cause problems. When you click "Save" in the connections editor:
* If `malloy.connections` is unset, or is only set in user config, user config is updated
* If `malloy.connections` is set in workspace config, both user config and workspace config are updated


bporterfield:
👍

PR #275, opened 2022-01-26
Show HTML/SQL/JSON tabs in query result panel

christopherswenson:
Removes the "Render" code lens, and instead add HTML / SQL / JSON toggles in the webview. Defaults to HTML. SQL and JSON are currently highlighted the same way as in the docs, and does not respect the user's VSCode theme. This is because theme colors aren't passed into the webview nicely. See [here](https://stackoverflow.com/questions/55877879/vs-code-webview-css-variable-names-for-used-theme-tokencolors) and [here](https://github.com/Microsoft/vscode/issues/56356). A possible solution is to use [shiki](https://shiki.matsu.io/) which can highlight according to TextMate themes.

Closes https://github.com/looker-open-source/malloy/issues/273

Screen Shot 2022-01-26 at 11 36 24 AM
Screen Shot 2022-01-26 at 11 36 28 AM
Screen Shot 2022-01-26 at 11 36 31 AM

### DIFF
--- a/docs/_scripts/build_docs/run_code.ts
+++ b/docs/_scripts/build_docs/run_code.ts
@@ -17,7 +17,7 @@
  * that depend on that model. If `--watch` is enabled, changes to a model file
  * will cause relevant documents to recompile.
  */
-import { DataStyles, HTMLView, JSONView } from "@malloydata/render";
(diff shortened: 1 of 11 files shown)


swuecho:
Can not wait to try out! Thanks.

PR #281, opened 2022-01-26
new connection instructions page and shortened readme

anikaks:
I made a dedicated connection instructions page, and then removed them from the readme. Also moved to a list instead of separate headers for Install instructions to streamline the readme. 

I also renamed a couple headers in the docs and added connection instructions under "Getting Started" - open ot moving this or adding an admin/VSCode Plugin section.


anikaks:
Note: we need to do a fresh docs build when this goes out as the connection instructions now only exist in docs.

anikaks:
I was also thinking maybe we should move the Quickstart to the Getting Started Section so that people find it more easily. Any concerns ? CC @lloydtabb

bporterfield (approved):
👍

PR #289, opened 2022-01-28
Make all XXX_TRUNC and extract function use dialects

lloydtabb:


packages/malloy/src/malloy.ts
```diff
@@ -2183,7 +2183,11 @@ class DataTimestamp extends ScalarData<Date> {
 
   public get value(): Date {
     // TODO properly map the data from BQ/Postgres types
-    return new Date((this._value as unknown as { value: string }).value);
```
lloydtabb, line 2186: @christopherswenson ^^

PR #295, opened 2022-01-29
Add sql: command to model

mtoy-googly-moogly:
Parser support is in, we will CERTAINLY be changing the syntax, consider this a placeholder.

Current syntax:

```
// All sql strings start with || and end with ;;
// There is no quoting, waiting to see how it is used before I go all in

// This would have a run button ..
sql:
  || DROP TABLE myComputation ;;
  || CREATE TABLE myComputation AS
      SELECT
        a,
        b
      FROM someWhere
    ;;
    on "connectionName"

// This would define a name usable as an explore ...
sql: myComputation is
  || SELECT a,b FROM someWhere ;;
  on "connectionName"
```

The parser works, and generates "need the Schema for this construct" response if you write these and use it in an explore. Lots to do before it works:

- [x] What are going to do about connections
- [ ] What does a StructDef with SQL in it look like
- [ ] What is the call to ask the database to fill in that information.
- [ ] Named SQL statements stored in model.
- [ ] Un-named SQL statements collected and available in translator
- [ ] IDE needs hover buttons and run actions


anikaks:
👀  I'm intrigued... is this what it sounds like??

christopherswenson:
I'll express my vote for syntax that begins with `explore: thing_from_sql is ...`. Not sure what the `...` looks like.

mtoy-googly-moogly:
@christopherswenson the syntax you mention would be something like 

```
explore: anExploreBasedOnSQL is from_sql(symbolDefinedByAnSQLStatement) {
  ...
}
```

christopherswenson:
Okay, so many thoughts.

I strongly disagree with making SQL blocks be in a different namespace. That means you can write

```malloy
sql: flights is ...
explore: flights is ...
```

And then if you refer to `flights` later, it's not obvious which one you mean.

Can we allow also writing the following?

```malloy
explore: sql_based_explore is from_sql(||
  SELECT * FROM WHATEVER
  ;;
  on "some-connection"
)
```

Putting `sql:` at the top level makes me also want to put

```malloy
string: MY_STR_CONSTANT is 'foo'
number: MY_NUMBER_CONSTANT is 42
timestamp: TOMORROW is now + 1 day
```

PR #298, opened 2022-01-30
Typescript errors even when files are closed in VS Code

bporterfield:
* Update typescript
* Update eslint
* Add ts config to not complain about error in catch being unknown
* Add task to watch all TS files, even closed ones, instead of waiting til a file is opened or compilation run to complain
* Add tsconfig.type-check-only for running background watcher task without emiting anything


christopherswenson:
Might be worth writing down somewhere in the README that you may need to allow automatic tasks to run.

> If this still not work on project folder open try Ctrl+shift+P and Tasks: Manage Automatic Tasks in Folder and choose "Allow Automatic Tasks in folder" on main project folder or running folder.

PR #332, opened 2022-02-03
Finish support for sql: statement

mtoy-googly-moogly:
Parser support is in, and merged to main. Many final implementation details remain.

Current syntax:

```
// All sql strings start with || and end with ;;
// There is no quoting, waiting to see how it is used before I go all in

// This would have a run button ..
sql:
  || DROP TABLE myComputation ;;
  || CREATE TABLE myComputation AS
      SELECT
        a,
        b
      FROM someWhere
    ;;
    on "connectionName"

// This also have a run button, would define a name usable as an explore ...
sql: myComputation is
  || SELECT a,b FROM someWhere ;;
  on "connectionName"

explore: exploreMy is from_sql(myComputation) { ... }
```

The parser works, and generates "need the Schema for this construct" response if you write these and use it in an explore. Lots to do before it works:

- [x] What are going to do about connections
- [x] What does a StructDef with SQL in it look like
- [x] What is the call to ask the database to fill in that information.
- [x] Named SQL statements stored in model.
- [x] Un-named SQL statements collected and available in translator
- [x] IDE needs hover buttons and run actions
- [ ] Errors from SQL schema fetch are reported back to the compiler
- [ ] Decide if SQL blocks in `from_sql` are allowed to have multiple statements


mtoy-googly-moogly:
This sort of works ... all the "fetch stuff" api's that the translator loop calls need to have a way to return errors instead of throwing exceptions ... or the call to the fetcher needs to catch the exceptions and grab the text from the exception and return it to the parser.

I'm OK with merging and doing the error handling work this week, in a different branch, or waiting.

mtoy-googly-moogly:
I added a checkbox to make a call on if multiple line SQL blocks are OK for explores or not, if they are not, I want code changes before we merge.

PR #333, opened 2022-02-03
Implement basic completions for properties

christopherswenson:
This adds VSCode completions for model, explore, and query properties. 
Screen Shot 2022-02-03 at 3 50 03 PM

Model: `explore` and `query`
Explore: `dimension`, `measure`, `join_one`,  `join_many`,  `join_cross`, `where`, `primary_key`, `rename`, `accept`, `except`, and `query`
Query:   `group_by`,  `project`,   `index`, `aggregate`,  `top`,   `limit`, `order_by`, `where`, `having`, and `nest`

Since there's only one kind of completion in existence at the moment, this PR makes no attempt to lay out a grand vision for how completions will categorized and mapped in the code server. 

I anticipate that future completions will need access to the full parse in addition to the AST, and therefore the `completionsStep` will likely end up depending on the `translateStep`.


packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -505,6 +514,37 @@ class MetadataStep implements TranslationStep {
   }
 }
 
+class CompletionsStep implements TranslationStep {
```
christopherswenson, line 517: @mtoy-googly-moogly Can you take a look at this and tell me whether this is the way you'd like me to do this? Tried to fit into your `TranslationStep` world as best as possible, though what I've done is a little like fitting an octagon into a circle — it works, but it feels a little funny doing it. Specifically, this is a request that could be repeated with different inputs (though in practice it's not, a new parse is created for each completions request currently).

packages/malloy-vscode/src/server/completions/completion_docs.ts
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ */
+
+const DOCS_ROOT = "https://looker-open-source.github.io/malloy/documentation";
+
+export const COMPLETION_DOCS: {
```
christopherswenson, line 16: This is by no means intended to be the final form of this mini docs segment. Among other things, it would be nice if these snippets were testable, and if the links were guaranteed to exist in the documentation. However, it might be okay to not overthink this too much for the time being. 

@anikaks Would you mind looking through this and seeing if there's anything you would change here?

packages/malloy/src/lang/parse-malloy.ts
mtoy-googly-moogly: We talked in chat, but for the record 👍  -- You did it exactly as I would hope someone would coming into this new design. Nice work, glad it wasn't too much of a bother.

mtoy-googly-moogly: also ...OMG COMPLETIONS 💯

packages/malloy-vscode/src/server/completions/completion_docs.ts
```diff
@@ -0,0 +1,269 @@
+/*
+ * Copyright 2021 Google LLC
```
mtoy-googly-moogly, line 2: Isn't it 2022 now?

PR #345, opened 2022-02-04
Syntax highlight in IDE, and fetch schemas for queries in BQ

christopherswenson:


mtoy-googly-moogly:
I merged this into sql-statement

PR #349, opened 2022-02-07
More progress on SQL blocks

christopherswenson:
Summary:
* Add ability to run SQL Blocks
* Disable semantic highlighting of Malloy code in VSCode editor and replace with TextMate highlighting
* Use Shiki to highlight docs using the same TextMate grammar
* Fix a concurrency issue with BQ schema fetching where `projectId` could become invalid temporarily
* Replace `SchemaReader` and `SQLRunner` with `InfoConnection` and `Connection extends InfoConnection`. 

## Running SQL Blocks
Adds the appropriate VSCode extension commands to run SQL blocks, and the appropriate additions to the `Runtime` API to enable this.

Specifically, `loadSQLBlockByName`, `loadSQLBlockByIndex`, and other related methods, as well as a `SQLBlockMaterializer` class similar to `PreparedResultMaterializer`.

## TextMate Highlighting
I've been putting this off for some time, because it has seemed "purer" to use the real Malloy grammar to do semantic highlighting. However, semantic highlighting through the language server has some severe limitations (e.g. no ability to specify embedded language highlighting for SQL blocks, less control over the way things highlight, etc.). Furthermore, using TextMate grammars is simply the way highlighting "is done" in VSCode today. Better to fit in than to fight it in this case.

The semantic highlighting has been _disabled_, not removed. This is because I anticipate use cases for semantic highlighting on top of the TextMate highlighting (e.g. highlighting explore references differently than field references). Having this plumbing still around will be helpful when we go to implement such features in the future.

## Shiki Highlighting

Shiki can highlight code according to TextMate grammars, so this replaces the old Prism highlighting. This keeps the number of grammars we need to maintain at 2, because we add the TextMate grammar but remove the Prism one. 

This also allows us to highlight according to VSCode themes. We are now highlighting according to the actual Light+ theme, rather than something close to it. This also gives us the ability to add a dark mode using the Dark+ theme for instance.

Highlighting in the WebView for the SQL result tab still uses Prism, due to some technical snags I ran into when trying to replace it (problems fetching the theme file inside of the iFrame). That is not high priority, so I'm leaving it alone for now. It should be possible to either do the highlighting in the extension and pass the HTML to the extension or to enable Shiki to fetch the resources it needs in the iFrame.

## Concurrency Issue Fix

The BQ API for fetching table metadata does not seem to have a way to specify a different project Id than the one set on the `BigQuery` instance. We currently use a hack where we set the `BigQuery.projectId` to that of the table we need to fetch, and then set it back after initiating the request. This introduces a concurrency problem, because it is possible for a query to attempt to run while we `await` that request and before we set the project Id back to what it's supposed to be.

I have fixed this issue while keeping the hack: we just need to move the resetting of the project Id to before the `await`, while still keeping it after the creation of the metadata request.

This looks like:
```ts
if (projectId) this.bigQuery.projectId = projectId;
const table = this.bigQuery.dataset(datasetNamePart).table(tableNamePart);
const metadataPromise = table.getMetadata();
this.bigQuery.projectId = this.projectId;
const [metadata] = await metadataPromise;
```

## Connection and Info Connections

The `SQLRunner` and `SchemaFetcher` API was kind of confusing, especially now that a `SQLRunner` sometimes needs to fetch schemas (e.g. when running a SQL block). So now we have a "light" connection, an `InfoConnection`, which is capable of getting schemas, but not running queries, and a `Connection`, which extends `InfoConnection`, and adds the capability of running queries. 

`Malloy.run` requires a `Connection`, whereas `Malloy.compile` only needs an `InfoConnection`.

We keep the concept of "Lookups", now as `LookupConnection<InfoConnection>` or `LookupConnection<Connection>`. This allows custom connection mapping.


packages/malloy-vscode/src/server/lenses/lenses.ts
```diff
@@ -111,24 +111,27 @@ export function getMalloyLenses(document: TextDocument): CodeLens[] {
...
     } else if (symbol.type === "sql") {
+      // TODO crs-sql-statement Cannot access these by name currently
       lenses.push({
         range: symbol.range.toJSON(),
         command: {
           title: "Run",
-          command: "malloy.runSqlQuery",
-          arguments: [symbol.name],
+          command: "malloy.runUnnamedSQLBlock",
+          arguments: [currentUnnamedSQLBlockIndex],
```
christopherswenson, line 120: @mtoy-googly-moogly Since the `.name` field of a SQL Block is now no longer the same as what you type in `sql: xyz is ...`, I have to access the named SQL blocks by index. Can `.name` be the given name if it's present?

mtoy-googly-moogly: Every sql block has "name: digest-hash"

In the sqlBlockList returned by the translator, every entry also has an "as:" field which will either be the name assigned by the code, or a manufactured unique name.

mtoy-googly-moogly: I was following the model/ pattern, which is that the "name" of a thing is always "thing.as || thing.name". If you''d prefer for the name to be the name and the hash to be in a field called "digest", we could do that.

mtoy-googly-moogly: I'll make the "as" undefined for anonymous blocks next time I am in that code.

packages/malloy/src/malloy.ts
```diff
@@ -283,58 +279,113 @@ export class Malloy {
...
+      }
+      return new Result(
+        {
+          structs: [schema],
+          sql: sqlBlock.select,
+          result: data.rows,
+          totalRows: data.totalRows,
+          lastStageName: schema.name,
+          // TODO feature-sql-block There is no malloy code...
+          malloy: "",
```
christopherswenson, line 362: @lloydtabb What do you suggest we do here?

packages/malloy/src/malloy.ts
```diff
@@ -283,58 +279,113 @@ export class Malloy {
...
+          connectionName: schema.structRelationship.connectionName,
+          // TODO feature-sql-block There is no source explore...
+          sourceExplore: "",
```
christopherswenson, line 365: And here @lloydtabb

packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -210,15 +216,42 @@ export class BigQueryConnection extends Connection {
...
+  public async runSQL(
+    sqlCommand: string,
+    options: Partial<BigQueryQueryOptions> = {},
+    rowIndex = 0
+  ): Promise<MalloyQueryData> {
+    const { data } = await this._runSQL(sqlCommand, options, rowIndex);
+    return data;
+  }
+
+  public async runSQLBlockAndFetchResultSchema(
```
christopherswenson, line 243: @bporterfield I've called it `runSQLBlockAndFetchResultSchema` for now... Because the most useful method is to run specifically a SQL block, and not just any old SQL. Not too sure if this is the right API though.

packages/malloy-vscode/src/server/lenses/lenses.ts
mtoy-googly-moogly: Think of the "name" as like "the actual table name" which the user also doesn't choose. If I have a table structedf, the "name" is the physical name of the table and the "as:" is what the user calls it.

christopherswenson: 🤷 That's fine by me!

packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -105,7 +105,10 @@ export class BigQueryConnection extends Connection {
   private temporaryTables = new Map<string, string>();
   private defaultProject;
 
-  private resultCache = new Map<string, MalloyQueryData>();
+  private resultCache = new Map<
```
bporterfield, line 108: Do we actually want dialect-specific result caching? Not asking for this PR, but generally...

packages/malloy-db-bigquery/src/bigquery_connection.ts
bporterfield: seems good enough 👍

bporterfield (approved):
🥳

PR #360, opened 2022-02-09
Fix infinite loop when source file is empty string

christopherswenson:
Fixes an issue where if an empty string is passed to the translator as the source for a URL, then the translator asks for that URL again, due to a bad check (`!thing` vs `thing === undefined`).


christopherswenson:
Fixes https://github.com/looker-open-source/malloy/issues/359

christopherswenson:
@mtoy-googly-moogly If this passes CI, I will merge. Can you please add a test for this?

mtoy-googly-moogly:
This fix is 100% wrong, and has broken imports, and is only passing tests because I have no tests. You basically checked in a line of code which makes it impossible for the parser to request a URL resolution.

PR #367, opened 2022-02-11
Catch-undefined-field-refs-in-queries

### LINKED ISSUE #292: Error from referencing a field that has not been defined anywhere appears at the top of the file
Error from referencing a field that has not been defined anywhere appears at the top of the file, instead of appearing where the issue is.

explore: flights is table('malloy-data.faa.flights'){}

query: flights -> {
  aggregate: flight_count
}

Screen Shot 2022-01-28 at 2 48 54 PM

mtoy-googly-moogly:
Fixes #292

image

When a query references a field, generate an error at compile time instead of waiting until SQL generation when there is less context.


lloydtabb:
So these are all related to grouping by joined explores.  Certainly not a heavily used feature but interesting.  It might be worth disabling and having some deeper conversation about...

PR #370, opened 2022-02-11
Add Jump to Definition feature to IDE

christopherswenson:
This PR builds upon https://github.com/looker-open-source/malloy/pull/364, using the new `location` field on model objects (queries, explores, SQL blocks, and fields), along with a new `model.getReference(position)` method to provide jump-to-definition functionality within the VSCode editor.

## Why

Jump-to-definition is an important feature when developing code, allowing users to better understand new code they encounter and aiding navigation within it. For instance, if someone runs a query with a field `user_lifetime_value`, instead of having to go search through the file (or possibly multiple files in a large project) to find the definition, they can simply command-click (or similar) and find the code that defines it.

## Design

To implement the `model.getReferences(position)` method, we collect all references as part of translation. In particular, the `MalloyTranslation` class gets a new field `references` of type `ReferenceList`, and all AST nodes (i.e. `MalloyElement`s) have a method `addReference`, which finds the active translator and adds a reference to the list.

### What is a `DocumentReference`?

A `DocumentReference` has

```ts
type: "exploreReference" | "joinReference" | "sqlBlockReference" | "queryReference";
text: string;
location: DocumentLocation;
definition: HasLocation;
```

The `location` field specifies the location of the reference itself, and `definition.location` specifies where the reference is defined.

### When are references tracked?

References are added during the `translateStep`. There are two places where `addReference` is called: 
1) In `MalloyElement.modelEntry` (which looks up a query, SQL block, or explore in the model). Here, if a model entry is found, it is recorded via `addReference`
2) In `StructSpace.lookup` (which looks up a field or join within an explore). Again, if a field is found, it is recorded via `addReference`

In both places, in order to know the actual location of the reference, we need to pass in an AST node (which has a `.location`) rather than a string. `StructSpace.lookup` can now take a `FieldReference` node (which is a `ListOf<FieldName>`), and `MalloyElement.modelEntry` can now accept a `ModelEntryReference` (which has a name string). In order for these to be available, every AST node which previously had some kind of `fieldName: string` or `modelEntName: string` needs to have a `fieldReference: FieldReference` or `modelEntReferene: ModelEntryReference`, which is simply a matter of producing those extra AST nodes to wrap the model entry name / field path parts during parsing.

### How are references stored and looked up?

References are stored in a `ReferenceList`, which is an implementation of a sorted array with insert and lookup via binary search. References are stored sorted by their `location.range.end` field (by line first, then character). When finding a reference at a given position, *we assume that references do not overlap*, which allows us to simply search for the position where we would insert a new reference ending at that position, and checking whether the range of the reference at that index encompasses the given position. This assumption is okay because references should always look like an identifier, and every identifier should point to exactly one thing. Additionally, dotted references (e.g. `some_join.another_join.some_field` are stored as multiple separate references, with each dotted part pointing to a different object.

### Future Work

Currently, two references are not tracked. These are `top: n by <reference>` and `order_by: <reference>`. These are not tracked due to the fact that they are evaluated against the _output field space_ of a query, which is currently not accessible. @mtoy-googly-moogly is planning on making this available to fix `having` statements, and when that happens, it should be simple to re-enable these references as well.


packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -950,6 +1027,38 @@ export class FieldName
...
+  get head(): FieldName {
+    return this;
+  }
+
+  get rest(): FieldName | undefined {
+    return undefined;
+  }
+}
+
+export class FieldPath extends FieldName implements FieldReferenceInterface {
```
mtoy-googly-moogly, line 1040: Maybe a path is a `ListOf<FieldName>` that also implements FieldReferenceInterface?

christopherswenson: I admit this is kinda weird.... but currently a `FieldPath` implements `FieldName`, so the `rest` could be a `FieldPath`. So the structure of `explore.join.join.join.fieldname` would be
```ts
new FieldPath(
  new FieldName("explore"),
  new FieldPath(
    new FieldName("join"),
    new FieldPath(
      new FieldName("join"),
      new FieldPath(
        new FieldName("join"),
        new FieldName("fieldname")
      )
    )
  )
)
```
But I'm going to try `Listof<FieldName>` because that makes more sense....

packages/malloy/src/lang/field-space.ts
```diff
@@ -243,30 +260,34 @@ export class NewFieldSpace extends StructSpace {
...
           continue;
         }
-        const oldValue = this.entry(def.oldName);
-        if (oldValue instanceof SpaceField) {
-          this.setEntry(
-            def.newName,
-            new RenameSpaceField(oldValue, def.newName, def.location)
-          );
-          this.dropEntry(def.oldName);
+        const oldValue = def.oldName.getField(this);
```
mtoy-googly-moogly, line 275: I still wonder if this should count as a reference

packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -786,11 +803,7 @@ export abstract class MalloyTranslation {
       line: startToken.line - 1,
       character: startToken.charPositionInLine,
     };
-    if (
-      this.parseStep.sourceInfo &&
-      stopToken.stopIndex != -1 &&
-      stopToken.stopIndex != startToken.startIndex
-    ) {
+    if (this.parseStep.sourceInfo && stopToken.stopIndex != -1) {
```
mtoy-googly-moogly, line 806: i prefer to explicity not look for the stop index if i already know it is the same as the start

packages/malloy/src/model/malloy_types.ts
```diff
@@ -38,15 +38,58 @@ export function paramHasValue(p: Parameter): boolean {
...
   range: DocumentRange;
 }
 
+interface DocumentReferenceBase {
+  text: string;
+  location: DocumentLocation;
+  definition: HasLocation;
+}
+
+export interface DocumentExploreReference extends DocumentReferenceBase {
```
mtoy-googly-moogly, line 61: should these be `stuctDef: StructDef`, `fieldDef: FieldEf` instead of `definition: ` ?

mtoy-googly-moogly (changes requested):
I've mad a pass with some comments, some will require changes, some are optional.

packages/malloy/src/model/malloy_types.ts
christopherswenson: I see your point... though it's just nice to always be able to say `reference.definition.location`.

packages/malloy/src/lang/field-space.ts
christopherswenson: `file1.malloy`
```
explore: base is table('somewhere') {
  dimension: thing_with_horrible_name is COMPLEX_EXPRESSION;
}
```
`file2.malloy`
```
import "file1.malloy"

explore: thing is base {
  rename: thing is thing_with_horrible_name
}
```

In this case, I want to Command+Click on `thing_with_horrible_name` and see how it's defined.

packages/malloy/src/lang/parse-malloy.ts
christopherswenson: Okay, I see what's happening now. There was a bug with `stopToken.stopIndex != startToken.startIndex` because it turns out `stopIndex` is the index of the last character, not the index after the last character. So single-character tokens were getting the wrong locations. I fixed the bug by looking up the end location in this case, but the correct fix is `stopToken.stopIndex >= startToken.startIndex` or `stopToken.stopIndex != startToken.startIndex - 1`, which is the case you're trying to avoid looking up the end location.

christopherswenson:
```
if (
  this.parseStep.sourceInfo &&
  stopToken.stopIndex != -1 &&
  // If the `stopToken.stopIndex` is one before the `startToken.startIndex`, it means
  // it's length 0 and we don't need to look up the end.
  stopToken.stopIndex != startToken.startIndex - 1
) {
```

packages/malloy/src/lang/ast/ast-expr.ts
```diff
@@ -424,9 +429,9 @@ export class ExprIdReference extends ExpressionDef {
   }
 
   apply(fs: FieldSpace, op: string, expr: ExpressionDef): ExprValue {
-    const def = fs.lookup(this.refString);
-    if (def.found instanceof SpaceParam) {
-      const cParam = def.found.parameter();
+    const entry = this.fieldReference.getField(fs);
```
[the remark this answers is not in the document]
christopherswenson: Great catch. Fixed in https://github.com/looker-open-source/malloy/pull/370/commits/551693747d0a48bc863a75f9e75a393a2cad169d

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -355,13 +386,22 @@ export abstract class Mallobj extends MalloyElement {
 
 class QueryHeadStruct extends Mallobj {
   elementType = "internalOnlyQueryHead";
-  constructor(readonly fromRef: model.StructRef) {
+  constructor(readonly fromRef: ModelEntryReference | model.StructRef) {
```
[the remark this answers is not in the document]
christopherswenson: Ah. Yes. Fixed in https://github.com/looker-open-source/malloy/pull/370/commits/620eeb9eb6edc53a44b3a321872a0dd9aec620eb

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -553,12 +593,28 @@ export class IsValueBlock extends MalloyElement {
...
+  }
+}
+
 export class NamedSource extends Mallobj {
   elementType = "namedSource";
   protected isBlock?: IsValueBlock;
 
   constructor(
-    readonly name: string,
+    readonly ref: ModelEntryReference | string,
```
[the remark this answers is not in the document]
christopherswenson: Yep! Good catch! Fixed in https://github.com/looker-open-source/malloy/pull/370/commits/794c24773e505984ce2dbda5828419ebae7fd77d

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -979,30 +1049,70 @@ export class FieldName
...
+    return this.list[this.list.length - 1].refString;
+  }
+
+  getField(inFs: FieldSpace): LookupResult {
+    return inFs.lookup(this.list);
+  }
 }
 
-interface FieldReferenceInterface {
+interface NameLike {
```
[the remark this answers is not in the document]
christopherswenson: Cancelled https://github.com/looker-open-source/malloy/pull/370/commits/fba0ad66c8ec4cef01a720de3f789852ea8a263f

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -1490,39 +1603,65 @@ export class PrimaryKey extends MalloyElement {
...
 
   byElement(): model.OrderBy {
-    const orderElement: model.OrderBy = { field: this.field };
+    const orderElement: model.OrderBy = { field: this.modelField };
     if (this.dir) {
       orderElement.dir = this.dir;
     }
     return orderElement;
   }
+
```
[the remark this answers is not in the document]
christopherswenson: Love it. https://github.com/looker-open-source/malloy/pull/370/commits/b2d0bef475090604442983e6953b2283f10812b8

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -1533,6 +1672,10 @@ export class Ordering extends ListOf<OrderBy> {
   orderBy(): model.OrderBy[] {
     return this.list.map((el) => el.byElement());
   }
+
```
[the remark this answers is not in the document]
christopherswenson: https://github.com/looker-open-source/malloy/pull/370/commits/b2d0bef475090604442983e6953b2283f10812b8

packages/malloy/src/lang/parse-malloy.ts
mtoy-googly-moogly: I think the only zero length token might be the EOF token, and that won't be a startToken, and so I think we can live with no change here.

packages/malloy/src/lang/ast/ast-main.ts
```diff
@@ -1759,81 +1885,94 @@ export class PipelineDesc extends MalloyElement {
...
   }
 }
 
-export class NestReference extends MalloyElement {
+export class NestReference extends FieldReference {
   elementType = "nestReference";
-  constructor(readonly name: string) {
-    super();
+  constructor(readonly name: FieldName) {
+    super([name]);
```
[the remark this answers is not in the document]
christopherswenson: This... doesn't really matter for right now. Until the day comes when we fix `nest: some_join.some_query`, which turns out is hard.

PR #377, opened 2022-02-14
Fix/improve error locations

christopherswenson:
This PR attaches correct (or at least, more correct) locations to errors. Firstly, it fixes an issue from https://github.com/looker-open-source/malloy/pull/364 where semantic errors appeared on the line before they were supposed to. It also adjusts the location of syntax errors to only include the offending token, and not the offending token plus the rest of that line.

### DIFF
--- a/packages/malloy/src/lang/parse-malloy.ts
+++ b/packages/malloy/src/lang/parse-malloy.ts
@@ -669,13 +669,13 @@ export abstract class MalloyTranslation {
             }
           }
           if (lineMap[this.sourceURL]) {
-            const errorLine = lineMap[this.sourceURL][lineNo - 1];
-            cooked = `line ${lineNo}: ${entry.message}\n  | ${errorLine}`;
+            const errorLine = lineMap[this.sourceURL][lineNo];
+            cooked = `line ${lineNo + 1}: ${entry.message}\n  | ${errorLine}`;
             if (charFrom > 0) {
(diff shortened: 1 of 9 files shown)


packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -61,13 +62,16 @@ class ParseErrorHandler implements ANTLRErrorListener<Token> {
...
+          start: { line: 0, character: 0 },
+          end: { line: 0, character: 0 },
+        };
     const error: LogMessage = {
-      sourceURL: this.sourceURL,
       message: msg,
-      begin: {
-        line: line,
-        char: charPositionInLine,
-      },
```
christopherswenson, line 70: There was previously no `end` specified on syntax errors, so the IDE (language server diagnostic code, specifically) was filling in an `end` of the end of the line.

mtoy-googly-moogly:
ok, i broke this fix with my PR -- fixing the broken fix, i should have tested before i pushed.

mtoy-googly-moogly:
Fixing it by adding tests with `markSource` 🥳  because @christopherswenson is amazing!

mtoy-googly-moogly:
@christopherswenson this makes ALL the location reference code in lang, error messages AND the external reference locations, use the DocumentLocation datatype.

Added a couple of tests to see that error messages highlight properly.

Ready for review.

christopherswenson:
@mtoy-googly-moogly Looks good to me 👍 Thanks for the help here!

PR #399, opened 2022-02-20
Allow expressions in join: ... with 

### LINKED ISSUE #347: When using `with` join syntax, cannot use a foreign key that's nested or accessed with `.`
It seems that `with` isn't accepting `.` in joins

this is erroring:
```malloy
explore: products is table('malloy-data.ecomm.products')

explore: inventory_items is table('malloy-data.ecomm.inventory_items')

explore: order_items_test is table('malloy-data.ecomm.order_items'){
  join_one: inventory_items with inventory_item_id
  join_one: products with inventory_items.product_id
}
```

`extraneous input '.' expecting {ACCEPT, DIMENSION, EXCEPT, JOIN_CROSS, JOIN_ONE, JOIN_MANY, MEASURE, PRIMARY_KEY, QUERY, RENAME, WHERE, '}', ';'}`

Screen Shot 2022-02-04 at 3 04 03 PM

but if I write out the join with `on` instead of using the shorter `with` syntax it's happy:
```malloy
explore: order_items_test is table('malloy-data.ecomm.order_items'){
  join_one: inventory_items with inventory_item_id
  join_one: products on inventory_items.product_id = products.id
}
```

mtoy-googly-moogly:
Fixes #347

Once upon a time we had plans to allow reversed foreign key join with something like

```
source: users {
  join_many: orders with orders.user_id
```

And automatically generate a join into the primary key of users.

Only we never implemented that, and in the grammar a `with` clause could ONLY be a field name with no dots.

---

With this PR the `with` clause takes any expression, and always joins based on the primary key of the source being join in. This will allow any expression based on the outer source to be used as a foreign key for joining the new source, including, as mentioned in #347 _exisitingJoinName.fieldName_

### DIFF
--- a/docs/_src/language/join.md
+++ b/docs/_src/language/join.md
@@ -11,7 +11,7 @@ In Malloy, syntaxes for join are:
 
 ```malloy
 join_one: <source-name> [is <source-exp>] on <boolean expression>
-join_one: <source-name> [is <source-exp>] with <foreign_key>        -- slightly more concise; works with primary to foreign key joins
+join_one: <source-name> [is <source-exp>] with <foreign key expression>
 join_many: <source-name> [is <source-exp>] on <boolean expression>
 join_cross: <source-name> [is <source-exp>] [on <boolean expression>]
 ```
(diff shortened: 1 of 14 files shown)


lloydtabb:
Let's talk about this.  I think I want an expression.  There is no reason this needs to be a field...

PR #438, opened 2022-03-19
re-org on path to publishing packages. remove vscode from packages fo…

bporterfield:
…lder (its not really a package). also, rename files to use hyphens across project for consistency.


bporterfield:
will re-create in new PR

PR #447, opened 2022-03-31
Use original spacing for `code` of `FieldDef`s (continuation of #442)

christopherswenson:
This PR applies the same change as #442 applied to filter expressions, but to field expressions. Now, field expressions' `code` field (renamed from `source`) will match the original spacing of the source code.

### DIFF
--- a/packages/malloy/src/lang/ast/ast-expr.ts
+++ b/packages/malloy/src/lang/ast/ast-expr.ts
@@ -248,7 +248,7 @@ export class FieldDeclaration extends MalloyElement {
         template.aggregate = true;
       }
       if (this.exprSrc) {
-        template.source = this.exprSrc;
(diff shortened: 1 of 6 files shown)


mtoy-googly-moogly:
I thought the `code` was only for filter expressions because we only used that for drills. Who needs code for field expressions?

mtoy-googly-moogly:
... as in, can we just DELETE the code field for field expressions instead?

mtoy-googly-moogly:
I mention this because there seem to be zero consumers of this data.

mtoy-googly-moogly:
nvm ... i see we seem to use this in drills also.

i hate the way we do drills, at some point i am going to insist we stop doing that.

mtoy-googly-moogly (commented):
all good, merged!

PR #459, opened 2022-04-09
updated gemfile

anikaks:


anikaks:
`yarn docs-serve` was not working on my new mac; needed to use a newer version of eventmachine.

bporterfield:
closing b/c of inactivity - we're not going this route ya?

anikaks:
Since we couldn't get it working, for now we just can't use `yarn docs-serve` on M1 macs (or it requires this local change and to remember not to push it...). We should fix this at some point but it's not super urgent as only I am on an M1 mac so far, and I am ok to just use my old laptop for docs work for now.

PR #475, opened 2022-04-13
integrate some feedback

anikaks:


anikaks:
decided to hold off on adding `?` for now because it's not yet in pre-release. once it hits pre-release we can doc it everywhere.

PR #480, opened 2022-04-17
Render null dates, closes #448

### LINKED ISSUE #448: Render null dates as nulls, rather than "invalid"
Rather than showing as null, null dates produce this error: `Invalid field for date renderer`

Here's a query to produce null dates:
```
query: table('malloy-data.faa.airports') -> {group_by: always_null is null::date}
```

greg-finley:
Screen Shot 2022-04-17 at 9 18 59 AM


lloydtabb:
Awesome, thanks!

PR #481, opened 2022-04-18
Wrap numeric terms in parens, closes #454

### LINKED ISSUE #454: Need to generate parenthesis around numeric expressions
These two fields currently produce different results, and should not. 
```malloy
query: table('malloy-data.ecomm.inventory_items') -> {
  declare: gross_margin is product_retail_price - cost
  project: 
    gross_margin_pct_good is (gross_margin) / nullif(product_retail_price,0)
    gross_margin_pct_bad is gross_margin / nullif(product_retail_price,0)
}
```
Screen Shot 2022-04-05 at 6 53 48 PM

This is the SQL generated:
```sql
SELECT 
   ((base.product_retail_price-base.cost))/(nullif(base.product_retail_price,0)) as gross_margin_pct_good,
   base.product_retail_price-base.cost/(nullif(base.product_retail_price,0)) as gross_margin_pct_bad
FROM `malloy-data.ecomm.inventory_items` as base
```

greg-finley:
Basically, wrap any numerical operation in parens. Maybe it's now over–paren-ing in some spots (it's not immediately clear how to safely dedupe), but the bug is fixed and the resulting SQL from [the issue](https://github.com/looker-open-source/malloy/issues/454) becomes:

```sql
SELECT 
   ((((base.product_retail_price-base.cost)))/(nullif(base.product_retail_price,0))) as gross_margin_pct_good,
   ((base.product_retail_price-base.cost)/(nullif(base.product_retail_price,0))) as gross_margin_pct_bad
FROM `malloy-data.ecomm.inventory_items` as base
```

Screen Shot 2022-04-17 at 10 41 33 PM

### DIFF
--- a/packages/malloy/src/lang/ast/ast-types.ts
+++ b/packages/malloy/src/lang/ast/ast-types.ts
@@ -186,9 +186,9 @@ export function errorFor(reason: string): ExprValue {
  * If the passed expresion is not a single term, wrap it in parens
  * @param f expression fragment
  */
-function term(f: Fragment[]): Fragment[] {
+function composeTerm(f: Fragment[]): Fragment[] {
   if (f.length > 1) {
-    return ["(", ...f, ")"];
+    return wrapInParens(f);
   }
   if (f.length === 0) {
     // Trying to compose a binary expresion with an entity that has no value
@@ -197,6 +197,10 @@ function term(f: Fragment[]): Fragment[] {
     return ["__MISSING_VALUE__"];
   }
   return f;
+}
+
+function wrapInParens(f: Fragment[]): Fragment[] {
+  return ["(", ...f, ")"];
 }
 
 export function compressExpr(expr: Expr): Expr {
@@ -225,18 +229,21 @@ export function compressExpr(expr: Expr): Expr {
  * @param left
  * @param op
  * @param right
+ * @param needsParens
  * @returns Fragment list of the composed expression
  */
 export function compose(
   left: Fragment[],
   op: string,
-  right: Fragment[]
+  right: Fragment[],
+  needsParens?: true
 ): Fragment[] {
   const opAlpha = op.match(/^[A-Za-z]/);
   const leftSpace = left.length === 1 && opAlpha ? " " : "";
   const rightSpace = right.length === 1 && opAlpha ? " " : "";
   const newOp = leftSpace + op + rightSpace;
-  return [...term(left), newOp, ...term(right)];
+  const composed = [...composeTerm(left), newOp, ...composeTerm(right)];
+  return needsParens ? wrapInParens(composed) : composed;
 }
 
 export function dateOffset(
--- a/packages/malloy/src/lang/ast/apply-expr.ts
+++ b/packages/malloy/src/lang/ast/apply-expr.ts
@@ -223,7 +223,8 @@ function numeric(
     return {
       dataType: "number",
       aggregate: anyAggregate,
-      value: compose(lhs.value, op, rhs.value),
+      // set needsParens to true for mathematical order of operations
+      value: compose(lhs.value, op, rhs.value, true),
     };
   }
 


packages/malloy/src/lang/ast/ast-types.ts
```diff
@@ -199,6 +199,10 @@ function term(f: Fragment[]): Fragment[] {
   return f;
 }
 
+function wrapInParens(f: Fragment[]): Fragment[] {
+  return ["(", ...f, ")"];
+}
```
greg-finley, line 204: I considered only wrap if not currently wrapped, but I'm not sure if it's safe:

```ts
function wrapInParens(f: Fragment[]): Fragment[] {
  if (f.length && f[0] === '(' && f[f.length] === ')') {
    return f;
  }
  return ["(", ...f, ")"];
}
```

lloydtabb:
I've been slow to fix this one because I want to keep the SQL readable.  Looking to get some paren hints from the parser.

lloydtabb:
/gcbrun 

but probably won't merge.  Thanks for the effort and raising the importance.  

@mtoy-googly-moogly , perhaps we can talk about knowing about fragment complexity?

greg-finley:
Thanks for taking a look!

I'm curious to learn, if you have a moment. What's the ultimate solution?
1. "Wrap all numeric math in parens" _and_ "dedupe unnecessary parens"
2. "Add parens only when mathematically necessary"
3. Something else?

No. 1 would add parens to all math, but maybe that's ok? The "dedupe unnecessary parens" part [seems doable](https://stackoverflow.com/questions/26763975/removing-unnecessary-duplicates-parentheses-from-arithmetic-expressions-using-st) and maybe beneficial for all queries?

I think No. 2 would require a lot more state passed around and more complexity (adding five columns together requires no parens, but a multiplication needs to know if it's surrounded by operations lower on the order of operations, etc.).

lloydtabb:
My thinking is that the parser knows the complexity of an expression subtree.  An expression can know if it would need to be parenthesized if it were used as part of another expression.  I will know for sure with a conversation with @mtoy-googly-moogly .  

Of course, what you wrote will work, but I'd like the SQL code to be as readable as possible and we haven't really tried yet.

mtoy-googly-moogly:
Hey @greg-finley welcome to malloy! Thanks.

Here's my thinking.

Some of the oldest code in the translator is hidden in the fragment generation and I've made a lot of experiments with parens and learned a bit. If I could rewind time, I would always compose fragments with functions and never by hand, and I would enforce parens rules in the composition functions.  The function "does this need parens" or "de-dup parens" requires you to re-parse the string, which you really can't do because the fragments might contain quoted strings with parens in them. It is kind of too late, at fragment time, to look for paren rules. I know this because I went down this road.

So this code is wanting a spring cleaning and I have two plans. One is essentially what I mentioned above, to always compose fragments with functions, and make those functions smarter about parens.

The other, and the likely actual plan, is to have the source code-to-data translator write expression trees to the JSON intermediate instead of fragment trees, and the write a smarter tree walker/expression writer in the SQL generation step.

In the meantime, just write extra parens to keep things correct is the right answer.

mtoy-googly-moogly:
The extra parens bug me a lot. By the way. Another possible solution would be to just parse and then pretty-print the resulting SQL. This parser wouldn't have to parse the full sql language, just the subset that malloy generates. I don't know if this idea is brilliant or horrible.

greg-finley:
Awesome, thanks for all the explanation! I am new to AST stuff so need a bit of time to grok it.

Feel free to ditch this PR if it's not heading in the right direction, though I would say "query gives the right answer" might trump extra parens for the short term.

packages/malloy/src/lang/ast/ast-types.ts
mtoy-googly-moogly: Yeah, it isn't safe. There is no guarantee that a fragment doesn't look like

```
[ "(  'start of some string which contains)" ]
```

or

```
[ "(expr1) or (expr2)" ]
```

If fragments were type-protected and we could know that they were well formed, then something like this check would work.

mtoy-googly-moogly:
OK, I've looked at the PR. I feel like I have found a friend. Yes, the parens bug me, yes, you wanted the same thing to work that I did. No, that doesn't quite work. I'm going to suggest not taking the PR as it stands, but I feel horrible because you are so on the right track.

greg-finley:
No need to feel horrible! I just looked for a quick solution but you're right to consider these more holistic plans.

lloydtabb:
Fixed another way, see #483

PR #500, opened 2022-04-21
Handle Postgres table paths with capitals, closes #495, closes #499

### LINKED ISSUE #495: Cannot query Postgres tables with capital letters
Currently, you cannot directly query a Postgres table with capital letters in its name, because identifiers that are not double-quoted are treated as lowercase. 

When querying in Postgres, one simply double-quotes the name when needed, but we cannot do this inside of Malloy's `table()` function currently.

The temporary, less than ideal, workaround is to do something like this:
```
sql: airports_table is || select * from malloytest."Airports";;

query: from_sql(airports_table) -> my_query
```

### LINKED ISSUE #499: Always double-quote the name in postgres
Given this:

```
query: table('public.HolidayEvent') -> {
  project: id
  limit: 10
}
```

The query run fails with the error `relation public.holidayevent` does not exist.

To support case-sensitive in postgres, we should always double-quote the table name, similar to changes in https://github.com/looker-open-source/malloy/pull/487.

greg-finley:
Closes https://github.com/looker-open-source/malloy/issues/495

`packages/malloy-db-test/src/postgres.spec.ts` isn't quite set up to easily make and drop schemas and tables, but an automated test could be added eventually to mirror the manual test below:

Screen Shot 2022-04-20 at 11 36 26 PM

### DIFF
--- a/packages/malloy/src/dialect/postgres.ts
+++ b/packages/malloy/src/dialect/postgres.ts
@@ -61,8 +61,11 @@ export class PostgresDialect extends Dialect {
   divisionIsInteger = true;
   functionInfo: Record<string, FunctionInfo> = {};
 
-  quoteTableName(tableName: string): string {
-    return `${tableName}`;
+  quoteTablePath(tablePath: string): string {
+    return tablePath
+      .split(".")
+      .map((part) => `"${part}"`)
+      .join(".");
   }
 
   sqlGroupSetTable(groupSetCount: number): string {
--- a/packages/malloy-db-test/src/postgres.spec.ts
+++ b/packages/malloy-db-test/src/postgres.spec.ts
@@ -126,7 +126,7 @@ describe("postgres tests", () => {
     expect(result.data.value[0].n).toBe(1);
   });
 
-  it(`sql_block`, async () => {
+  it(`quote field name`, async () => {
     const result = await runtime
       .loadQuery(
         `
--- a/packages/malloy/src/dialect/dialect.ts
+++ b/packages/malloy/src/dialect/dialect.ts
@@ -47,8 +47,8 @@ export abstract class Dialect {
   abstract divisionIsInteger: boolean;
   protected abstract functionInfo: Record<string, FunctionInfo>;
 
-  // return a quoted string for use as a table name.
-  abstract quoteTableName(tableName: string): string;
+  // return a quoted string for use as a table path.
+  abstract quoteTablePath(tablePath: string): string;
 
   // returns an table that is a 0 based array of numbers
   abstract sqlGroupSetTable(groupSetCount: number): string;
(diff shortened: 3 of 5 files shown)


packages/malloy/src/dialect/postgres.ts
```diff
@@ -61,8 +61,11 @@ export class PostgresDialect extends Dialect {
   divisionIsInteger = true;
   functionInfo: Record<string, FunctionInfo> = {};
 
-  quoteTableName(tableName: string): string {
-    return `${tableName}`;
+  quoteTablePath(tablePath: string): string {
+    return tablePath
+      .split(".")
+      .map((part) => `"${part}"`)
```
greg-finley, line 67: I debated doing "add quotes if it doesn't currently have them", like:

```ts
      .map((part) => {
        if (part.startsWith('"')) {
          return part;
        }
        return `"${part}"`;
      })
 ```
 
 but unclear if it's needed

**Comment by lloydtabb** [comment]
This will make things better.  Not sure if table names can have quoted characters in their names.
We probably need a test here.  
I don't love that we take a quoted string and re-interpret it.

**Comment by lloydtabb** [comment]
@mtoy-googly-moogly I think we should talk about the `table` function.  We are interpreting table paths and we're fighting ourselves and the semantics of the database in the same string...

**Comment by greg-finley** [comment]
> This will make things better. Not sure if table names can have quoted characters in their names. We probably need a test here. I don't love that we take a quoted string and re-interpret it.

Yeah, I feel the same.

Btw, for tests, I was thinking that the tests should make and destroy tables with capital letters:

```ts
beforeAll(async () => {
  await runtime.executeSQLRaw(
    'create schema if not exists "SchemaUpper";'
  );
  await Promise.all([
    runtime.executeSQLRaw('create table if not exists "SchemaUpper"."SchemaTableUPPER";'),
    runtime.executeSQLRaw('create table if not exists "RegularUpper";'),
  ]);
});

afterAll(async () => {
  await Promise.all([
    runtime.executeSQLRaw('drop schema if exists "SchemaUpper";'),
    runtime.executeSQLRaw('drop table if exists "RegularUpper";'),
  ])
  await runtimeList.closeAll();
});
```

... and then do tests on these tables. But we'd need to implement `executeSQLRaw` on `runtime`, or have some other way to control these tables.

lloydtabb:
You can get the connection and then runSQL against that...  I'm worried about tests failing because when they test the table is there, but someone else running the test hits the drop table code...  We should probably test with temp tables or something...

greg-finley:
Ah, gotcha.

Good point, I was considering just local, but you're right that tests could step on each other if they are in the same db. I can take this as a follow-up.

greg-finley:
I think the right balance is create the schemas/tables idempotently (ie “if not exists”) but don’t remove them. This way new DBs get seeded and further runs are no-op.

PR #501, opened 2022-04-22
Parens around join filters, closes #469

### LINKED ISSUE #469: Bug: When joining a source which has been filtered, parens are not generated where they should be.
This model
```
source: users is table('malloy-data.ecomm.users'){
  primary_key: id
  where: state = 'California' | 'New York' 
}
source: order_items is table('malloy-data.ecomm.order_items'){
  join_one: users with user_id
}
query: order_items -> {
  group_by: users.state
}
```

generates this SQL:
```
SELECT 
   users_0.state as state
FROM `malloy-data.ecomm.order_items` as order_items
LEFT JOIN `malloy-data.ecomm.users` AS users_0
  ON users_0.id=order_items.user_id AND (users_0.state='California')or(users_0.state='New York')
GROUP BY 1
ORDER BY 1 asc
```

We should be generating parentheses around the whole `where`  which is being translated into the join.

greg-finley:
Query [from issue](https://github.com/looker-open-source/malloy/issues/469) becomes

```sql
SELECT 
   users_0.state as state
FROM `malloy-data.ecomm.order_items` as order_items
LEFT JOIN `malloy-data.ecomm.users` AS users_0
  ON users_0.id=order_items.user_id AND ((users_0.state='California')or((users_0.state='New York')or(users_0.state='Oregon')))
GROUP BY 1
ORDER BY 1 asc
```

An alternative would be "add parens only if `conditions` has an 'or'".

### DIFF
--- a/packages/malloy-db-test/src/nomodel.spec.ts
+++ b/packages/malloy-db-test/src/nomodel.spec.ts
@@ -192,6 +192,7 @@ runtimes.runtimeMap.forEach((runtime, databaseName) => {
       `
       )
       .run();
+    // https://github.com/looker-open-source/malloy/pull/501#discussion_r861022857
     expect(result.data.value).toHaveLength(3);
     expect(result.data.value).toContainEqual({ c: 1845, state: "TX" });
     expect(result.data.value).toContainEqual({ c: 500, state: "LA" });
--- a/packages/malloy-db-test/src/nomodel.spec.ts
+++ b/packages/malloy-db-test/src/nomodel.spec.ts
@@ -39,7 +39,7 @@ afterAll(async () => {
 
 runtimes.runtimeMap.forEach((runtime, databaseName) => {
   // Issue: #151
-  it(`unknonwn dialect  - ${databaseName}`, async () => {
+  it(`unknown dialect  - ${databaseName}`, async () => {
     const result = await runtime
       .loadQuery(
         `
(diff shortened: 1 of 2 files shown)


lloydtabb:
Greg, thanks for all your contributions.  It would be great to have a test for this...  Perhaps next to this one?

https://github.com/looker-open-source/malloy/blob/ab3130c51ddaa5156f08732196fb7e6fbb86030e/packages/malloy-db-test/src/nomodel.spec.ts#L137

greg-finley:
Sounds good, will add it over the weekend.

packages/malloy-db-test/src/nomodel.spec.ts
```diff
@@ -150,6 +175,31 @@ runtimes.runtimeMap.forEach((runtime, databaseName) => {
...
+      }
+      `
+      )
+      .run();
+    expect(result.data.value[0].c).toBe(1845);
+    expect(result.data.value[0].state).toBe("TX");
+    expect(result.data.value[1].c).toBe(500);
+    expect(result.data.value[1].state).toBe("LA");
+    expect(result.data.value[2].c).toBeNull();
+    expect(result.data.value[2].state).toBeNull();
```
greg-finley, line 200: Hmmm ... Postgres sorts it the opposite way. Will look into why.

greg-finley: Ah, when it's PG, when need to do `NULLS LAST` when ordering descending. It also works in BigQuery.

greg-finley: Put a test for now that doesn't consider the order, since BQ and PG sort nulls differently. Depending on whether Malloy should always sort nulls the same way regardless of language, this could be split into its own issue and these tests changed back to check the order after that.

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#order_by_clause
> If null ordering is not specified with NULLS FIRST or NULLS LAST:
> NULLS FIRST is applied by default if the sort order is ascending.
> NULLS LAST is applied by default if the sort order is descending.

https://www.postgresql.org/docs/current/queries-order.html
> By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.

PR #504, opened 2022-04-23
Tests for Postgres schemas and tables with capital letters

greg-finley:
Following the discussion on: https://github.com/looker-open-source/malloy/pull/500

Lloyd mentioned using temporary tables for the tests, but unfortunately [sounds like](https://kb.objectrocket.com/postgresql/create-a-temp-table-in-postgresql-761#:~:text=Temporary%20tables%20exist%20in%20their,a%20transaction%20or%20a%20session.) it doesn't work for schemas:
> Temporary tables exist in their own unique schema, so you can’t assign a schema name when you create this kind of table. 

My idea is create the tables in an idempotent way before the tests. They are created on the first run and no-op subsequently, so concurrent test runs should not step on each other.

### DIFF
--- a/packages/malloy-db-test/src/postgres.spec.ts
+++ b/packages/malloy-db-test/src/postgres.spec.ts
@@ -20,6 +20,19 @@ if (runtime === undefined) {
   throw new Error("Couldn't build runtime");
 }
 
+// Idempotently create schema and tables with capital letters to use in tests.
+beforeAll(async () => {
+  await runtime.connection.runSQL('create schema if not exists "UpperSchema";');
+  await Promise.all([
+    runtime.connection.runSQL(
+      'create table if not exists "UpperSchema"."UpperSchemaUpperTable" as select 1 as one;'
+    ),
+    runtime.connection.runSQL(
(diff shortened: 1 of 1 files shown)


packages/malloy-db-test/src/postgres.spec.ts
```diff
@@ -20,6 +20,19 @@ if (runtime === undefined) {
...
+  await runtime.connection.runSQL('create schema if not exists "UpperSchema";');
+  await Promise.all([
+    runtime.connection.runSQL(
+      'create table if not exists "UpperSchema"."UpperSchemaUpperTable" as select 1 as one;'
+    ),
+    runtime.connection.runSQL(
+      'create table if not exists "UpperTablePublic" as select 1 as one;'
+    ),
+  ]);
+});
```
greg-finley, line 34: these could instead be part of `test/sql/malloytest-postgres.sql.gz` that we were just discussing, though I someone would need to run it one time on your side.

greg-finley, line 34: these could instead be part of `test/sql/malloytest-postgres.sql.gz` that we were just discussing, though I someone would need to run it one time on your side.

PR #505, opened 2022-04-25
Add instructions on local Postgres tests

greg-finley:


developing.md
```diff
@@ -11,6 +11,15 @@ yarn install
 yarn package-extension
 ```
 
+Code formatting can be achieved via:
+```bash
+yarn lint --fix
```
lloydtabb, line 16: I would recommend installing these in VSCode.  We are all using VSCode to do development.

image

@christopherswenson do you have an opinion here?

greg-finley: Good point, mentioned them here: https://github.com/looker-open-source/malloy/pull/505/commits/9b7c032e99a6e2d2f5126107e2960daeb294cc10

PR #519, opened 2022-04-28
Move to a custom matcher for SQL result checking

mtoy-googly-moogly:
This is not correct yet. I have some test errors introduced by moving this into a matcher than I don't quite understand, but I am ready for a review of the overall approach.

Not 100% sure this is how we should go, but after working in a couple of different problems, I have managed to abstract out a pattern for a custom matcher which I find useful. If we like this, I'll make more tests use this instead of the way they are testing now, it makes the tests, and test failures, way easier to understand.

It has the following features

1) To use this feature you need three lines of code in your file ..

```Typescript
import "./is-sql-eq";
import { mkSqlEqWith } from "./sql-eq";

// and the inside the loop where you are iterating over the connections generating a runtime ...
// you set up some inital values ... that your expressions can use, and that can come
// from a select statement, or a refinement, or both

  const sqlEq = mkSqlEqWith(runtime, {
    sql: "SELECT statement goes here",
    malloy: " + { dimension: malloy_code_goes_here }",
  });

// for example, here is code which loads different SQL depending on the connection

  const sqlEq = mkSqlEqWith(runtime, { sql: basicTypes[databaseName] });
```

2) The code to test a malloy fragment against an expected value result is pretty short and clear

```Typescript
      test(`timestamp delta hours - ${databaseName}`, async () => {
        const eq = await sqlEq("t_timestamp + 10 hours", "@2021-02-24 13:05:06");
        expect(eq).isSqlEq();
      });
 ```
 
3) Here's some sample output where the test fails. You can see the Malloy expression and the SQL result which didn't match and the SQL which generates that. I found it clearer to debug if I write the computation in it's own stage.
 ```
  ● time operations - postgres › delta - postgres › timestamp delta hours - postgres

    sqlEq failed
        Expected: t_timestamp + 10 hours == @2021-02-24 13:05:06.
        Received: 2021-02-24 13:00:00-10
    SQL:
        WITH __stage0 AS (
          SELECT 
             DATE_TRUNC('second', TIMESTAMP '2021-02-24 13:05:06') as "expect",
             DATE_TRUNC('hour', ((basicTypes."t_timestamp")+make_interval(hours=>10))) as "got"
          FROM ( 
              SELECT
                DATE('2021-02-24') as t_date,
                '2021-02-24 03:05:06'::timestamp with time zone as t_timestamp
             ) as basicTypes
        )
        , __stage1 AS (
          SELECT 
             CASE WHEN DATE_TRUNC('second', base."expect")=DATE_TRUNC('hour', base."got") THEN '=' ELSE concat('sqlEq failed
            Expected: t_timestamp + 10 hours == @2021-02-24 13:05:06.
            Received: ',cast(DATE_TRUNC('hour', base."got") as varchar)) END as "calc"
          FROM "__stage0" as base
        )
        SELECT row_to_json(finalStage) as row FROM __stage1 AS finalStage

      332 |       test(`timestamp delta hours - ${databaseName}`, async () => {
      333 |         const eq = await sqlEq("t_timestamp + 10 hours", "@2021-02-24 13:05:06");
    > 334 |         expect(eq).isSqlEq();
          |                    ^
      335 |       });
      336 |       test(`timestamp delta week - ${databaseName}`, async () => {
      337 |         const result = await sqlEq(

      at Object.<anonymous> (packages/malloy-db-test/src/datetimes.spec.ts:334:20)
```

How I accomplish this is a little wonky and needs review though.

### DIFF
--- a/packages/malloy-db-test/src/datetimes.spec.ts
+++ b/packages/malloy-db-test/src/datetimes.spec.ts
@@ -14,8 +14,9 @@
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 
-import { Result } from "@malloydata/malloy";
 import { RuntimeList } from "./runtimes";
+import "./sql-eq";
+import { mkSqlEq } from "./sql-eq";
 
 // No prebuilt shared model, each test is complete.  Makes debugging easier.
 
@@ -32,31 +33,6 @@ afterAll(async () => {
   await runtimes.closeAll();
 });
 
-declare global {
-  // eslint-disable-next-line @typescript-eslint/no-namespace
-  namespace jest {
-    interface Matchers<R> {
-      isSqlEq(): R;
-    }
-  }
-}
-
-expect.extend({
-  isSqlEq: (result: Result) => {
-    const wantEq = result.data.path(0, "calc").value;
-    if (wantEq != "=") {
-      return {
-        pass: false,
-        message: () => `${wantEq}\nSQL:\n${result.sql}`,
-      };
-    }
-    return {
-      pass: true,
-      message: () => "SQL expression matched",
-    };
-  },
-});
-
 const basicTypes: Record<DialectNames, string> = {
   bigquery: `
     SELECT * FROM UNNEST([STRUCT(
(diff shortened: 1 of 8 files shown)


mtoy-googly-moogly:
@bporterfield I've made a custom matcher private to malloy-db-test/src but I don't know if the machinations I had to go through to get there are something we can support or if they just accidentally work.

@lloydtabb this is how I think we should write tests, if you like it and ben approves, I'll move all result tests to this pattern.

PR #522, opened 2022-04-29
Get schema info for unqualified PG tables using search-path

greg-finley:
Now you can `query: table('TEMP')` instead of `query: table('public.TEMP')`.

Instead of assuming `public` (which I think is the right move), an alternative would be to allow users to set a default schema, which I explored in the first commit.


greg-finley:
When the user is querying an unqualified table name, PG uses the [search_path](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH). We can use the search_path the same way to find the table's metadata.

It's a little tricky to add more tests here that wouldn't mess up a shared db connection.

greg-finley:
Did we still want to consider this one? If so, I can fix the merge conflicts.

PR #492, opened 2022-04-20
Add SQL copy button, for #451

greg-finley:
An idea for a SQL copy button. It copies the SQL without the HTML tags.

It needs some styling help (please fix if you want—definitely not my strong suit). It could switch to say "COPIED!" when it's clicked and hover to blue like the `SQL`/`HTML`/`JSON` buttons.

Screen Shot 2022-04-19 at 10 31 55 PM

Screen Shot 2022-04-19 at 10 34 34 PM

PR #523, opened 2022-04-29
Avoid selecting UI elements

whscullin:
This addresses https://github.com/looker-open-source/malloy/pull/451 in the simplest way. We could still also consider merging #492.


greg-finley:
I like this over mine, as it avoids adding clutter 👍

PR #524, opened 2022-05-01
Use uuids for PG temp tables

greg-finley:


lloydtabb:
So I'm still ruminating on how we deal with table names in Malloy.  it continues to be a problem and I'd rather not make things more complex at the moment..

greg-finley:
Sounds good.

PR #528, opened 2022-05-03
Add download capability to VSCode

christopherswenson:
Fixes: https://github.com/looker-open-source/malloy/issues/321

Adds downloads to the VSCode extension.

Current options:
* **Amount**: all rows _or_ current result set _or_ limited rows 
* **Row limit**: When amount is limited rows, the number of rows to download
* **Format**: JSON _or_ CSV

When the amount is "current result set", the data is saved to a file from memory. Otherwise, it is streamed from the DB.

For CSV format, nested tables are formatted as JSON.

Adding HTML downloads is left for a later task, as the render library does not support streaming rendering yet.

Screen Shot 2022-05-09 at 12 12 06 PM

API usage:

```ts
const writeStream = fs.createWriteStream("foo.csv");
const rowStream = runtime
  .loadModel(`source: airports is table('malloytest.airports') {}`)
  .loadQuery("query: airports -> { project: code }")
  .runStream({ rowLimit: 1 });
const csvWriter = new CSVWriter(writeStream);
await csvWriter.process(rowStream);
```

```ts
const rowStream = runtime
  .loadModel(`source: airports is table('malloytest.airports') {}`)
  .loadQuery("query: airports -> { project: code }")
  .runStream({ rowLimit: 1 });
for await (const row of rowStream) {
  console.log(JSON.stringify(row.value));
}
```


christopherswenson:
Intended use:

https://github.com/looker-open-source/malloy/pull/528/files#diff-3064a479acebd5f04467915c96d7c5d3d5f97cfabbb6d59ee894b0425423e6a3R37-R48

packages/malloy-vscode/src/extension/webviews/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
christopherswenson: Heck, thanks 😊

christopherswenson:
@whscullin Can you re-review? I think this is pretty good to go at this point. One thing I'm not super happy with is [this hack](https://github.com/looker-open-source/malloy/pull/528/files#diff-7110e75a965296a7e458e8f96cc1622e647ab988f24bbec00210e435baf6f663R452-R455), but it seems to just be a bug with using `pg-query-stream` with the connection pool.

packages/malloy-db-test/src/streaming.spec.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable no-console */
```
whscullin, line 1: No longer needed?

christopherswenson: Good catch, thanks!

packages/malloy-vscode/src/extension/webviews/query_page/DownloadButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+interface DownloadButtonProps {
+  onDownload: (options: QueryDownloadOptions) => Promise<void>;
+}
+
+export const DownloadButton: React.FC<DownloadButtonProps> = ({
+  onDownload,
+}) => {
+  const [open, setOpen] = useState(false);
+
+  // console.log(Spinner);
```
whscullin, line 30: Leftover?

whscullin (approved):

packages/malloy-vscode/src/extension/webviews/query_page/DownloadButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+    fill: #d4d6d8;
+  }
+  &:hover {
+    .hoverfill {
+      fill: rgb(240, 246, 255);
+    }
+    .primaryfill {
+      fill: #4285f4;
+    }
+  }
```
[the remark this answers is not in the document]
christopherswenson: Yep, agreed. These should definitely be swapped over when we're ready to commit to the theming. 

Seems like there are three categories of places that need to respect themes:
* UI elements like this
* Code
* Results rendering

For UI elements, the simple `var(--vscode-...)` should be fine. And for results, we just need to add some class names to the generated elements, and then we should be able to style them in the same way.

For code, the `shiki` library (used elsewhere in the repo) is capable of rendering according to VSCode themes. There were some slight problems integrating that inside the webview, which is why I didn't do it yet. I think shiki was having trouble loading either the themes or the grammars while inside the webview. We should be able to solve that either by adjusting how it's requesting those, or just moving the syntax highlighting into the extension instead of the webview. See here for the issue for this: https://github.com/looker-open-source/malloy/issues/352

PR #578, opened 2022-06-14
Track duckdb next

whscullin:
This moves us to a bleeding edge version so we get the latest patches, but I think we'll still need to manually poke the lock file periodically, it won't continuously update us.


lloydtabb (approved):
Looks good to me!

PR #581, opened 2022-06-16
Revert "Track duckdb next"

whscullin:
Reverts looker-open-source/malloy#578. I'm seeing SEGVs so maybe this is a bit too bleeding edge.

### DIFF
--- a/packages/malloy-db-duckdb/package.json
+++ b/packages/malloy-db-duckdb/package.json
@@ -12,6 +12,6 @@
     "malloyc": "ts-node ../../scripts/malloy-to-json"
   },
   "dependencies": {
-    "duckdb": "next"
+    "duckdb": "^0.3.4"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2926,10 +2926,10 @@ domutils@^2.5.2, domutils@^2.6.0, domutils@^2.7.0:
     domelementtype "^2.2.0"
     domhandler "^4.2.0"
 
-duckdb@next:
-  version "0.3.5-dev1253.0"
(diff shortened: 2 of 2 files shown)


lloydtabb:
This maybe because you need to rebuild the database.  I saw this this
morning, rebuilt and everything was fine.

whscullin:
Interesting. This was opening the vscode extension, so I'm not sure it's just the db format, although I did run into that when running tests. Maybe once my search path changes land I will try again.

whscullin:
I've tracked this down to the read only flag being used in conjunction with the in memory DB: `new Database(':memory:', OPEN_READONLY)` === 💥. Not sure if this is node specific but I will gather some details and see if there's an issue or open one.

whscullin:
OK, what's happening is that `new Database(':memory:', OPEN_READONLY)` raises an error, but since we don't have a callback registered, we never see the error. Later, when we try to open the connection on a non-initialized DB, 💥. To properly see errors, we need a callback, and constructors are bad places to do that since they are generally synchronous by nature. It sort of seems we should be splitting our db initialization into two phases, construction and initialization so we can capture async behaviors during startup better.

lloydtabb:
@christopherswenson what was the original plan in dealing with the readonly flag?  Can we just always open it up readwrite? What do you think of splitting the constructor initalizer for Connections?

lloydtabb:
Awesome detectivework @whscullin

christopherswenson:
Currently VSCode runs two simultaneous connections for every configured connection. One for the VSCode extension itself (to fetch schema and run queries) and one for the language server (to compile/find errors, add code lenses, etc.). If duckDB tries to open the same database file in RW mode twice, it crashes.

PR #583, opened 2022-06-22
Update to duckdb 0.4.0

whscullin:
Gives us search paths, readonly mode and we should be able to remove the JSON wrapper now.

### DIFF
--- a/packages/malloy-db-duckdb/src/duckdb_connection.ts
+++ b/packages/malloy-db-duckdb/src/duckdb_connection.ts
@@ -28,7 +28,7 @@ import {
   FetchSchemaAndRunStreamSimultaneously,
   StreamingConnection,
 } from "@malloydata/malloy/src/runtime_types";
-import { Database, OPEN_READONLY } from "duckdb";
+import { Database, OPEN_READONLY, OPEN_READWRITE } from "duckdb";
 import { RunSQLOptions } from "@malloydata/malloy/src/malloy";
 
 const duckDBToMalloyTypes: { [key: string]: AtomicFieldTypeInner } = {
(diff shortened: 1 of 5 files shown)


packages/malloy-db-duckdb/src/duckdb_connection.ts
```diff
@@ -43,22 +43,28 @@ const duckDBToMalloyTypes: { [key: string]: AtomicFieldTypeInner } = {
...
-    this.name = name;
-
-    // TODO temp! For now, just connect to the test database
-    this.database = new Database(databasePath, OPEN_READONLY);
+    this.database = new Database(
+      databasePath,
+      databasePath === ":memory:" ? OPEN_READWRITE : OPEN_READONLY,
+      (err) => {
+        if (err) {
+          return console.error(err);
```
whscullin, line 60: Unfortunately since this is an async callback we can't really do anything about errors that happen here, we should divide creating connections into construction and initialization so we can handle any errors that happen during setup.

christopherswenson: Makes sense. Do you think we should do that now?

christopherswenson (changes requested):
I think this looks pretty good. One thing is we also need to update the pre-built binaries in `third_party/github.com/duckdb/duckdb` and update their references in `packages/malloy-vscode/scripts/build-extension.ts`.

You can find them at `https://duckdb-node.s3.amazonaws.com/duckdb-v0.4.0-node-v93-<target>-x64.tar.gz` where <os> is `linux` or `darwin`.

packages/malloy-db-duckdb/src/duckdb_connection.ts
christopherswenson: Maybe something like this (in the `connections.ts` file in the composer demo):

```ts
class ConnectionMap implements LookupConnection<TestableConnection> {
  private readonly connections: Map<string, TestableConnection> = new Map();

  private async createConnection(connectionName: string) {
    let connection: TestableConnection;
    switch (connectionName) {
      case "bigquery":
        connection = new BigQueryConnection("bigquery");
        break;
      case "postgres":
        connection = new PostgresConnection("postgres");
        break;
      case "duckdb":
        connection = new DuckDBConnection("duckdb", ":memory:");
        break;
      default:
        throw new Error(`No connection named ${connectionName}`);
    }
    await connection.initialize();
    return connection;
  }

  async lookupConnection(
    connectionName = "bigquery"
  ): Promise<TestableConnection> {
    let connection = this.connections.get(connectionName);
    if (connection === undefined) {
      connection = await this.createConnection(connectionName);
      this.connections.set(connectionName, connection);
    }
    return connection;
  }
}

export const CONNECTIONS = new ConnectionMap();
```

christopherswenson (approved):

PR #584, opened 2022-06-22
Make extension build succeed even if duckdb is not available

christopherswenson:
We don't have node binaries for DuckDB for every machine/architecture we want to support. This change makes it possible to build the extension even if the DuckDB binary doesn't exist, by making the extension check whether it is available (through a constant which is defined during the build).


bporterfield (approved):
Looks great!

PR #585, opened 2022-06-27
Set duckdb search path to current directory

whscullin:
Also replace our URL implementation with the system URL implementation.

Currently only supports flat malloy directory structures, (data files can referenced in subdirectories, but relative malloy imports will confuse the search path) Contemplating how to get more robust search paths, possibly may require the addition of a config file to denote the root directory, or other annotation.


packages/malloy-vscode/src/common/connection_manager.ts
```diff
@@ -133,4 +137,15 @@ export class ConnectionManager {
       }
     }
   }
+
+  getConnectionManager(url: URL): LookupConnection<Connection> {
```
christopherswenson, line 141: Maybe this should be called `getConnectionMap` or something, and `.connections` should be made private.

packages/malloy-vscode/src/common/dynamic_connection_manager.ts
```diff
@@ -0,0 +1,35 @@
...
+    private workingPath: string
+  ) {}
+
+  async lookupConnection(
+    connectionName?: string | undefined
+  ): Promise<Connection> {
+    try {
+      return await this.fixedConnectionMap.lookupConnection(connectionName);
+    } catch (err) {
+      if (connectionName === "duckdb") {
```
christopherswenson, line 22: Maybe instead of making a dynamic "duckdb" connection which sets the working path corresponding to the malloy file's directory, we can just make the working path of a _configured_ connection with DuckDB backend (and any name) be that directory.

christopherswenson:
The `URL` changes all look fine to me, though there are a few failing tests that are related. I have some thoughts/questions about the connection manager stuff.

lloydtabb:
Actually, never mind.  Out of date build.  Sorry.

packages/malloy-vscode/src/common/connection_manager.ts
```diff
@@ -18,149 +18,185 @@ import { DuckDBConnection } from "@malloydata/db-duckdb";
...
+          backend: ConnectionBackend.DuckDB,
+          id: "duckdb-default",
+          isDefault: false,
+        },
+      ];
     }
-    return this._managers[dirname];
+    configs.forEach((config) => {
+      if (config.isDefault) {
+        this.configs[DEFAULT_CONFIG] = config;
```
christopherswenson, line 197: Maybe we should do a `this.defaultConfig`  or at least use a `Symbol("default-config")` just so we don't prevent someone from naming their connection "__default__" (not sure why you would do that, but someone might want to!)

christopherswenson:
Like the new organization between the `ConnectionManager` and the `DynamicConnectionLookup`. Had just a couple small comments. 👍

christopherswenson:
Looks like test failure is just lint errors.

packages/malloy-vscode/src/common/connection_manager.ts
```diff
@@ -11,38 +11,170 @@
...
+      break;
+    }
+  }
+
+  // Retain async signature for future reference
+  return Promise.resolve(connection);
+};
+
+export class DynamicConnectionLookup implements LookupConnection<Connection> {
+  connections: Record<string | symbol, Promise<Connection>> = {};
```
christopherswenson, line 115: Could even do `string | typeof DEFAULT_CONFIG` here if you wanted to get really fancy.

whscullin:
The prettier process in VSCode apparently locked up and I had to restart VSCode to get it to work again. 🤦

christopherswenson:
> The prettier process in VSCode apparently locked up and I had to restart VSCode to get it to work again. 🤦

This happens to me frustratingly often.

christopherswenson:
Looks good. Added a couple other stylistic comments. Feel free to ignore if you disagree!

PR #596, opened 2022-07-01
Make the packager work under linux and darwin

lloydtabb:


lloydtabb:
added change to another pull request.

Issue #599, opened 2022-07-03
composer: dates showing up as NaN-NaN-NaN


lloydtabb:
From

   malloy_examples/data.gov/auto_recall.malloy

Query 'By Year'

![image](https://user-images.githubusercontent.com/1093458/177050300-94f8cc61-d7f0-426f-880f-5cbd35663437.png)

Seems to work ok in VSCode.

Maybe @whscullin ?

christopherswenson:
Is this a discrepancy between BigQuery/Postgres and DuckDB behavior? That comment I added was because we weren't doing anything in particular to standardize the result type of values (in this case for dates) returned from the database. We may want to add a mapping stage which transforms the result into our desired format, using JSON-serialized dates (or something else). I believe BQ has their own Date type which behaves mostly like the JS Date type. I'm not sure about DuckDB or Postgres.

whscullin:
Possibly. DuckDB is definitely directly returning a `Date` object. I'm not sure what the other two are returning offhand.

PR #606, opened 2022-07-11
Allow sourcemaps from packages

whscullin:
Using Electron's developer tools works for `node_modules` and `src` files, but malloy packages are resolved to local paths outside of the electron app's directory. This sort of solution is [recommended](https://esbuild.github.io/api/#customizing-server-behavior) by esbuild when slightly more complex server behavior is needed.

I needed this to track down #599


bporterfield (approved):
👍 !

PR #609, opened 2022-07-12
Add postgres type "character"

uznomis:
I was testing malloy with postgres on localhost. I created a table with a "character(50)" column, and malloy doesn't seem to like this postgres type. I traced down to its mapping and added this new/missing mapping.


lloydtabb (approved):
thanks

PR #615, opened 2022-07-19
Update documentation

danesolberg:
Hi all, just trying out Malloy and updating some docs!

- Update referenced file locations in the developer documentation
- Add a Quickstart Tips section to make it simpler (at least in my mind) to try out Malloy (both BQ and Postgres)
    - Flesh out steps to connecting to BigQuery
    - Add a walkthrough for spinning up a Postgres instance in Docker, loading it with the NTSB/FAA dataset, and modifying sample NTSB models to run against Postgres

Please feel free to take/change any or all of this.  It might be better to put it in the actual documentation and link to it from the main README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,66 @@ Currently, the Malloy extension works on Mac and Linux machines.
 
 4. **Write some Malloy!**: Start with the [Quickstart](https://looker-open-source.github.io/malloy/documentation/language/basic.html). It may be helpful to check out one of the walkthroughs under Documentation below, or try some of the BigQuery [sample models](https://github.com/looker-open-source/malloy/tree/main/samples) on public datasets available on the repo before getting started.
 
+## Quickstart Tips
+
+Get up and running quickly with these steps.  Install the VS Code extension if you have not yet done so.  
+
+Malloy models in the documentation reference source data that lives in public BigQuery tables; so, if you have a Google Cloud account (or wish to make one), connecting the extension to **BigQuery** via your Google Cloud account is the fastest way to try out Malloy.  Otherwise, the sample Malloy models can be easily modified to work with a **Postgres** instance (see steps below).
+
+### BigQuery
+1. [Create](https://cloud.google.com) a Google Cloud account if you don't have one already
+2. Associate the [gcloud CLI](https://cloud.google.com/sdk/gcloud) with your account and configure gcloud to connect to Google Cloud services via a project.  The `project_id` value to enter in the CLI is found on the Google Cloud dashboard (in the screenshot below, the `project_id` would be "my-malloy-test-project")
+
+![google_cloud_project_id](https://user-images.githubusercontent.com/25882507/179831184-6206fff3-5f24-4a94-97c6-026502cf6df6.png)
+
+```bash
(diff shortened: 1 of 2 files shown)


lloydtabb:
Thanks Dane!

lloydtabb:
By the way, we work with duckdb now too.  Duckdb is natively built into Malloy. If you pull this repo, run the makefiles (to fetch the data), the models should just run.

https://github.com/lloydtabb/malloy_examples

danesolberg:
Awesome, thanks for the info.  This is even faster to test everything out.  Also it seems duckdb is supported in the marketplace version you just released, so even better!

anikaks:
Thank you @danesolberg ! I tried to [incorporate your changes](https://github.com/looker-open-source/malloy/pull/620) into the documentation for database connection; do you feel like they're still missing anything? 

I like the idea of putting the Wordle DuckDB mod into the Wordle walkthrough. What do you think about putting it [here](https://github.com/looker-open-source/malloy/tree/main/docs/_src/examples/wordle) and linking to it from the "Raw Materials" step of the [Intro](https://github.com/looker-open-source/malloy/blob/main/docs/_src/examples/wordle/wordle.md)?

The walkthrough for getting the flights data into Postgres is great! I'm thinking it should also have a place of its own. I'm thinking on the best place for it but wanted to get over the rest of this feedback; one possibility could be to put it in the [examples](https://github.com/lloydtabb/malloy_examples) repo @lloydtabb started, open to other options.

danesolberg:
Hi Anika - no problem! My goal was to continue streamlining the process of getting Malloy set up for users who want an easy introduction, so I think the way you've integrated my changes into the docs is great!  

- I'll rely on your judgement on the Postgres steps. I see that it's a WIP, so if you'd rather not highlight it then that makes sense.  I saw Postgres as an easy way to get started for people without quick access to BQ, but I think DuckDB also serves that purpose.
- The examples repo was helpful and adding the Postgres code snippets there fits well.  I would propose making the repo easier to find in that case, as I don't think I came across it until Lloyd linked it to me.
- It might be worth also creating an "Everything you need to get started quickly" page that pares down and consolidates this info for peoples' convenience.

I've found that user engagement goes up when there's a zero-effort way to try the product out, so that was my intention here! Thanks

anikaks:
Thank you!! Great feedback--I think what we'll do is move all of the examples in Lloyd's repo to this repo; I chatted with him and there's no real reason to have them separate. For now, do you want to submit a PR to add the Postgres example in [here](https://github.com/looker-open-source/malloy/tree/main/docs/_src/examples), or otherwise I can!

danesolberg:
Sure, I can take a crack at it.  I'll get the docs build working and see if I can integrate a Postgres example into it.

anikaks:
Not sure if you already got it working but feel free to put it in here, https://github.com/looker-open-source/malloy/tree/main/samples , I realized there is really no need to get the docs build working!

danesolberg:
@anikaks I opened that other PR for the Postgres example.  Should we close this one?

anikaks:
Sounds good, thanks @danesolberg !

PR #616, opened 2022-07-20
Fix issue on publish step extension on win32-x64.

gmaden:
Issue arose from merge conflict resolution (` vs ") to allow string interpolation of the duckdb versioning.


djouallah:
Thanks a lot, I confirm it works on my laptop :)

PR #627, opened 2022-07-21
Add ungroup function to parser

christopherswenson:


packages/malloy/src/lang/parse-to-ast.ts
```diff
@@ -900,6 +900,10 @@ export class MalloyToAST
     return new ast.ExprCountDistinct(this.getFieldExpr(pcx.fieldExpr()));
   }
 
+  visitExprUngroup(pcx: parse.ExprUngroupContext): ast.ExprUngroup {
+    return new ast.ExprUngroup(this.getFieldExpr(pcx.fieldExpr()));
```
mtoy-googly-moogly, line 904: If an AST object generates errors which need to be scoped to the object, this gesture copes the text extend from the parse node into the AST so that error reporting works.

If you don't do that, it walks up the AST until it finds a parent that as a known location.

```Typescript
return this.astAt(pcx, new ast.ExprUngroup(this.getFieldExpr(pcx.fieldExpr())));
```

packages/malloy/src/lang/grammar/Malloy.g4
```diff
@@ -543,6 +544,7 @@ WEEK: W E E K S?;
 WHEN: W H E N ;
 WITH: W I T H ;
 YEAR: Y E A R S?;
+UNGROUP: U N G R O U P;
```
mtoy-googly-moogly, line 547: I talked to lloyd a little and I think I like `ungrouped()` better than `ungroup()` because it is more like an adverb than a verb

packages/malloy/src/lang/ast/ast-expr.ts
```diff
@@ -738,6 +739,37 @@ export class ExprSum extends ExprAsymmetric {
   }
 }
 
+export class ExprUngroup extends ExpressionDef {
+  legalChildTypes = [FT.numberT, FT.stringT, FT.dateT, FT.timestampT];
+  elementType = "ungroup";
+  expr: ExpressionDef;
+  constructor(expr: ExpressionDef) {
```
mtoy-googly-moogly, line 746:
```
  constructor(readonly expr: ExpressionDef) {
    super({ expr });
  }
```

mtoy-googly-moogly:
Nice ... I left three comments, I think they need code changes, but I am open to conversation on any one them.

packages/malloy/src/lang/parse-to-ast.ts
christopherswenson: Yep, thanks. Totally forgot about doing that. The `visitExprCountDistinct` (directly preceding) should have that too, right?

packages/malloy/src/lang/grammar/Malloy.g4
christopherswenson: Works for me.

packages/malloy/src/lang/ast/ast-expr.ts
christopherswenson: Yep, that's fine by me.

packages/malloy/src/lang/parse-to-ast.ts
mtoy-googly-moogly: It doesn't hurt to put it everywhere.
If the node doesn't error report relative to itself, it isn't needed.

mtoy-googly-moogly:
LGTM -- Merge when ready
🐑 🇮🇹

PR #629, opened 2022-07-22
Third-party licenses available in desktop

bporterfield:
Enable a popup from the application menu to see licenses for software used in the app. Accessed via View -> View Open Source Licenses. Additionally:

- Include desktop app in license check script, and increase time between calls to github
- Add menu item in Help to access Malloy docs
- Dummy html page for when trying to view licenses when not packaged
- Flag to enable packaging of electron binary without notarization

### DIFF
--- a/demo/malloy-demo-composer/scripts/package.ts
+++ b/demo/malloy-demo-composer/scripts/package.ts
@@ -29,7 +29,7 @@ async function packageDemo(
 ) {
   doBuild(`${platform}-${architecture}`);
 
-  // include third_party_licenses.txt
+  // third_party_licenses.txt
   const licenseFilePath = path.join(outDir, thirdPartyNotices);
   await new Promise((resolve, reject) => {
     const licenseFile = fs.createWriteStream(licenseFilePath);
--- a/demo/malloy-demo-composer/scripts/package.ts
+++ b/demo/malloy-demo-composer/scripts/package.ts
@@ -25,7 +25,7 @@ const thirdPartyNotices = "third_party_notices.txt";
 async function packageDemo(
   platform: string,
   architecture: string,
-  noCodeSigning = false
+  signAndNotarize = true
 ) {
   doBuild(`${platform}-${architecture}`);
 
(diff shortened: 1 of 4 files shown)


demo/malloy-demo-composer/scripts/package.ts
```diff
@@ -12,47 +12,79 @@
...
-    if (
-      appleId === undefined ||
-      appleIdPassword === undefined ||
-      signerIdentity == undefined
-    ) {
-      console.log(
-        "Specify NOTARIZER_APPLE_ID, NOTARIZER_APPLE_ID_PASSWORD, and SIGNER_IDENTITY in environment."
-      );
-      exit(1);
+  if (!noCodeSigning) {
```
gmaden, line 52: Since this is the only usage of the new parameter, I would invert the parameter name for readability to avoid having the "not" noCodeSigning double negative when reading.

demo/malloy-demo-composer/src/electron/main.ts
```diff
@@ -126,4 +126,69 @@ async function registerIPC(): Promise<void> {
   ipcMain.handle("post:open_directory", async (_event) => {
     return await getOpenDirectory();
   });
+
+  // Native application menu
+  const template: (Electron.MenuItem | Electron.MenuItemConstructorOptions)[] =
```
gmaden, line 131: It seems unnecessary to add the full menu to the license popup (if we need a menu at all) since most of the actions are not relevant to that window.

If we do have a reason for wanting this menu then we should probably refactor it out to a shared location so we don't have to maintain in two places to keep them in sync.

demo/malloy-demo-composer/src/electron/main.ts
```diff
@@ -126,4 +126,69 @@ async function registerIPC(): Promise<void> {
...
+            label: "View Open Source Licenses",
+            click: async () => {
+              const thirdPartyWindow = new BrowserWindow({
+                width: 400,
+                height: 600,
+                webPreferences: {
+                  javascript: false,
+                },
+              });
+
```
gmaden, line 159: Having the full menu visible on the license popup seems weird since most of the actions are not relevant to the popup. Maybe we should we just hide the menu by adding something like: 
```suggestion
thirdPartyWindow.setMenu(null);
```

bporterfield: this took me a minute to understand, til I saw that this menu type is only available on linux/windows, which I guess is why I wasn't seeing it :D

demo/malloy-demo-composer/scripts/package.ts
bporterfield: 👍

PR #638, opened 2022-07-23
In SourceDefs, field references should be path arrays, not dotted strings

mtoy-googly-moogly:


mtoy-googly-moogly:
I forgot this half done piece of work, and the re-did it later, so let's shut this one down.

PR #682, opened 2022-08-01
Remove Query code lens

whscullin:
We've learned it's a confusing first time user experience that results in somewhat enigmatic error messages if you just hit enter.


whscullin:
Fixes https://github.com/looker-open-source/malloy/issues/679

PR #689, opened 2022-08-02
Make bar charts less cluttered and increase size of charts in tables

christopherswenson:
Screen Shot 2022-07-29 at 4 10 43 PM
Screen Shot 2022-07-29 at 4 10 55 PM


christopherswenson:
@anikaks Can you check this out and see if it seems like what you were looking for?

PR #704, opened 2022-08-03
Add telemetry to VSCode Extension

christopherswenson:
This PR adds GA4 telemetry to the VSCode extension. Specifically, it adds three metrics protocol events:
- `query_run` — This event occurs when a user runs a query, if that query compiles successfully. If there is a compile error in the query or model, no event occurs.
- `model_load` — This event occurs when a user opens a `.malloy` file. Opening a workspace which contains open Malloy files does not trigger this for the active file, but clicking on an existing tab for another malloy file will trigger it.
- `model_save` — This event occurs whenever a user saves a `.malloy` file.

This feature abides by two settings for disabling tracking. One is VSCode's internal [`isTelemetryEnabled`](https://code.visualstudio.com/api/references/vscode-api#env.isTelemetryEnabled) (IIRC this works whether you use `--disable-telemetry` or just set the telemetry level setting `telemetry.telemetryLevel` to "off"), and the other is a new Malloy setting `malloy.telemetry` which is on by default, but can be turned off in VSCode settings.

Currently this uses secrets available to our build system to set the GA4 API secret and measurement ID.


packages/malloy-vscode/package.json
```diff
@@ -342,6 +346,8 @@
     "@types/jsdom": "^16.2.11",
     "@types/mocha": "^8.2.2",
     "@types/prismjs": "^1.26.0",
+    "@types/node": "^12.19.8",
```
whscullin, line 349: I think this probably shouldn't be added at this point, since it will conflict with the node 16 binding.

bporterfield (approved):
👍 looks great, a few `yarn` commands & merge conflicts that will need a look

PR #705, opened 2022-08-03
plop malloy views in the main explorer, only show them if a .malloy f…

bporterfield:
…ile is opened

unclear if this is better. maybe?

### DIFF
--- a/packages/malloy-vscode/package.json
+++ b/packages/malloy-vscode/package.json
@@ -175,19 +175,22 @@
       }
     ],
     "views": {
-      "malloy": [
+      "explorer": [
         {
           "id": "malloyHelp",
           "name": "Help",
-          "type": "webview"
+          "type": "webview",
+          "when": "resourceFilename =~ /malloy/"
         },
         {
           "id": "malloySchema",
-          "name": "Schema"
+          "name": "Schema",
+          "when": "resourceFilename =~ /malloy/"
         },
         {
           "id": "malloyConnections",
-          "name": "Connections"
+          "name": "Connections",
+          "when": "resourceFilename =~ /malloy/"
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


bporterfield:
https://user-images.githubusercontent.com/108260/182693408-5ff5f38b-471d-4abb-90a6-d2c6e686ecc1.mov

christopherswenson:
Yeah this seems better to me? Definitely improves visibility at the expense of those views being organized into their own place

anikaks:
I'm in favor! It won't mess with anyone's experience with using VS Code when they aren't in malloy files so it seems unlikely to be an annoyance and we can always undo if folks complain.

whscullin:
Might be possible to have it stick around when a results tab is selected?

bporterfield:
@whscullin that seems like a great idea - unfortunately I've so far been unsuccessful trying to get a when clause that will actually work for a generated webview. Still searching....

bporterfield:
https://code.visualstudio.com/api/references/when-clause-contexts#inspect-context-keys-utility 

you...can't click on the webview and get any info :(

packages/malloy-vscode/package.json
```diff
@@ -175,19 +175,22 @@
...
     ],
     "views": {
-      "malloy": [
+      "explorer": [
         {
           "id": "malloyHelp",
           "name": "Help",
-          "type": "webview"
+          "type": "webview",
+          "when": "resourceFilename =~ /malloy/"
```
whscullin, line 183: I'd recommend `"resourceFilename =~ /\\.malloy$/"` to avoid false positives like `malloy.txt`

bporterfield: 👍

bporterfield:
@whscullin @christopherswenson now showing help/connections/schema when results panel is focused. Slightly annoying that there's a flash when clicking between .malloy file and results webview, but not that bad. LMK if y'all have any improvements.

PR #717, opened 2022-08-05
switch the method of adding query links to comment markers instead of markdown directives

christopherswenson:
Closes https://github.com/looker-open-source/malloy/issues/712


lloydtabb:
@mtoy-googly-moogly @christopherswenson Do we have a potential here for inventing a URL to data via malloy?  In Malloy a query is a reference to a malloy file and a query name/string...  By adding host, is it possible we could make something that works somewhat universally?  

   malloy://host/path-to-model/query
   
Just an idea..

christopherswenson:
AFAIK the only way to open links with custom schemes is to have an app preinstalled on your device that registers the URL scheme.

christopherswenson:
Commenting this so it isn't lost:

Note that starting a line with the HTML comment results in the entire line being considered HTML. So given:

```md
#### Paragraph with link in the middle

This is a test paragraph! Look, a thing! <!--malloy-query model="./1_names.malloy" source="names" query="by_name"--> `by_name` and I continue to talk.

#### Paragraph starting with link:

<!--malloy-query model="./1_names.malloy" source="names" query="by_name"--> `by_name` is super cool

#### With newline after comment:

<!--malloy-query model="./1_names.malloy" source="names" query="by_name"-->
`by_name` is a super cool thing
```

In GitHub and other "normal" markdown viewers, that looks like this:

#### Paragraph with link in the middle

This is a test paragraph! Look, a thing!  `by_name` and I continue to talk.

#### Paragraph starting with link:

 `by_name` is super cool

#### With newline after comment:


`by_name` is a super cool thing

Notice that the "Paragraph starting with link" one does not render the `by_name` as a code snippet.

And in the Composer markdown viewer, it looks like this:
Screen Shot 2022-10-27 at 12 05 31 PM

In the Composer, HTML is not rendered in Markdown, which is why you don't see the "Paragraph starting with link" one at all.

PR #718, opened 2022-08-05
Finish up all/exclude

mtoy-googly-moogly:
OK, we've lived with it enough and pared the feature down to a powerful minimum.

Checklist before merge:

* [x] `all(aggregateExpr, dim1)` implemented including ignore or reject of uplevel references
* [x] `all()`
* [x] `exclude(aggregateExpr, dim1, ...)`
* [x] All sample code updated and verified
* [x] Docs at least in reference section

I've written essentially a placeholder document for the language reference. I don't know how to test docs, so I am not sure I did the right thing, and really someone who isn't instantly overwhelmed with trying to imagine what query makes sense to use as an example should write some nice runnable fragments.

https://github.com/looker-open-source/malloy/blob/finish-all-exclude/docs/_src/language/ungrouped-aggregates.md


lloydtabb:
I added error checking in the runtime for now.  It was too easy to make an undetected mistake.  Once the compiler can check for errors, we should never hit my runtime errors.  We might just change them to assertions at that time...

lloydtabb:
There is some work to do to create examples and documentation but it all works.  Existing samples work.

PR #726, opened 2022-08-07
Prevent stack overflow when a query is defined with an existing name

### LINKED ISSUE #721: Stack overflow when query and dimension have same name
I watched this die as @lloydtabb and I were pair-designing, I think this would be the repro ...

```
source: flights is table('malloytest.flights') {
  query: carrier is {
    group_by: carrier
  }
}
```

mtoy-googly-moogly:
For #721 -- this doesn't fix anything though.


mtoy-googly-moogly:
This approach was wrong, abandoning.

PR #731, opened 2022-08-08
Run queries and streamed downloads in a worker

whscullin:
POC of moving queries off to threads, solves https://github.com/looker-open-source/malloy/issues/670. Still lots of cleanup:

 - [x] Restart worker on crash
 - [x] Downloads
 - [x] Proper Logging
 - [x] Debugger support
 - [x] Refactor some now duplicated code
 - [x] Multiple simultaneous queries.


lloydtabb:
This is going to be amazing.  Seems to run queries so much faster.

PR #743, opened 2022-08-15
Use vscode file buffer instead of file system

whscullin:
@anikaks pointed out that she now needed to save a file for VS Code to pick up changes, which didn't use to be the case. That's because the old code was using VS Code's notion of the file buffer, not the filesystem version. This restores that behavior by adding a message to retrieve the filed contents via VS Code.


anikaks:
I like this branch name

PR #753, opened 2022-08-18
Improve help panel

whscullin:
Play around with the contextual help - I switched to using highlights for more precision around valid click targets, and experimentally added a tooltip.

I initially found the tooltip much more readable (if harder to click) because my left bar is very crowded. But I discovered you can open a right task bar and drag panels there, and that's actually a very similar experience to other's we've created.

Screen Shot 2022-08-18 at 10 24 55 AM

Screen Shot 2022-08-18 at 10 29 53 AM

The other advantage with the tooltip is that we get markdown and consistent styling for "free", whereas formatting help in the help bar requires dragging in some variety of markdown widget.

I think we may want to switch away from highlights to use the full ast, since we have keywords that appear in multiple contexts that highlights don't disambiguate.


bporterfield (commented):
👍 love it, talking with Anika think we're happy to have the entire help panel removed and just keep the hovers.

anikaks:
what Ben said! This looks awesome.

packages/malloy-vscode/src/extension/webviews/help_page/App.tsx
```diff
@@ -11,67 +11,51 @@
  * GNU General Public License for more details.
  */
 
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import styled from "styled-components";
+import Markdown from "markdown-to-jsx";
```
christopherswenson, line 16: In other parts of the project (docs and composer), we’re using `remark` and `unified` for markdown parsing/rendering. In those cases we need direct control over how the markdown renders, but it may still be worth looking into whether we can use those instead of adding another markdown parser to our dependencies (`markdown-to-jsx`).

packages/malloy-vscode/src/extension/webviews/help_page/App.tsx
```diff
@@ -103,15 +87,6 @@ const ViewDiv = styled.div`
...
-`;
-
-const Row = styled.div`
-  display: flex;
   align-items: center;
-  justify-content: center;
-  gap: 5px;
-  width: 100%;
-
-  // TODO make the elements move to the left when the panel becomes "wide", to match the default list views
```
christopherswenson, line 116: We may want to keep this comment here as this is not fixed.

packages/malloy-vscode/src/server/hover/hover.ts
```diff
@@ -0,0 +1,46 @@
...
+  HoverParams,
+  MarkupKind,
+  Position,
+} from "vscode-languageserver/node";
+import { TextDocument } from "vscode-languageserver-textdocument";
+
+import { DocumentHighlight, Malloy } from "@malloydata/malloy";
+import { HIGHLIGHT_DOCS } from "../completions/completion_docs";
+
+const highlightForPosition = (
```
christopherswenson, line 12: This code seems to be duplicated in `extension.ts`.

packages/malloy/src/lang/parse-tree-walkers/document-highlight-walker.ts
```diff
@@ -194,6 +194,9 @@ export function passForHighlights(
       case MalloyParser.RENAME:
         register(token, HighlightType.Property.Rename, true);
         break;
+      case MalloyParser.SOURCE:
```
christopherswenson, line 197: Might be worth getting rid of the highlight for `explore` as I don't believe that's a valid keyword anymore.

christopherswenson (changes requested):
This is a great start. There are a few things I'd like us to discuss/consider before we move forward with this.

I think it’s important for us to use the full parse tree for these tips, so that we can distinguish between same-named properties in different circumstances, like `where` in a source vs a query, like you mentioned. This is probably not super urgent to fix, as I think `where` is the only one, and it basically does the same thing in either case. Also IIRC the document highlights are currently not being used for anything else at the moment. They used to be used when we used “semantic highlighting” to do syntax highlighting, but we switched to using a TextMate grammar (for [reasons](https://docs.google.com/document/d/1lb5dmSzLVj3d--wwsJhGcRuT3jf1QOZRhTnsbznhS8k/edit#bookmark=id.s72mgxu5e4f8)). I probably should’ve just deleted the highlight walker code at that time. If we decide we want to use it for this, then it’s fine to keep it in, just worth noting that I think this is the only consumer.

If we’re going to keep the sidebar help panel, I think we should make it work like the LookML one where your cursor doesn’t have to be on the keyword itself, it just has to be in the scope of that keyword — e.g. if you type `group_by:` and then start typing a field, it should still be showing the tip for `group_by`. Also I like in the LookML one that it shows a list of all the properties that you can put inside a thing. So if you’re inside a query definition, it would be nice to show that you can use `group_by` and `aggregate` etc. vs. when you’re in a source and you can use `dimension` and `measure` etc.. This may be a little tricky/different because we don't have closing delimiters for most properties, so you may only see this info for an empty query...

I am having some issues with the width of the Markdown in the sidebar view. It seems like for some keywords it’s too wide.

Screen Shot 2022-08-19 at 5 24 39 AM

It seems like we’re not in total agreement of whether the sidebar help is better or worse than the hover information, and whether we should keep it at all. I think it could be pretty useful with some tweaks, and can be helpful even in conjunction with the hovers — e.g. hovers give very brief help, sidebar gives slightly more verbose help, and docs give you everything.

christopherswenson:
Oh, another thing I noticed is that the completions don't seem to work in all cases anymore... I don't think this is new to this PR, but it seems like it broke at some point and we didn't realize.

christopherswenson:
Created a separate issue for the completions problems: https://github.com/looker-open-source/malloy/issues/756

bporterfield:
@christopherswenson I think we're all in agreement about removing help panel (@lloydtabb @anikaks agrees) 👍

anikaks:
confirmed, I'm fine with us removing the side bar in favor of the tooltip

christopherswenson:
Works for me!

whscullin (commented):
OK, I'll clean this up down to the tooltip. I was in the process of using the completion code as a template for finding the correct AST node when I got sidetracked by the same issues that @christopherswenson so nicely wrote up. I can continue to investigate that as well since I did wind up elbow deep in that as well, if you'd like.

packages/malloy-vscode/src/extension/webviews/help_page/App.tsx
whscullin: Removed sidebar code.

packages/malloy-vscode/src/extension/webview_views/help_view.ts
```diff
@@ -1,57 +0,0 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU General Public License for more details.
- */
-
-import * as vscode from "vscode";
-import * as path from "path";
-import { getWebviewHtml } from "../webviews";
-import { HelpMessageType, HelpPanelMessage } from "../message_types";
```
christopherswenson, line 17: Might be worth getting rid of these types in `message_types.ts` since they're no longer needed.

packages/malloy-vscode/src/server/completions/completion_docs.ts
```diff
@@ -11,18 +11,9 @@
...
-const DOCS_ROOT = "https://looker-open-source.github.io/malloy/documentation";
-
-const MODEL_EXPLORE_DOC = `Use \`explore\` to name, describe, and augment a data source.
-
-\`\`\`malloy
-explore: flights is table('malloy-data.faa.flights') {
-  measure: flight_count is count()
-}
-\`\`\`
+import { HighlightType } from "@malloydata/malloy";
```
christopherswenson, line 14: ^^^ Could update to 2022 in the header, but not a huge deal probably...

packages/malloy/src/lang/parse-tree-walkers/document-help-context-walker.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+    return result;
+  }
+
+  visitMalloyDocument(
+    ctx: MalloyDocumentContext
+  ): DocumentHelpContext | undefined {
+    if (this.inRange(this.rangeOf(ctx))) {
+      this.type = "model_property";
+      return this.visitChildren(ctx);
```
christopherswenson, line 97: Is it possible to just pass in the `type` into the `visitChildren` method rather than adding it as state on the walker? E.g. `return this.visitChildren(ctx, "model_property");`

packages/malloy/src/malloy.ts
```diff
@@ -785,6 +786,13 @@ export class Parse {
       (completion) => new DocumentCompletion(completion)
     );
   }
+
+  public helpContext(position: {
+    line: number;
+    character: number;
+  }): DocumentHelpContext | undefined {
+    return this.translator.helpContext(position).helpContext;
```
christopherswenson, line 794: I think it's important to have some tests for this behavior — it's pretty easy for the parser to change and totally break your walker without our realizing it. See `packages/malloy/src/lang/test/document-symbol-walker.spec.ts` for a rough example of how the tests could look.

christopherswenson (commented):
This is looking great — I've made a few small comments, and I guess it might be worth merging in main to make sure everything's good with the `npm` stuff?

packages/malloy/src/lang/parse-tree-walkers/document-help-context-walker.ts
whscullin: I couldn't figure out a way to have parameters traverse the `c.accept(this)` step above.

packages/malloy/src/malloy.ts
whscullin: Yeah, that's a good idea.

whscullin (commented):
I will add a test. This branch was rebases against main after the npm switchover so should be good to go there.

christopherswenson (approved):

PR #758, opened 2022-08-19
Add bigquery configuration parameter for query timeout

suryagaddipati:
Adds timeoutMs configuration parameter in connections for page for bigquery .


suryagaddipati:
Not quite sure why the test failed. I don't seem to have permissions to view cloud build.

packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -699,6 +704,7 @@ export class BigQueryConnection
       location: this.location,
       maximumBytesBilled:
         this.config.maximumBytesBilled || MAXIMUM_BYTES_BILLED,
+      jobTimeoutMs: Number(this.config.timeoutMs) || TIMEOUT_MS,
```
whscullin, line 707: Technically this parameter is a string, probably to avoid integer overflow, but since we using a number before I guess that also works, but if we find that people want really big timeouts, then we may want to change that.

whscullin (approved):
LGTM

Issue #707, opened 2022-08-03
now is broken in duckdb


lloydtabb:
In Auto Recalls

```
  query: by_year is {
    where: `Report Received Date` ? now - 100 days 
    group_by: recall_year is `Report Received Date`.year
    aggregate: 
      recall_count
    order_by: recall_year
  }
```

popmonkey:
should generate `CURRENT_TIMESTAMP` but Malloy is using `CURRENT_TIMESTAMP()`

popmonkey:
time for my first Malloy PR :D

PR #768, opened 2022-08-26
Enable dialect-specific `now` translations

charliermarsh:
# Summary

While playing around with DuckDB-Wasm, I ran into the issue documented in #707, which this PR attempts to resolve by enabling dialect-specific `now` translations.

The rough steps involved were:

1. Extend the `Dialect` interface to include a new `abstract sqlNow(): string` method.
2. Implement `sqlNow` on each dialect.
3. Change `ExprNow` to translate based on the current dialect.

Beyond the implementation itself, I also made some changes to `packages/malloy/src/lang/test/parse.spec.ts` to support testing the parser's ability to translate against specific dialects. (Let me know if you'd prefer those test to find a new home -- couldn't find anything similar.)

### DIFF
--- a/packages/malloy-db-test/src/datetimes.spec.ts
+++ b/packages/malloy-db-test/src/datetimes.spec.ts
@@ -201,6 +201,13 @@ runtimes.runtimeMap.forEach((runtime, databaseName) => {
       });
     });
 
+    describe(`now - ${databaseName}`, () => {
+      test(`generate the current timestamp - ${databaseName}`, async () => {
+        const eq = sqlEq("now + 30 days - 30 days", "now");
+        expect(await eq).isSqlEq();
+      });
+    });
+
     describe(`timestamp extraction - ${databaseName}`, () => {
       // 2021-02-24 03:05:06
       test(`extract second - ${databaseName}`, async () => {
--- a/packages/malloy/src/dialect/postgres.ts
+++ b/packages/malloy/src/dialect/postgres.ts
@@ -201,6 +201,10 @@ export class PostgresDialect extends Dialect {
     return `GEN_RANDOM_UUID()`;
   }
 
+  sqlNow(): string {
+    return `CURRENT_TIMESTAMP()`;
+  }
+
   sqlFieldReference(
     alias: string,
     fieldName: string,
--- a/packages/malloy/src/dialect/postgres.ts
+++ b/packages/malloy/src/dialect/postgres.ts
@@ -201,10 +201,6 @@ export class PostgresDialect extends Dialect {
     return `GEN_RANDOM_UUID()`;
   }
 
-  sqlNow(): string {
-    return `CURRENT_TIMESTAMP()`;
-  }
-
   sqlFieldReference(
     alias: string,
     fieldName: string,
@@ -268,6 +264,10 @@ export class PostgresDialect extends Dialect {
   //  and have a reaper that read comments.
   sqlCreateTableAsSelect(_tableName: string, _sql: string): string {
     throw new Error("Not implemented Yet");
+  }
+
+  sqlNow(): Expr {
+    return mkExpr`CURRENT_TIMESTAMP`;
   }
 
   sqlTrunc(sqlTime: TimeValue, units: TimestampUnit): Expr {
--- a/packages/malloy/src/lang/test/parse.spec.ts
+++ b/packages/malloy/src/lang/test/parse.spec.ts
@@ -21,18 +21,18 @@ import {
   StructDef,
 } from "../../model";
 import { makeSQLBlock } from "../../model/sql_block";
-import { ExpressionDef } from "../ast";
+import { ExpressionDef, ExprValue } from "../ast";
 import { StaticSpace } from "../field-space";
 import { DataRequestResponse } from "../parse-malloy";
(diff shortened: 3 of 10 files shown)


packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -277,11 +282,11 @@ expect.extend({
 });
 
 class BetaExpression extends Testable {
-  constructor(src: string) {
-    super(src, "justExpr");
+  constructor(src: string, dialect = "standardsql") {
```
charliermarsh, line 285: Is there a `DialectKind` union defined anywhere (or similar)?

packages/malloy/src/lang/test/test-translator.ts
```diff
@@ -32,109 +33,112 @@ export function pretty(thing: any): string {
...
+      fields: [
+        { type: "string", name: "astr" },
+        { type: "number", name: "af", numberType: "float" },
+        { type: "number", name: "ai", numberType: "integer" },
+        { type: "date", name: "ad" },
+        { type: "boolean", name: "abool" },
+        { type: "timestamp", name: "ats" },
+      ],
     },
-    structRelationship: { type: "basetable", connectionName: "bigquery" },
```
charliermarsh, line 59: Bad diff, but this is just a re-indentation. I moved the mock schema into this constructor (`makeMockSchema`) to enable parameterizing it on `dialect`.

packages/malloy/src/lang/test/test-translator.ts
```diff
@@ -182,55 +187,62 @@ export class TestTranslator extends MalloyTranslator {
    *     query: aturtle is { group_by: astr; aggregate: acount }
    *   }
    */
-  internalModel: ModelDef = {
```
charliermarsh, line 185: (This is now set in the constructor, since it too is parameterized on `dialect`.)

mtoy-googly-moogly:
This is correct but I am in the middle of changing my mind about how dialect specific code generation works and this uses the old way and not the new way. I'll write up the change requests in a review. Thanks for the PR!

charliermarsh:
@mtoy-googly-moogly - Cool, no worries, also happy to rebase as needed.

charliermarsh:
Would be nice if dialect could be set as some kind of high-level context rather than threaded through the tree :)

mtoy-googly-moogly:
There are two things which need to be addressed, but I am hitting my Friday wall. This is a great change which I am only not taking because you don't know secret undocumented things that exist only in my head, which is obviously more my problem than yours :)

Thing 1: The pattern for other DialectFragments should be copied, ExprNow should simply generate a `function: "now"` dialect fragment and the dialect code should handle those. Calling a dialect function from src/lang is something which I am trying to eliminate.

Thing 2a: This then removes the need for any dialect specific testing in src/lang, since all dialects will generate the same model (and dialect specific testing is NOT in src/lang on purpose)

Thing 2b: If you wanted a test, to see if the right code is generated, that probably goes somewhere in malloy_db_test

If you want to do this work and update the PR that would be AMAZING. If not, let's talk next week.

charliermarsh:
All makes sense! Thanks for laying it out in detail, and no worries at all -- I've been working in this codebase for, like, a few hours, so I won't pretend to have a great sense for what goes where / the direction it's headed :)

I had a feeling that something was awry (along the lines of "Thing 1") when I tried to add tests to #770 (by augmenting the parser test suite), and realized that the resolved expressions in the delta-case were a higher-order thing...

I'll give these changes a shot based on your roadmap above. Have a nice weekend!

packages/malloy/src/dialect/postgres.ts
```diff
@@ -266,6 +266,10 @@ export class PostgresDialect extends Dialect {
     throw new Error("Not implemented Yet");
   }
 
+  sqlNow(): Expr {
+    return mkExpr`CURRENT_TIMESTAMP`;
```
charliermarsh, line 270: I think this was broken for Postgres too? (We can use either `CURRENT_TIMESTAMP` or `now()`, but `CURRENT_TIMESTAMP()` throws.)

charliermarsh:
Okay @mtoy-googly-moogly - hopefully this is closer to what you described?

packages/malloy-db-test/src/datetimes.spec.ts
```diff
@@ -201,6 +201,13 @@ runtimes.runtimeMap.forEach((runtime, databaseName) => {
       });
     });
 
+    describe(`now - ${databaseName}`, () => {
+      test(`generate the current timestamp - ${databaseName}`, async () => {
+        const eq = sqlEq("now + 30 days - 30 days", "now");
+        expect(await eq).isSqlEq();
+      });
+    });
```
charliermarsh, line 209: (Kind of a bad test, but I wanted to avoid introducing a need to freeze the current time in any way. Open to suggestions.)

lloydtabb (approved):
Thanks Charlie, really nice job.

packages/malloy-db-test/src/datetimes.spec.ts
lloydtabb: Clever :)  I like it.

PR #770, opened 2022-08-26
Fix time alterations in DuckDB dialect

charliermarsh:
# Summary

Malloy currently uses `make_interval` for time intervals in DuckDB, but I think DuckDB requires a different syntax (see: [Timestamp Functions](https://duckdb.org/docs/sql/functions/timestamp)), closer to standard SQL, like:

- `CURRENT_TIMESTAMP  - INTERVAL 5 DAY`
- `CURRENT_TIMESTAMP  + INTERVAL 1 MONTH`

...and so on.

### DIFF
--- a/packages/malloy/src/dialect/duckdb.ts
+++ b/packages/malloy/src/dialect/duckdb.ts
@@ -123,16 +123,6 @@ const pgExtractionMap: Record<string, string> = {
   day_of_year: "doy",
 };
 
-const pgMakeIntervalMap: Record<string, string> = {
-  year: "years",
-  month: "months",
-  week: "weeks",
-  day: "days",
-  hour: "hours",
-  minute: "mins",
-  second: "secs",
-};
-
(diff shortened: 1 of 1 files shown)


mtoy-googly-moogly:
@charliermarsh if I understand this, without this change the code generated by Malloy for `timestamp + N units` has never worked in duckdb because it was copied from the Postgres dialect and never tested?

charliermarsh:
@mtoy-googly-moogly - Yeah, that's my understanding though I admittedly didn't look through the history. I was trying to get @lloydtabb's [political ads](https://gist.github.com/lloydtabb/c4dbd691f9fedbcc5b3d9dfa54e95fbc) demo working with DuckDB-Wasm, and was hitting an error in DuckDB due to the `make_interval` call that spun out of `where: date_range_start = now - 30 days for 30 days`.

mtoy-googly-moogly:
OK, thanks @charliermarsh -- merging.

PR #774, opened 2022-08-29
 Rough demo of duckdb-wasm

whscullin:
This is a very rough draft of a duckdb-wasm demo, there's a fair bit of work to be done. Ideally we'd be able to edit the malloy files and support reading local and arbitrary data files. There are memory issues with some queries that I'm not sure how to address (they don't happen when reading the data off disk).

The duckdb connections need a common base class, and maybe they should just actually be in one package.


lloydtabb:
Some notes about what aren't working quite right yet.

- [ ] Change the font to be smaller for the code
- [ ] Eliminate the thumbnail view from the code
- [ ] errors show up at the very bottom and are invisible sometimes (it get rendered below the bottom so you can't see message)
- [ ] Error: Invalid hint: undefined error sometimes.  Not sure what this is. (auto recall brakes dashboard)
- [ ] It would be great to have a link to the language documentation
- [ ] Upload a file...
- [ ] Query shouldn't need names 
- [ ] recognize upper_case in identifiers
- [ ] recognize backquote in identifiers

image

PR #775, opened 2022-08-30
Pin composer esbuild version

whscullin:
Composer was crashing on startup with that "default_connection is not defined" error, turns out it specifies its own esbuild version. Also re-deletes yarn.lock which slipped in during a merge.


bporterfield (approved):
Thanks, this slipped my mind with move to npm and we don't build composer in tests. We also won't build it at all soon :D but that's another story...

PR #805, opened 2022-09-14
Update readme, add ungrouped docs

anikaks:


anikaks:
Wondering if we should also add a link to the fiddle somewhere obvious in docs... header? in TOC?

PR #811, opened 2022-09-14
Reintroducing the expressjs server for composer deployment

gmaden:


demo/malloy-demo-composer/src/app/utils.ts
```diff
@@ -26,6 +26,13 @@ export function snakeToTitle(snake: string): string {
     .join(" ");
 }
 
+export function isElectron() {
```
whscullin, line 29: I'm not sure if you're doing this for comparison purposes, but just to clarify the ultimate goal is to eliminate Electron.

gmaden: It's mostly to not break things while introducing the server.
It also helps in the meantime to know what bugs are introduced by the server vs existing functionality.

I'll clean up the electron components in it's own PR.

whscullin (approved):
🚢 🇮🇹

Issue #837, opened 2022-09-29
Remove duplicate TextMate grammars


christopherswenson:
We now have three copies of the Malloy TextMate grammar that could get out of sync. Let's find some way to merge these.

PR #843, opened 2022-09-30
Remove unused grammar

whscullin:
In #837 @christopherswenson mentions we have three copies of this file, but we don't actually use this one at the moment, we use a simplified Monarch grammar, so now just two 🎉


christopherswenson:
Sweet 👍

PR #844, opened 2022-10-03
Fix interval extraction for Postgres and DuckDB

christopherswenson:
Fixes https://github.com/looker-open-source/malloy/issues/801

Initial task was to fix https://github.com/looker-open-source/malloy/issues/801, which was that `TRUNC` was used in DuckDB where `FLOOR` was intended (this was a copypasta when interval extraction was initially implemented).

Doing so revealed several other problems.

1. Date literals didn't work in DuckDB

```
query: stuff -> {
  group_by: thing is @2020-01-01
}
```

```
Catalog Error: Scalar Function with name date does not exist!
Did you mean "day"?
LINE 2:    DATE('2020-01-01') as "thing"
           ^
```

We were just using the wrong syntax for creating a date from a literal.

2. Adding durations in week or quarter units didn't work in DuckDB

```
query: stuff -> {
  group_by: thing is now + 1 week
}
```
```
Parser Error: syntax error at or near "day"
LINE 2:    ((CURRENT_TIMESTAMP)+INTERVAL 1*7 day) as "thing"
```

It seems like you can't use expressions in the `INTERVAL N DAY` syntax, which we were doing I think because documentation wasn't clear whether you could use `WEEK` and `QUARTER` as units there. ~~It seems you can.~~ I guess you cannot. Back to my hacky solution: `time + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY`...

3. Week and day interval extraction in Postgres and DuckDB did not match BigQuery behavior

In BigQuery, `days(@2022-10-03 11:59 to @2022-10-04 00:00)` would be 1, whereas in Postgres and DuckDB, it would be 0 (because 24 hours had not elapsed).

I've updated the DuckDB and Postgres implementations to match the BigQuery behavior and added some more notes to the documentation describing the desired functionality.

I also wrote some tests for the interval extraction, which also expose the other two issues.

### DIFF
--- a/packages/malloy/src/dialect/duckdb.ts
+++ b/packages/malloy/src/dialect/duckdb.ts
@@ -127,8 +127,6 @@ const inSeconds: Record<string, number> = {
   second: 1,
   minute: 60,
   hour: 3600,
-  day: 86400,
-  week: 604800,
 };
 
 export class DuckDBDialect extends Dialect {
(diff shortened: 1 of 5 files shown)


packages/malloy/src/dialect/duckdb.ts
```diff
@@ -357,16 +354,22 @@ export class DuckDBDialect extends Dialect {
...
+      repeat = 7;
     }
     const interval = mkExpr`INTERVAL ${n} ${timeframe}`;
-    return mkExpr`((${expr.value})${op}${interval})`;
+    const opInterval = mkExpr`${op} ${interval}`;
+    const repeatedOpInterval = Array.from(
+      { length: opInterval.length * repeat },
+      (_, i) => opInterval[i % opInterval.length]
+    );
+    return mkExpr`((${expr.value}))${repeatedOpInterval}`;
```
christopherswenson, line 372: @mtoy-googly-moogly This is the part that generates `time + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY + INTERVAL N DAY` to add `N` weeks... Any other ideas how to implement this?

mtoy-googly-moogly: I don't understand why this isn't 

    const interval = mkExpr`INTERVAL ${n * repeat} ${timeframe}`

mtoy-googly-moogly: ( just edited ^^^ )

christopherswenson: So it can't be literally that because `n` is an `Expr`, not a `number`. But we can do 

```ts
if (timeframe == "quarter") {
  n = mkExpr`${n}*3`;
  timeframe = "month";
} else if (timeframe == "week") {
  n = mkExpr`${n}*7`;
  timeframe = "day";
}
const interval = mkExpr`INTERVAL (${n}) ${timeframe}`
```

Which is identical to what you had in the first place, except I've added parentheses around the expression when creating the interval. DuckDB didn't seem to like `INTERVAL 1*7 DAY` that we were generating, and I couldn't find any examples of using expressions in the interval creation syntax, so I thought that it had to be a literal. It doesn't, but in this case it does have to be in parentheses.

mtoy-googly-moogly (approved):
🐑 🇮🇹

PR #854, opened 2022-10-06
Add vega dependency to render

whscullin:
If you install `@malloydata/render` from npm by itself, it doesn't include the required `vega` and `vega-lite` dependencies.


christopherswenson:
Good catch. We should be able to remove the dependency from the VSCode extension, too, right? now that it's there. https://github.com/looker-open-source/malloy/blob/main/vscode-extension/package.json#L334-L335

PR #865, opened 2022-10-12
Clean up imports and exports

whscullin:
Makes sure we're not deep importing from Malloy packages because this reflects bad API hygiene, and make sure we're only importing each package once.


whscullin:
@christopherswenson you are probably most likely to be impacted by this because of your connection work. I'm hoping to do this before publishing a new version.

christopherswenson (approved):
Seems good to me. I didn't try compiling anything to confirm things still work properly. Let me know if you want me to.

PR #867, opened 2022-10-13
Bump package versions to fix some packages

whscullin:
There's some weirdness in the way existing packages reference versions (@malloydata/render requires @malloydata/malloy 0.0.1 but the default @malloydata/malloy is now at 0.0.2), this attempts to fix that. I'm tempted to bump everything to 0.0.3 and make life simpler, however.

### DIFF
--- a/demo/malloy-demo-bq-cli/package.json
+++ b/demo/malloy-demo-bq-cli/package.json
@@ -18,7 +18,7 @@
     "run": "ts-node src/entry.ts"
   },
   "dependencies": {
-    "@malloydata/malloy": "*",
-    "@malloydata/db-bigquery": "*"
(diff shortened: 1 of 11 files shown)


christopherswenson:
I'm a bit confused — the PR comment says that "`@malloydata/render` requires `@malloydata/malloy` 0.0.1", but in the changelog it shows that you change the dependency from `*` to `0.0.3`. Where are you getting the idea that it required `0.0.1` specifically?

My understanding was that `*` just meant "go use the one that's in the repo locally". Maybe that's wrong. 

I guess regardless, if everything still works with version numbers specifically listed, that seems like a good step toward breaking up the monorepo, where the version numbers will matter.

christopherswenson (approved):

PR #872, opened 2022-10-13
Fix Google San font use

whscullin:


demo/malloy-demo-composer/public/index.html
```diff
@@ -28,14 +28,14 @@
...
       as="font"
-      type="font/woff2"
+      type="font/ttf"
       crossorigin
     />
     <style>
       @font-face {
         font-family: "Google Sans";
         font-style: normal;
-        src: url("fonts/GoogleSans[GRAD,opsz,wght]") format("ttf");
```
whscullin, line 38: The correct format is apparently "truetype", or it just figures it out if I leave it off.

PR #888, opened 2022-10-26
Convert order by index to order by name in composer

christopherswenson:
Fixes https://github.com/looker-open-source/malloy/issues/883

In https://github.com/looker-open-source/malloy/issues/883, there was an issue when removing fields that are used in an order by which referenced the field by index instead of by name. Specifically, the code which removes the order by on a field when the field is removed did not account for order bys which referenced fields by index. This meant that if you had `order_by: 3` and you removed a field, you could change the meaning of the order by unintentionally, or invalidate it entirely. 

Rather than handle all of the edge cases (such as adding a field before the order by field, which would require incrementing the index), this PR just adds a conversion step when loading a query (either via "Load Query" or "Duplicate") which replaces order by index references with the name of the field.

### DIFF
--- a/demo/malloy-demo-composer/src/core/query.ts
+++ b/demo/malloy-demo-composer/src/core/query.ts
@@ -628,7 +628,26 @@ export class QueryBuilder extends SourceUtils {
     if (definition === undefined) {
       throw new Error("Field is not defined..");
     }
-    stage.fields[fieldIndex] = JSON.parse(JSON.stringify(definition));
+    const definitionCopy: FieldDef = JSON.parse(JSON.stringify(definition));
+    if (definitionCopy.type === "turtle") {
+      for (const stage of definitionCopy.pipeline) {
+        if (stage.type === "reduce" && stage.orderBy) {
+          const newOrderBy = stage.orderBy
+            ? stage.orderBy.map((orderBy) => {
+                if (typeof orderBy.field === "string") {
+                  return { ...orderBy };
+                } else {
+                  const field = stage.fields[orderBy.field - 1];
+                  return { ...orderBy, field: this.nameOf(field) };
+                }
+              })
+            : undefined;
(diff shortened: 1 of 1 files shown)


demo/malloy-demo-composer/src/core/query.ts
```diff
@@ -262,7 +262,17 @@ export class QueryBuilder extends SourceUtils {
           existingStage.limit = stage.limit;
         }
         if (stage.orderBy) {
-          existingStage.orderBy = stage.orderBy;
+          const newOrderBy = stage.orderBy
+            ? stage.orderBy.map((orderBy) => {
```
whscullin, line 266: This is already behind `if (stage.orderBy)` so the ternary is redundant.

demo/malloy-demo-composer/src/core/query.ts
```diff
@@ -618,7 +628,26 @@ export class QueryBuilder extends SourceUtils {
     if (definition === undefined) {
       throw new Error("Field is not defined..");
     }
-    stage.fields[fieldIndex] = JSON.parse(JSON.stringify(definition));
+    const definitionCopy: FieldDef = JSON.parse(JSON.stringify(definition));
+    if (definitionCopy.type === "turtle") {
+      for (const stage of definitionCopy.pipeline) {
+        if (stage.type === "reduce" && stage.orderBy) {
+          const newOrderBy = stage.orderBy
```
whscullin, line 635: ditto.

whscullin:
Couple minor things but otherwise LGTM

anikaks:
LGTM

anikaks (approved):

whscullin (approved):

christopherswenson:
Checks are failing due to the `Expected end of input but got string literal "GEOGRAPHY"` issue, which is unrelated to this PR. Going to merge.

PR #897, opened 2022-10-28
Fix row limit config, add row limit support to DuckDB.

whscullin:
Fixes a couple issues:

* The logic to pass updated options to the worker was looking at changes to connection configs, but row limit is a malloy level config option.
* The logic to retrieve the row limit was never added to duckdb connections, so it was always defaulting to 10.


lloydtabb:
If you merge main, this should work now.

PR #904, opened 2022-10-31
Convert test to use describe.each()

whscullin:
The way we name tests using template strings breaks VSCode's inline test debugging, if we want it to work again we need to switch to naming scheme that uses describe blocks for variable names so that test names are static. 

Here's expr.spec.ts converted to do that. If this seems reasonable we can switch other tests, as well.


christopherswenson:
This looks good to me. I can try it out tomorrow.

mtoy-googly-moogly:
actually no ... stop ... let me look

mtoy-googly-moogly:
Yeah I want to talk about this ... I'm not in love with how we run tests across databases, but I think this is worse.

mtoy-googly-moogly (approved):
OK, I've merged main into this PR, changed the thing I did in main to do it the way @whscullin does it here. This is ready to go.

Apologies again for not reading this the first time and just having a canned reaction to the word "testing".

PR #909, opened 2022-11-03
Multili stage index queries

lloydtabb:
Indexes on sources with many one to many joins can have very large cross products.  We break the index into multiple stages.  

Stuff left to do..

- [ ] Sampling in BigQuery needs a temorary table
- [ ] Need to figure out the minimum number of stages we should generate for performance tuning
- [ ] Deep wildcarding (**) seems to be broken.


lloydtabb:
Going to land these changes in preperation for a better duckdb implementation of unnest.  Indexing still needs better limiting, but this is part of the solution.

PR #974, opened 2022-12-25
Lay the ground work for Analytic functions in the Parser and Code generator.

lloydtabb:
In expressions, type can now be
"scalar"
"aggregate"
"analytic"

### DIFF
--- a/packages/malloy/src/lang/ast/ast-types.ts
+++ b/packages/malloy/src/lang/ast/ast-types.ts
@@ -18,6 +18,7 @@ import {
   TimestampUnit,
   Expr,
   TimeFieldType,
+  ExpressionType,
 } from "../../model/malloy_types";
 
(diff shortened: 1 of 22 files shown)


packages/malloy/src/lang/ast/ast-types.ts
```diff
@@ -100,8 +101,8 @@ export class FT {
     const strings = types.map((type) => {
       if (type) {
         let inspected: string = type.dataType;
-        if (type.aggregate) {
-          inspected = `aggregate ${inspected}`;
+        if (type.expressionType != "scalar") {
+          inspected = `expressionType ${type.expressionType}`;
```
mtoy-googly-moogly, line 105: OK, I think this is wrong.

for an aggregate timestamp, the old code would print `aggregate timestamp`
this code will print `expressionType aggregate`

PR #977, opened 2023-01-01
Add example for multiple criteria in Pick

nryberg:
Some CASE statements depend on multiple criteria and fields, this lays it out.


nryberg:
I created a separate pull request for example links corrections.

PR #980, opened 2023-01-02
Corrects path to dot malloy files for examples

nryberg:
The current "this model _here_" link resolves to a non-existent GitHub location resulting in 404 errors.  The new links correct this.


lloydtabb:
Looks good.  Thanks.  These changes will appear next time we push docs (a manual process unfortunately)

PR #984, opened 2023-01-06
Duckdb wasm jest

whscullin:
Unfortunately had to back nodejs duckdb to 0.5.1 so the test files were compatible, so this is waiting on a new duckdb-wasm for now.


packages/malloy-db-duckdb/src/duckdb_wasm_connection.ts
```diff
@@ -74,15 +75,19 @@ const unwrapTable = (table: Table): QueryDataRow[] => {
   return table.toArray().map(unwrapRow);
 };
 
+const isNode = () => typeof navigator === "undefined";
+const DIST = "./node_modules/@duckdb/duckdb-wasm/dist/";
```
whscullin, line 79: Need to not hardcode this, too.

PR #989, opened 2023-01-11
Maden/cleanup

gmaden:
Just use builtin crypto library instead

### DIFF
--- a/packages/malloy/src/lang/grammar/build_parser.js
+++ b/packages/malloy/src/lang/grammar/build_parser.js
@@ -45,7 +45,10 @@ function run(cmd) {
 }
 
 process.chdir(path.join(langSrc, "grammar"));
-const versionDigest = crypto.createHash('md5').update(digestSrcFiles.map((fn) => readFileSync(fn, "utf-8")).join("")).digest('hex');
+const versionDigest = crypto
+  .createHash('md5')
+  .update(digestSrcFiles.map((fn) => readFileSync(fn, "utf-8")).join(""))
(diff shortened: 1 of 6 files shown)


packages/malloy/src/lang/grammar/build_parser.js
```diff
@@ -18,7 +18,7 @@
 /* eslint-disable no-console */
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { readFileSync, writeFileSync, existsSync, rmSync } = require("fs");
-const md5 = require("md5");
+const crypto = require('crypto');
```
whscullin, line 21: Unfortunately I don't think this is cross platform, the browser crypto implementation is different, and I'm not sure the browserify-crypto package will fit our needs.

whscullin: I guess I flagged the wrong instance of this, this is fine, it's the one's internal to the malloy library, like https://github.com/malloydata/malloy/pull/989/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R71 that will be tricky cross platform.

gmaden: Yeah I knew what you meant.
Unfortunately it looks like dynamically loading libraries is async, as well as the browser crypto.subtle libraries, so that blocks making it easy to just use whichever is available at runtime without spreading asyc/await everywhere...

Issue #952, opened 2022-12-04
Docs Barn Raising


lloydtabb:
I'd like to take some time and all take a pass through the documentation.  There is some cruft, multiple documents that point to the same thing.  Some bad instructions (connections for example) and we should be using DuckDB instead of BigQuery for examples for a number of reasons.  Here is a list of things that I think should happen.

- [ ] Move to a new and separate repository with its own build scripts
- [ ] It should deploy new docs when a commit is made to 'main' 
- [ ] We should swtich from using BigQuery for examples to DuckDB
- [ ] Examples should be included in docs, not read from malloy-samples (this causes regular breakage)
- [ ] Connection documentation is pain wrong, should talk about VSCode, Composer, NPM etc.
- [ ] Examples should be trimmed
- [ ] Verify that all the syntax we are using is consistent and what we'd like it to be.
- [ ] Review the introduction documents and make some hard choices 

Bonus points

- [ ] Make the examples live or fiddlable (open in Fiddle)

nryberg:
Happy to help migrate from BigQuery to DuckDB - getting lost in the weeds for data sourcing - if you can point me in the right direction?

whscullin:
https://github.com/malloydata/malloy/issues/952 and the new repo https://github.com/malloydata/malloy.github.io address 1 and 2. I still need to replace the existing docs pages with redirects. At this point nothing structural has changed about the docs.

whscullin:
https://github.com/malloydata/malloy/pull/992 addresses redirecting old links.

PR #991, opened 2023-01-13
Move docs

whscullin:
Remove files that have moved to https://github.com/malloydata.github.io and as many packages and scripts as I could find. Also removed the one test that compiled our samples, which probably needs to live somewhere, but not clear to me what our approach should be.

Step one of: https://github.com/malloydata/malloy/issues/952

build-docs script will fail until I delete that trigger.


whscullin:
https://github.com/malloydata/malloy-samples/pull/6 reintroduces tests for the sample files.

PR #1002, opened 2023-01-25
Removing circular refs in html_view and container ts files

srschandan:


packages/malloy-render/src/html/html_view.ts
```diff
@@ -165,12 +165,18 @@ export function makeRenderer(
...
       document,
       isContainer(field),
       options
     );
+    // return ContainerRenderer.make(
+    //   HTMLDashboardRenderer,
+    //   document,
+    //   isContainer(field),
+    //   options
+    // );
```
whscullin, line 179: I think this is probably stray.

whscullin (approved):
Seems reasonable to me.

packages/malloy-render/src/html/html_view.ts
srschandan: Ack.

Issue #1004, opened 2023-01-30
Compound having expressions are broken


lloydtabb:
```
source: airports is table('duckdb:airports.parquet') {
  measure: airport_count is count()
  measure: county_count is count(distinct county)

  query: by_state is {
    having: airport_count > 20 or county_count > 100
    group_by: state
    aggregate: airport_count
  }
}
```

mtoy-googly-moogly:
Yeah the type checking engine inside the expression compiler is pretty sketchy. Almost certainly got broken by introducing window functions. Probably easy to fix.

I'd wait til @gmaden merges the re-org of lang/ though.

christopherswenson:
Oh, well here's a fix for this... Can wait for the re-org... https://github.com/malloydata/malloy/pull/1006

PR #1006, opened 2023-01-31
Fix legal child types for logical binary boolean operators

christopherswenson:
Looks like the format for legal child types was changed, but this case was not updated to match the new format.

The result was that you could not have an aggregate expression consisting of a boolean and/or of other aggregates.

Fixes https://github.com/malloydata/malloy/issues/1004


mtoy-googly-moogly (changes requested):
That test should go in parse spec in lang -- the `malloytest.airports` model is available in the lang test framework

mtoy-googly-moogly:
The test could be as simple as `test("compound having still works", modelOK(`SOURCE CODE`))`

mtoy-googly-moogly (commented):
LGTM

PR #1007, opened 2023-01-31
Break apart and sort lang package.

gmaden:
Initial explosion of the lang package into minimal files ~1 class per file. Circular dependencies have been removed where possible and dependencies have been explicitly defined.

Classes have been naively sorted so expect some changes in the future to get a more intuitive structure for the code.

There remain a few files with some class reference circles, but those will require more thought and care to tease apart (if indeed that is what we want).


whscullin:
Do we want to standardize on kabob case vs snake case for file names? Our existing convention has been snake case.

mtoy-googly-moogly (approved):
@gmaden and I talked this through and the best plan forward is to accept the PR so that the restructured code can happen, and then to consider possible changes as a followup piece of work.

mtoy-googly-moogly:
@whscullin i don't think there is a consistent standard. if people love snake case i can adjust. i prefer not, which is why things are the way they are, and have been in lang since the beginning of time, but i recognize i am basically an ignorable relic of a long forgotten past and if the cool kids want snake case that is ok with me

PR #1019, opened 2023-02-10
Allow unsupported types in schemas

mtoy-googly-moogly:
OId behavior was the create a string expression with the value "this field has an unsupported type in it"

This should allow unsupported types to be passed to functions, and to be type cast. Beyond that they would still be unsupported.

* `query: x -> { project: * }` with unsupported fields works
* `query: x -> { group_by: unsupportedTypeField }` also works
* `query: x -> { where: unsupportedTypeField != NULL; ... }` also works
* `source: { dimension: x is` _any other expression with unsupported fields_ `}` will error
* `source: { dimension: xOk is(` _unsupported expression_ `)::__any malloy_type__ }` will compile, it may fail when you run the SQL (like you can't cast a GEOGRAPHY to a STRING in BigQuery)

- [x] Add an "unsupported" type in the schema
- [x] Don't allow unsupported field expressions in compiler
- [x] Tests updated
- [ ] Figure out what to do about the existing hack  #1018
- [ ] Make a plan to fix composer(s) and VSCode for unsupported data in the results.

### DIFF
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -519,7 +519,8 @@ export class BigQueryConnection
         structDef.fields.push(innerStructDef);
       } else {
         const malloyType = this.bqToMalloyTypes[type] || {
-          "type": "unsupported"
+          "type": "unsupported",
+          "rawType": type.toLowerCase()
         };
         structDef.fields.push({ name, ...malloyType } as FieldTypeDef);
       }
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -518,7 +518,9 @@ export class BigQueryConnection
         this.addFieldsToStructDef(innerStructDef, field);
         structDef.fields.push(innerStructDef);
       } else {
-        const malloyType = this.bqToMalloyTypes[type] || "unsupported";
+        const malloyType = this.bqToMalloyTypes[type] || {
+          "type": "unsupported"
+        };
         structDef.fields.push({ name, ...malloyType } as FieldTypeDef);
       }
     }
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -518,16 +518,8 @@ export class BigQueryConnection
         this.addFieldsToStructDef(innerStructDef, field);
         structDef.fields.push(innerStructDef);
       } else {
-        const malloyType = this.bqToMalloyTypes[type];
-        if (malloyType) {
-          structDef.fields.push({ name, ...malloyType } as FieldTypeDef);
-        } else {
-          structDef.fields.push({
-            name,
-            "type": "string",
-            "e": [`'BigQuery type "${type}" not supported by Malloy'`]
-          });
-        }
+        const malloyType = this.bqToMalloyTypes[type] || "unsupported";
+        structDef.fields.push({ name, ...malloyType } as FieldTypeDef);
       }
     }
   }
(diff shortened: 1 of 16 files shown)


lloydtabb:
In General, I'm really supportive of this pattern.  

It would be really nice if you could compare unsupported types  `unsupportedTypeField = unsupportedTypeField`

I think there is going to be a bunch of work in Composer and VSCode extension to make these fields  work as output and I think we should disallowing them in results for now.  

I think for now, we should never return an unsupported field in a query result.  Disallow:

  `project: unsupportedTypeField`  // error
  
and

  `group_by: unsupportedTypeField`  // error

You can always make a string cast of the unsupported field type to get at the data if you must..

mtoy-googly-moogly:
@lloydtabb

1) We can't compare unsupported types because in the compiler we don't know what type it is, because that was the least amount of work. it might take me a day or more to wrangle adding the actual datatype to FieldDef ands fixing the hundreds of places that break when I do that.
2) Allowing unsupported types in results lets you write transformation pipelines that carry along unsupported data out to the end. We just need to make the extension and the composer show an ❌ or something.

lloydtabb:
Mostly unsupported types are some binary representation of something.

Let's take UUID.  In Postgres this is a 128 bit binary thing.  The thing it is use for is generally primary/foreign keys.  I think we need to support equality, so we can do joins.  I don't think any other expressions are necessary.  Casting to strings will defeat indexes on Postgres and we won't be able to use them.

I'm worried about API access to results with unknown type columns.  We return data as JSON and unknown types are going to be return from queries from the API even if VSCode Extension and Composer don't render them.  This code will be fragile.  Forcing them to return these results as strings will guarantee that the dependent code  continues to work.

christopherswenson:
I am okay with all this, so if we can get to a place where Lloyd is happy, this gets a 👍🏼 from me.

Some other thoughts/questions: 

I assume that even if in the future we add an `unknown` and/or `any` type (esp for use in parameterized things/functions), that would still be different from `unsupported`? Would we ever want to allow a function with a signature like `(x: string) -> unsupported` (e.g. a wrapper around BQ's `net.ip_from_string`. In that case do we (eventually) want the user to be able to declare something to be of type `unsupported`, and if so, is "unsupported" the right name, or is there a better one? Would a user want to say `unsupported("IP")` to somehow mark that they knew it to be specifically the `IP` type so that we can typecheck equality comparisons on unsupported types? Or similarly something like `type: bq_ip is unsupported("ip")` ... `function make_ip(x::string)::bq_ip is ...`?

mtoy-googly-moogly:
@christopherswenson: Why it is "unsupported" and not "unknown"

We do know the type (though we don't store in the structdef atm), we just dont' support that type.

Unsupported is better. In the far future world where someone who is better at compilers than I am is maintaining the compiler type algebra checker and we have "unknown" and "any", there won't be any unsupported types. Unsupported is a stop gap because we don't have enough people to just go support all the types.

christopherswenson:
I agree unsupported != unknown.

I'm not sure I agree that Malloy will eventually support all types that the database supports. It seems like there might always be some random type we don't know how to think about and don't feel like implementing, so we make it "unsupported."

I was wondering if a user ever needs to type in "unsupported" into their code, to specify the type as specifically unsupported. (For instance, if they were to write a function signature that returned an unsupported type, how would they write that?) And if so, is "unsupported" the best name? Or is there something better? E.g. "custom" or "db_type" or even some way of including the name of the db type: "unsupported<ip>". (I'm not arguing that any of these are better, I'm just asking us to think about it)

mtoy-googly-moogly:
I think "I can write type signatures" world, it is an interesting question how you would write the signature for a type that Malloy does not support.

We could worry about that on the future day when that PR appears. In this PR the word for these unsupported types is not a user visible entity except as it appears in error text.

mtoy-googly-moogly:
I don't like `db_type` or `custom` for a number of reasons. Maybe there is a 

```
  sqlType: "geography"
```

property for a signature that returns an unsupported type, with no mention of support status, so that if we ever DO support geography types, the sagnature is still valid.

PR #1060, opened 2023-03-14
Function overhaul

christopherswenson:
**This PR is extremely experimental, and there are several key design decisions which have not been finalized; everything here is subject to change**
 
### Things left to do (not an exhaustive list):
- [x] Write docs
- [x] Decide whether we're keeping the "raw import" idea (the answer was "no")
- [ ] Add a "strictness" flag for whether unknown functions require `!`
- [x] Implement the remaining functions 
- [x] Write tests
- [x] Remove `format` ~~and `ifnull`~~
- [ ] Decide whether it's okay how `**` just explodes
- [x] Find a way to make this fail to typecheck (because a measure cannot be used in a `project`):

```
query: aircraft -> {
  project: state
  calculate: lag_count is lag(count())
}
```
- [x] Figure out a more robust way to access "output" fields in a refinement
- [x] Find a way to make this succeed typechecking (because a measure can be used as an "output" field in a `calculate` in a `group_by`):
```
query: aircraft -> {
  group_by: state
  calculate: lag_count is lag(aircraft_count)
}
```
I made this work by adjusting the `evalSpace` for looked-up `ExprIdReference`s to be 'output'. See below for why this isn't super ideal.
- [x] Find a way to make this error during typechecking (rather than erroring in the compiler, because `seats` is in the output space and the compiler doesn't know what to do about that when producing an aggregate):
```
query: aircraft -> {
  group_by: aircraft_models.seats
  calculate: a is lag(seats.sum()) 
} 
```
Update: I made this compile, where `seats.sum()` is equivalent to `aircraft_models.seats.sum()` NOT `sum(aircraft_models.seats)`, which seems logical to me.
- [x] Fix `project: *` to somehow add all the fields to the output space
```
query: aircraft -> {
  project: *
  order_by: year_built // Now: Errors because year_built is "not in the output space"
}
```
- [x] I seem to have broken `accept` and `except` somehow
- [x] Fix issue where the following query fails (in BigQuery) due to the field `births` not being directly grouped-by:
```
query: state_facts -> {
  group_by: 
    state,
    births_ballpark is ceil(births / 1000000) * 1000000
  order_by: births_ballpark desc
  calculate: births_ballpark_rank is rank()
}
```
- [x] Handle the case where you have a function name which conflicts with a Malloy keyword and you want to call it "raw": `` `date`!date('2023-05-24')``.

### Things which don't work perfectly but they may be good enough for now:

- [ ] Make symmetric aggregates work for custom aggregate functions in BQ
- [ ] You can't reference "input" fields in a `calculate` in a `project` (this should work, but currently doesn't because we find "state" in the input space and `lag` expects its argument to be in the output space.
```
query: aircraft -> {
  project: id 
  calculate: prev_state is lag(state)
}
```
- [ ] Make it so you can order by analytic fields. Currently any `order by` of an analytic field is illegal. This is because analytic fields currently only ever use their stage's order by for ordering the window. A window's `order by` clearly cannot reference itself, so any self-referencing `order by` is removed, and then there's no `order by` left over. In the future we need to A) make a way to require that an analytic field has an `order by`, and know how to inherit the outer one if none is specified.
```
query: aircraft -> {
  group_by: tail_num
  calculate: prev_tail_num is lag(tail_num) 
  order_by: prev_tail_num
} 
```
- [ ] Understand what this query means:
```
query: aircraft -> {
  declare: a is 1
  join_one: numbers with a
  group_by: numbers // What does it mean to group by a join
  calculate: x is lag(numbers) // Error: no matching overload for lag(struct), not sure if this is correct
}
```
- [ ] Aggregate expressions are currently considered "output-like", whereas aggregate fields are considered "input-like." This is a bad inconsistency, because you should be able to do {following query}. Also, aggregates I think are really more "input-like." Essentially we need to just better understand what output-input means. *UPDATE*: I made it so when you reference an aggregate, it becomes "output-like." Which solves the inconsistency for now, but it's still bad (probably) that aggregates are considered "output-like" at all.
```
query: aircraft -> {
  group_by: state
  calculate: x is lag(aircraft_count)
}
```

### Other issues found while working on this, which are not addressed here:

- In a `nest` where there are fields in the outer query and in the inner query with the same name, `exclude` currently picks the outer one, where it really should pick the closest (inner) one. Here in the last `aggregate`, `state` should refer to the one from two lines above, not the one at the beginning of the query. So in this case, the inner query's `aircraft_count` and `all_aircraft_count` should have the same value.
```
query: aircraft -> {
  group_by: state
  aggregate: aircraft_count
  aggregate: all_aircraft_count is exclude(aircraft_count, state)
  nest: my_turtle is {
    group_by: state
    aggregate: aircraft_count
    aggregate: all_aircraft_count is exclude(aircraft_count, state)
  }
}
```
- [ ] Figure out number subtypes: currently `round(1.5, 2.1)` is not a translation error, even though the dialect will return an error like "precision must be an integer".
- [ ] This query doesn't work for reasons I don't understand.
```
// Runtime error:
// Binder Error: column "seats" must appear in the GROUP BY clause or must be part of an aggregate function.
// Either add it to the GROUP BY list, or use "ANY_VALUE(seats)" if the exact value of "seats" is not important.
query: aircraft -> {
  group_by: aircraft_models.seats
  aggregate: percent_of_max is aircraft_models.seats / all(max(aircraft_models.seats))
}
```
Is it just a general rule that you can't operate on a `scalar` with an `ungrouped_aggregate`?
- [ ] `top: 10 by state` doesn't seem to work at all — the ordering seems to still happen by the default ordering.


packages/malloy/src/lang/ast/field-space/project-field-space.ts
```diff
@@ -33,23 +35,31 @@ export class ProjectFieldSpace extends QuerySpace {
...
+      type = qd.type;
+      expressionType = qd.type === 'turtle' ? undefined : qd.expressionType;
+    }
+    if (type === 'turtle' || expressionIsAggregate(expressionType)) {
+      // We don't need to log here, because an error should have already been logged.
+      return false;
+    }
+    // TODO it would be really nice to attach this error message to the specific field,
+    // rather than the whole query.
+    if (expressionType === 'aggregate_analytic') {
```
mtoy-googly-moogly, line 60: expressionIsAnalytic && expressionIsAggregate ?

christopherswenson: I’m not sure what to do about this one. `expressionIsAnalytic()` means currently scalar analytic or aggregate analytic. And `expressionIsAggregate()` means aggregate or ungrouped aggregate.

Maybe I add `expressionInvolvesAggregate()` which means aggregate, ungrouped, aggregate, and aggregate analytic, then do `expressionIsAnalytic() && expressionInvovesAggregate()` as you suggest.

christopherswenson: And `expressionIsAggregate()` shouldn’t include aggregate analytic because you can’t use `aggregate: x is lag(count())`

packages/malloy/src/lang/ast/field-space/reference-field.ts
```diff
@@ -39,10 +40,14 @@ export class ReferenceField extends SpaceField {
...
+    this.res = check;
     return this.fieldRef.refString;
   }
 
   typeDesc(): TypeDesc {
-    return {dataType: 'unknown', expressionType: 'scalar'};
+    if (this.res !== undefined && this.res.found) {
+      return this.res.found.typeDesc();
+    }
+    return {dataType: 'unknown', expressionType: 'scalar', evalSpace: 'input'};
```
christopherswenson, line 51: Remind myself why the `typeDesc` can't just trigger a `getQueryFieldDef`...?

packages/malloy/src/lang/ast/field-space/static-space.ts
```diff
@@ -98,15 +98,16 @@ export class StaticSpace implements FieldSpace {
     delete this.map[name];
   }
 
-  protected entry(name: string): SpaceEntry | undefined {
+  // TODO this was protected
+  entry(name: string): SpaceEntry | undefined {
```
christopherswenson, line 102: Ask MToy about this

christopherswenson (commented):
Got as far as `packages/malloy/src/lang/ast/query-items/field-declaration.ts`... more left to look over...

packages/malloy/src/lang/ast/types/malloy-element.ts
```diff
@@ -333,6 +334,10 @@ export class RunList extends ListOf<DocStatement> {
   executeList(doc: Document): ModelDataRequest {
     while (this.execCursor < this.elements.length) {
       if (doc.errorsExist()) {
+        // TODO make a better way to stop cascading errors -- this way means that if there are
+        // actual different errors in multiple places, we only see the first one. A better way
+        // might be to make things that have error pass all kinds of typechecks, which prevents
+        // the real cascade.
```
christopherswenson, line 340: I think this is a separate project — If instead we just have the `unknown` type (maybe rename to `error` type) be accepted everywhere instead of rejected everywhere, I think we remove a ton of cascading for cases that are real cascading issues without removing totally unrelated error messages.

packages/malloy/src/lang/ast/types/malloy-element.ts
```diff
@@ -438,10 +449,14 @@ export class Document extends MalloyElement implements NameSpace {
...
   getEntry(str: string): ModelEntry {
-    return this.documentModel[str];
+    return this.globalNameSpace.getEntry(str) ?? this.documentModel[str];
   }
 
   setEntry(str: string, ent: ModelEntry): void {
+    // TODO this error message is going to be in the wrong place everywhere...
+    if (this.globalNameSpace.getEntry(str) !== undefined) {
+      this.log(`Cannot redefine '${str}', which is in global namespace`);
+    }
```
christopherswenson, line 459: I think I 'fixed' this — this error message is a nice fallback error message, but places that actually do a redefine should check manually to see if it exists, so that it can attach the error to the right AST element.

christopherswenson: I think I made a comment above about this, not realizing that that was the point. I'll check back on that later.

packages/malloy/src/lang/grammar/MalloyParser.g4
```diff
@@ -445,10 +447,14 @@ fieldCollection
...
 collectionMember
-  : fieldPath                         # nameMember
-  | (fieldPath DOT)? (STAR|STARSTAR)  # wildMember
-  | fieldDef                          # newMember
+  : fieldPath
+  | collectionWildCard
+  | fieldDef
   ;
 
 fieldPath
```
christopherswenson, line 460: Michael may want to look at the changes to the grammar.

packages/malloy/src/lang/malloy-to-ast.ts
```diff
@@ -223,23 +224,10 @@ export class MalloyToAST
   }
 
   protected getFieldDefs(
-    cxList: ParserRuleContext[],
-    isAgg?: boolean
+    cxList: parse.FieldDefContext[],
+    makeFieldDef: ast.FieldDeclarationConstructor
```
christopherswenson, line 228: Changed the way that field declaration "subtypes" work. Now instead of creating a generic `FieldDeclaration` and then attaching subtype info, we just construct different subclasses of `FieldDeclaration` depending on the place the field declaration occurs. This allows us to do nice typechecking of what kinds of fields are used in different kinds of definitions.

packages/malloy/src/lang/malloy-to-ast.ts
```diff
@@ -624,41 +618,112 @@ export class MalloyToAST
...
     return this.onlyQueryRefs(itemList);
   }
 
   visitAggregateStatement(pcx: parse.AggregateStatementContext): ast.Aggregate {
-    return new ast.Aggregate(this.getQueryItems(pcx.queryFieldList()));
+    return new ast.Aggregate(
+      this.getQueryItems(
+        pcx.queryFieldList(),
+        ast.AggregateFieldDeclaration,
+        ast.AggregateFieldReference
```
christopherswenson, line 655: Same thing essentially with field references.

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -303,11 +303,494 @@ describe('model statements', () => {
...
+          "Cannot define 'y', value has unknown type"
+        );
+      });
+      test('lag can check that other args are constant', () => {
+        expect(`query: a -> {
+          group_by: x is 1
+          calculate: s is lag(x, x)
+        }`).compileToFailWith(
+          // TODO improve this error message
+          "Parameter 2 ('offset') of lag must be literal or constant, but received output"
```
christopherswenson, line 636: We might want to improve this error message, because "output" is not an obvious term.

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -303,11 +303,494 @@ describe('model statements', () => {
...
+        modelOK(`query: a -> {
+          group_by: x is 1
+          calculate: s is lag(x, 2)
+        }`)
+      );
+      test('cannot name top level objects same as functions', () => {
+        expect(
+          markSource`query: ${'concat is a -> { group_by: x is 1 }'}`
+        ).compileToFailWith(
+          // TODO improve this error message
```
christopherswenson, line 650: Might want to mention that it's a function from the global namespace.

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -303,11 +303,494 @@ describe('model statements', () => {
...
+      );
+      // TODO this doesn't work today, we're not rigorous enough with integer
+      // subtypes. But we should probably make this typecheck properly.
+      test.skip('cannot use float in round precision', () => {
+        expect(`query: a -> {
+          group_by: x is round(1.5, 1.6)
+        }`).compileToFailWith(
+          // TODO improve this error message
+          "Parameter 2 ('precision') for round must be integer, received float"
+        );
```
christopherswenson, line 669: We don't do number subtypes well today. So this kind of error ends up being a SQL error instead of a compile error.

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -539,20 +1022,35 @@ describe('qops', () => {
...
+    ));
   test(
     'order by asc',
-    modelOK('query: a->{ order_by: af asc; group_by: astr }')
+    modelOK('query: a->{ order_by: astr asc; group_by: astr }')
   );
   test(
     'order by desc',
-    modelOK('query: a->{ order_by: af desc; group_by: astr }')
+    modelOK('query: a->{ order_by: astr desc; group_by: astr }')
```
christopherswenson, line 1053: We didn't require in the translator that order by fields be in the output space, though if they weren't we generated probably incorrect SQL. Now we do, so I updated these tests.

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -1046,12 +1556,37 @@ describe('expressions', () => {
...
+                  'function': 'numberLiteral',
+                  'literal': '3',
+                  'type': 'dialect',
+                },
+              ],
+              'type': 'dialect',
+            },
+            '))',
+          ],
+        });
```
christopherswenson, line 1589: This test broke when I made number literals be a dialect function, rather than just a string fragment. I re-wrote the test, which revealed that we're generating double parentheses and I'm not sure why.

packages/malloy/src/model/malloy_query.ts
```diff
@@ -346,6 +364,175 @@ class QueryField extends QueryNode {
...
+
+  private expandFunctionCall(
+    dialect: string,
+    overload: FunctionOverloadDef,
+    args: Expr[]
+  ) {
+    const paramMap = this.getParameterMap(overload, args.length);
+    if (overload.dialect[dialect] === undefined) {
+      throw new Error(`Function is not defined for dialect ${dialect}`);
+    }
```
christopherswenson, line 407: It would be great if we could catch this in the translator, not the compiler?

packages/malloy/src/model/malloy_query.ts
```diff
@@ -346,6 +364,175 @@ class QueryField extends QueryNode {
...
+      const funcCall: Expr = this.expandFunctionCall(
+        context.dialect.name,
+        overload,
+        mappedArgs
+      );
+
+      if (expressionIsAnalytic(overload.returnType.expressionType)) {
+        // TODO probably need to pass in the function and arguments separately
+        // in order to generate parameter SQL correctly in BQ re: partition
+        return this.generateAnalyticFragment(
```
christopherswenson, line 509: It seems like this is actually ok?

packages/malloy/src/model/malloy_query.ts
```diff
@@ -593,77 +775,143 @@ class QueryField extends QueryNode {
...
+            expressionIsScalar(exprType)
+          ) {
+            // ANY_VALUE here is a no-op which turns the expression into an aggregate, which allows
+            // BigQuery to use a field which is not in the group_by directly (this happens when the
+            // order_by of the window function (from the query's order_by) is on a field which is an
+            // expression rather than a column -- therefore the group_by contains an expression based
+            // on a field and not the field itself, causing an error when it's referenced in the window
+            // order_by.
+            // TODO decide if there's a better way to do this
+            orderSQL = `ANY_VALUE(${orderSQL})`;
```
christopherswenson, line 860: Lloyd might have an opinion here. Carlin gave me this idea to fix an issue where expressions involving fields that are in the output don't work in the order by in a window function's OVER clause. So in BigQuery, for these we wrap them in `ANY_VALUE` to turn them into aggregates, which are allowed, while not actually changing the semantics.

packages/malloy/src/model/malloy_query.ts
```diff
@@ -593,77 +775,143 @@ class QueryField extends QueryNode {
...
+        if (
+          arg.length !== 1 ||
+          typeof arg[0] === 'string' ||
+          arg[0].type !== 'dialect' ||
+          arg[0].function !== 'numberLiteral'
+        ) {
+          throw new Error('Invalid number of rows for window spec');
+        }
+        // TODO this does not handle float literals correctly
+        return arg[0].literal;
```
christopherswenson, line 902: This is another case where we don't handle float vs integer well.

packages/malloy/src/model/malloy_query.ts
```diff
@@ -1956,8 +2232,24 @@ class QueryQuery extends QueryField {
...
+          );
+        }
+        // TODO Do we need to call `addStructToJoin` here in the case when there is no `structPath`
+        // and the function is an aggregate function?
+        for (const e of expr.args) {
+          this.addDependantExpr(resultStruct, context, e, joinStack);
+        }
+        if (expressionIsAnalytic(expr.overload.returnType.expressionType)) {
+          resultStruct.root().queryUsesPartitioning = true;
+        }
```
christopherswenson, line 2252: Might need Lloyd's opinion here.

packages/malloy/src/model/malloy_types.ts
```diff
@@ -434,38 +431,106 @@ export function mkExpr(
...
+  if (e1 === 'scalar_analytic' || e2 === 'scalar_analytic') {
+    ret = 'scalar_analytic';
+  }
+  if (e1 === 'aggregate_analytic' || e2 === 'aggregate_analytic') {
+    ret = 'aggregate_analytic';
+  }
+  if (e1 === 'scalar_analytic' && e2 === 'aggregate') {
+    ret = 'aggregate_analytic';
+  } else if (e1 === 'aggregate' && e2 === 'scalar_analytic') {
+    ret = 'aggregate_analytic';
```
christopherswenson, line 525: I don't have the brain power right now to think about this case of `analytic` + `ungrouped_aggregate`. Maybe need to think about it.

test/src/databases/all/functions.spec.ts
```diff
@@ -0,0 +1,1071 @@
...
+
+  describe('replace', () => {
+    it(`works - ${databaseName}`, async () => {
+      await funcTestMultiple(
+        ["replace('aaaa', 'a', 'c')", 'cccc'],
+        ["replace('aaaa', r'.', 'c')", 'cccc'],
+        [
+          "replace('axbxc', r'(a).(b).(c)', '\\\\0 - \\\\1 - \\\\2 - \\\\3')",
+          databaseName === 'postgres' ? '\\0 - a - b - c' : 'axbxc - a - b - c',
+        ],
```
christopherswenson, line 211: DuckDB and BigQuery support using `\\0` to refer to "the whole match" when doing a regexp replace. Postgres doesn't.

test/src/databases/all/functions.spec.ts
```diff
@@ -0,0 +1,1071 @@
...
+        ["substr('foo bar baz', -3)", 'baz'],
+        ['substr(null, 1, 2)', null],
+        ["substr('aaaa', null, 1)", null],
+        ["substr('aaaa', 1, null)", null]
+      );
+    });
+  });
+
+  describe('stddev', () => {
+    // TODO symmetric aggregates don't work with custom aggregate functions in BQ currently
```
christopherswenson, line 247: Worth noting that custom agg functions don't support symmetric aggregates in BQ today. We error if you try to use an agg function in a case where symmetric aggregates are required.

test/src/databases/all/functions.spec.ts
```diff
@@ -0,0 +1,1071 @@
...
+      expect(result.data.path(1, 'prev_sum_of_seats').value).toBe(0);
+      expect(result.data.path(2, 'prev_sum_of_seats').value).toBe(230);
+    });
+  });
+
+  describe('trunc', () => {
+    it(`works - ${databaseName}`, async () => {
+      await funcTestMultiple(
+        ['trunc(1.9)', 1],
+        // TODO Remove when we upgrade to DuckDB 0.8.X -- DuckDB has some bugs with rounding
```
christopherswenson, line 555: A bunch of tests fail for DuckDB WASM right now, so I've skipped them. I think they should all be fixed once we can upgrade to 0.8.

test/src/databases/all/functions.spec.ts
```diff
@@ -0,0 +1,1071 @@
...
+        ['sign(-2)', -1],
+        ['sign(0)', 0],
+        ['sign(null)', null]
+      );
+    });
+  });
+  describe('is_inf', () => {
+    it(`works - ${databaseName}`, async () => {
+      await funcTestMultiple(
+        ["is_inf('+inf'::number)", true],
```
christopherswenson, line 680: Maybe in the future we can have a symbol for `infinity` and for `nan`.

test/src/databases/all/functions.spec.ts
```diff
@@ -0,0 +1,1071 @@
...
+        ["repeat('foo', 1)", 'foo'],
+        ["repeat('foo', 2)", 'foofoo'],
+        ['repeat(null, 2)', null],
+        ["repeat('foo', null)", null]
+      );
+    });
+    // TODO how does a user do this: the second argument needs to be an integer, but floor doesn't cast to "integer" type.
+    it.skip(`works floor decimal - ${databaseName}`, async () => {
+      await funcTest("repeat('foo', floor(2.5))", 'foofoo');
+    });
```
christopherswenson, line 904: There's currently no way to use `floor(some_decimal)` in a place that requires an integer, because there's no way to cast to an integer, and `floor` doesn't do it.

christopherswenson (commented):
Here's a list of all the TODO I've added...

Locations of TODO comments I've added:
- [ ] [Improve API for `sqlAggDistinct`](https://github.com/malloydata/malloy/pull/1060/files#diff-cd1285bcd8120c9aee229297e70d4439ad0aee650c488139b0dc3915c2abb53dR285)
- [ ] `global-name-space.ts` added 
- [ ] [Use DuckDB's `//` operator for integer division when we upgrade to 0.8](https://github.com/malloydata/malloy/pull/1060/files#diff-42674399c40455cb1b50a8b08b4b8025b0314a602462dd18978e64629c606016R40)
- [ ] `greatest_and_least.ts` added
- [ ] [`null` propagation behavior for `concat`](https://github.com/malloydata/malloy/pull/1060/files#diff-3a3e13a091e7251a7475f942e36fa68b9ee6f4309cc1c902662dcdb3ed22a259R37)
  In DuckDB, `concat('foo', null) = 'foo'` in BigQuery, it is `null`. There is currently no compatibility transform here.
- [ ] [Better concept of generics for function definitions](https://github.com/malloydata/malloy/pull/1060/files#diff-4639b99a989fb9f3bf1a56459bfe9256024db9625f4debf8b5f6142224b2484fR47)
  Today you just have to write separate overloads for each type, but it would be nice to declare type generics with a name and list of allowed types, then use that name in the return type / params.
- [ ] [More parameters to `regexp_extract`](https://github.com/malloydata/malloy/pull/1060/files#diff-baafa44cb9f5c8f159da7ebd1d5b6b42207518c7b80e43bf04aae26ce77610feR42)
- [ ] [Should `regexp_replace` be a separate function from `replace`](https://github.com/malloydata/malloy/pull/1060/files#diff-87822ce1db382669594fb7b509a22d2ac683bc06f5808f27f00ca52a3ce72a58R44)
- [ ] [No good way of requiring that a parameter be an integer](https://github.com/malloydata/malloy/pull/1060/files#diff-6b485bdf9b7d4f5ad649e4a7db46d2f273fea13a5d9405720d3d40008e03c4e3R35)
- [ ] [Maybe add a `mode` parameter for `round`](https://github.com/malloydata/malloy/pull/1060/files#diff-6b485bdf9b7d4f5ad649e4a7db46d2f273fea13a5d9405720d3d40008e03c4e3R45)
- [ ] [Weird assumption that the OVER clause goes right after the SQL for analytic functions](https://github.com/malloydata/malloy/pull/1060/files#diff-2ff19fcdb53ceefd5b53aec1dd06b005f411608864eddd3d89083537267feffcR26)
- [ ] [`trunc` integer constraint](https://github.com/malloydata/malloy/pull/1060/files#diff-161fe5905d83ae693d57dbf7d376fea4161de292348c97eaa199b8c164bee9f8R35)
- [ ] `replace.ts` added
- [ ] [Casting to `NUMERIC` for `ROUND` in Postgres](https://github.com/malloydata/malloy/pull/1060/files#diff-53ad85fe5bec3d2dbd7c360fdc918b81b85a72cda552eea7b116eeaccfdda7cdR44)
- [ ] [`sqlAggDistinct` in Postgres doesn't preserve type](https://github.com/malloydata/malloy/pull/1060/files#diff-3eb136d89450c797665bf32d8d0e518ab54687d222b0fbc9f98e74564a067211R41) [also here](https://github.com/malloydata/malloy/pull/1060/files#diff-8a35d64ac038ac72be9f1f6b2eace8f88e1260d44bb4cf622420f9256de46994R400)
- [ ] `query_spaces.ts` added: Big one. Expanding * happens in two places. It should only happen in the translator.
- [ ] `postgres/functions/trunc.ts` added
- [ ] [Parameter eval space](https://github.com/malloydata/malloy/pull/1060/files#diff-ba5305abdd4fb71d0a31bd0fa8cd476925f5d8909bb2acfb7ed35320d414ec61R77)
- [ ] [Are nulls always allowed to be passed as an argument?](https://github.com/malloydata/malloy/pull/1060/files#diff-c3be56d5908c3aff0a90481512380f0f524ccc2be1fe38c462de9a47d9e5b364R281)
- [ ] [Error cascading for functions](https://github.com/malloydata/malloy/pull/1060/files#diff-c3be56d5908c3aff0a90481512380f0f524ccc2be1fe38c462de9a47d9e5b364R285)
  `unknown` typed parameters are always accepted. This should probably be true in general, not just here.
- [ ] [Clean up for ungroup expression typechecking](https://github.com/malloydata/malloy/pull/1060/files#diff-fc34f2e464f1993d2069df32b3431aa16d65a595b5fcdb9e15ac756d6e71b03cR79)
- [ ] `expr-ungroup.ts` Added (for refactoring to just check output space directly)
- [ ] not yet added: check about `top: ... by ...` and whether it still works (or was even working before)
- [ ] [Better error message location when an aggregate analytic is used in a `project`](https://github.com/malloydata/malloy/pull/1060/files#diff-abf4bae228fbeb88e19e7bcf2960960d8e3313cec1e4136621a878f0dc33b91dR44)
- [ ] Added: "we don't handle referencing a join as a field correctly in all cases today" in `parse.spec.ts`
- [ ] [Maybe rename `QuerySpace` to `QueryOutputSpace`](https://github.com/malloydata/malloy/pull/1060/files#diff-75c615a8233cced63b6f5fbf9949053ce7116443c02a1c5a3eb15cba877450aeR84)
- [ ] [Should turtles be added to the query output space when refining?](https://github.com/malloydata/malloy/pull/1060/files#diff-75c615a8233cced63b6f5fbf9949053ce7116443c02a1c5a3eb15cba877450aeR112)
- [ ] [Fix `**`](https://github.com/malloydata/malloy/pull/1060/files#diff-75c615a8233cced63b6f5fbf9949053ce7116443c02a1c5a3eb15cba877450aeR181)
- [ ] Duplicate for [Correct error message location for using aggregate analytic in project](https://github.com/malloydata/malloy/pull/1060/files#diff-75c615a8233cced63b6f5fbf9949053ce7116443c02a1c5a3eb15cba877450aeR295)
- [ ] [Mention of issue where we add fields with no definition to the query space when refining or wildcarding](https://github.com/malloydata/malloy/pull/1060/files#diff-75c615a8233cced63b6f5fbf9949053ce7116443c02a1c5a3eb15cba877450aeR302)
- [ ] [I don't understand how `having` typechecking works](https://github.com/malloydata/malloy/pull/1060/files#diff-cbc67f9091fccdb4085ceba5c625d8d5bf749dd21e55f786d34c579354d17adbR79)
- [ ] [Now that order by looks up fields in the output space we can probably enable jump-to-definition for them](https://github.com/malloydata/malloy/pull/1060/files#diff-253de87ba5100bd80aa25a9128fc073f37ae0ef2b34a5bbedcb6ec1a4e5b006cR49)
- [ ] [Get rid of `opOutputStruct` (future project)](https://github.com/malloydata/malloy/pull/1060/files#diff-ffdc0f37bd0bac7da47568250d9e674137c469d5d57e7fdd06e3950a6b99b119R132)
- [ ] [Now that top by looks up fields in the output space we can probably enable jump-to-definition for them](https://github.com/malloydata/malloy/pull/1060/files#diff-c7063ad8eba693d5e190cf5b9d7f666c81ca653cf536b5ec72dccbbef5c8ddc0R48)
- [ ] [Implement a better way to stop cascading errors than just stopping compilation after hitting one top level object with an error](https://github.com/malloydata/malloy/pull/1060/files#diff-42730e6a28ab08c93144b0d755af3d2dc484bedbbfaaf1f901432f0baadc5900R337)
  If instead we just have the `unknown` type (maybe rename to `error` type) be accepted everywhere instead of rejected everywhere, I think we remove a ton of cascading for cases that are real cascading issues without removing totally unrelated error messages.
- [ ] [Bad error location for redefining things in the global namespace](https://github.com/malloydata/malloy/pull/1060/files#diff-42730e6a28ab08c93144b0d755af3d2dc484bedbbfaaf1f901432f0baadc5900R456)
- [ ] [Not sure about eval space of parameters](https://github.com/malloydata/malloy/pull/1060/files#diff-5fe46a010c7eba301e3d83cfdccb52605d701a4ce10901a079a8eed85c6f483cR45)
- [ ] [Better error message when referring to 'output' eval space](https://github.com/malloydata/malloy/pull/1060/files#diff-5fe46a010c7eba301e3d83cfdccb52605d701a4ce10901a079a8eed85c6f483cR45)
- [ ] [Better error message when redefining a global function](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R650)
- [ ] [Cannot typecheck number subtypes](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R661)
- [ ] [Better error message for eval space](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R705)
- [ ] [Ability to order by an analytic field](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R733)
- [ ] [Don't know how to deal with `group_by: join_name`](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R765)
- [ ] [Too many parentheses in pick statement](https://github.com/malloydata/malloy/pull/1060/files#diff-ddb4173915729e101e4d1482297bf97fb16392c33cca9b039373e8f91dc83011R1566)
- [ ] [Incorrect assumption that all parameters to aggregate-returning functions need to have filters applied to them](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R492)
- [ ] Might be okay actually: [unsure about generating partition statements for analytics in BigQuery](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R507)
- [ ] [Correctly filter out 'self' analytic when generating OVER ORDER BY when we allow ordering by analytic values](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R841)
- [ ] [Decide if using `ANY_VALUE` is okay for ordering by dimensional expressions in window ORDER BY](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R859)
- [ ] [Don't understand float literals when generating window BETWEEN clauses](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R901)
- [ ] [Not sure about adding struct to join when using custom aggregate function](https://github.com/malloydata/malloy/pull/1060/files#diff-99024c4735395f44ec7c632e0eb4feec5a3cf890db23a0cc9a4ca2d5f1e47dd2R2245)
- [ ] [Better name for `maxExpressionType`](https://github.com/malloydata/malloy/pull/1060/files#diff-19b9cde78325f8c932fef3893b07c3d31268d9205d7bcca54b832863457e6672R503)
- [ ] [Case when combining `ungrouped_aggregate` with `analytic`](https://github.com/malloydata/malloy/pull/1060/files#diff-19b9cde78325f8c932fef3893b07c3d31268d9205d7bcca54b832863457e6672R508)
- [ ] [Consistent null behavior for `concat`](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R123)
- [ ] [DuckDB WASM bug with rounding with precision](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R134)
- [ ] [DuckDB WASM bug with flooring negative numbers](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R150)
- [ ] [DuckDB WASM bug with ceil-ing negative numbers](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R165)
- [ ] [Symmetric aggregates don't work for custom agg functions in BQ](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R247)
- [ ] [DuckDB WASM bug with trunc-ing negative values](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R555)
- [ ] [More exhaustive tests for trig functions](https://github.com/malloydata/malloy/pull/1060/files#diff-c3d8473b0c6147f94f3e629f393fc1ba03571f350c942aa013616a15c1039002R597)

packages/malloy/src/lang/ast/elements/import-statement.ts
```diff
@@ -60,7 +60,11 @@ export class ImportStatement extends MalloyElement implements DocStatement {
         for (const [importing, entry] of Object.entries(
           trans.getChildExports(this.fullURL)
         )) {
-          doc.setEntry(importing, {entry, exported: false});
+          if (doc.getEntry(importing)) {
+            this.log(`Cannot redefine '${importing}'`);
+          } else {
+            doc.setEntry(importing, {entry, exported: false});
+          }
```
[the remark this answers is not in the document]
christopherswenson: Ah I think the point is just to move the error message to the right place.

christopherswenson:
Related pull requests:
* Docs: https://github.com/malloydata/malloydata.github.io/pull/31
* Samples: https://github.com/malloydata/malloy-samples/pull/21
* VSCode: https://github.com/malloydata/malloy-vscode-extension/pull/152

christopherswenson:
Well, in attempting to fix an issue where this query didn't work in BigQuery:

```
query: state_facts -> {
 group_by: lower_state is lower(state)
  calculate: prev_state is lag(lower_state)
}
```

I managed to either break a bunch of things in BigQuery for analytic functions or expose that they were already broken and untested. I will take a look at the failing tests next.

PR #1061, opened 2023-03-14
feat: make the use of Nix hermetic

fzakaria:
The use of Nix before used <nixpkgs> which is not hermetic. 

Please see https://nix.dev/tutorials/towards-reproducibility-pinning-nixpkgs

* moved default.nix to shell.nix to better describe intent
* added sources.json to pin nixpkgs
* added more documentation
* added direnv integration


nix/sources.nix
```diff
@@ -0,0 +1,194 @@
+# This file has been generated by Niv.
```
fzakaria, line 1: This file is auto generated by https://github.com/nmattia/niv

shell.nix
```diff
@@ -0,0 +1,20 @@
+let
```
fzakaria, line 1: If you'd like to use `nix build` itself -- I can help there.
Useful if you want others to install via Nix.

lloydtabb: /gcbrun

lloydtabb (commented):
These look good to me.  @bporterfield @whscullin ?

PR #1063, opened 2023-03-20
Remove special casing for UUID in postgres

### LINKED ISSUE #1021: Remove interim uuid support
We had a new user who was hitting a wall so we added a quick workaround to treat uuid fields in postgres as if they were strings.

With #1019 we should remove this, and let uuid columns be treat like all other data which doesn't have a mapping in the Malloy expression space.

FIling this is as an issue because I want to remember to do this AFTER https://github.com/malloydata/malloy-vscode-extension/pull/69 lands and the release that contains it ships.

fzakaria:
Malloy code was augmented to make any unknown types cause failures.
The special casing for UUID in postgres can now be removed.

fixes #1021

### DIFF
--- a/packages/malloy-db-postgres/src/postgres_connection.ts
+++ b/packages/malloy-db-postgres/src/postgres_connection.ts
@@ -72,7 +72,6 @@ const postgresToMalloyTypes: {[key: string]: AtomicFieldTypeInner} = {
   numeric: 'number',
   bytea: 'string',
   pg_ndistinct: 'number',
-  uuid: 'string',
 };
 
 interface PostgresQueryConfiguration {


fzakaria:
I found the test in https://github.com/malloydata/malloy/blob/main/test/src/render/render.spec.ts but not sure how to get Postgres running to validate if it needs changing.

mtoy-googly-moogly:
@fzakaria letting ci run as a fine way to test that

mtoy-googly-moogly:
For the record. Malloy used to essentially delete unsupported types from schemas.

Now unsupported types are available in sources, you can do the following with columns that have a type unsupported by Malloy

* check if two expressions of the same unsupported type are equal
* check in an expression of unsupported type is null
* cast
* pass them to functions
* have an unsupported type in a `group_by:` or `project:`

The UUID hack was a workaround while a better solution could be crafted, thanks for removing it!

PR #1065, opened 2023-03-20
infer sql block connection from turducken source

### LINKED ISSUE #1008: "File not found" in SQL Turducken when querying from Source
When querying from a Source inside of a Turducken, I'm getting a "No files found that match pattern" error. The same query compiles just fine when used outside of the Turducken.

In my example, the Turducken is a SQL query that has a Malloy query as the `from` clause. The inner Malloy query uses a named Source. For a simple example, see [this snippet from the Malloy samples repo](https://github.com/malloydata/malloy-samples/pull/13).

image

mtoy-googly-moogly:
This is an attempt to fix #1008.

If an SQL block has no connection, and it can find a connection in a turducken inside the block, use that as the connection for the sql block.

Not sure how to write a test for this.

### DIFF
--- a/packages/malloy/src/lang/ast/sql-elements/sql-statement.ts
+++ b/packages/malloy/src/lang/ast/sql-elements/sql-statement.ts
@@ -21,7 +21,12 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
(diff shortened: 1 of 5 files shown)


mtoy-googly-moogly:
For some reason I never merged this, and in the new syntax it is impossible to write sql which doesn't name a connection, so now it should never be merged.

PR #1074, opened 2023-03-29
Accept all duckdb numerics as numbers

### LINKED ISSUE #1073: json columns from duckdb show as "unsupported" types ...
**id_title.json:**
```JSON
[ { "id": 1, "title": "one" } ]
```

**id_title.malloy:**
```Malloy
sql: read_json_table is { connection: "duckdb" select: """
  SELECT *
  FROM read_json_auto("/Users/mtoy/malloy/id_title.json", auto_detect=true)
"""}

query: from_sql(read_json_table) -> { aggregate: id_sum is sum(id)}
```

Errors because the type of id ends up being "unsupported". Not clear if this is a bug in duckdb or Malloy schema reading ...

Wherever the bug is, the error `'sum' can't use type unsupported` should actually show the raw type, something like `'sum' can't operate on unsupported type 'typeFromSchema'`

mtoy-googly-moogly:
@lloydtabb was there any reason we didn't do this from the start?

Fixes #1073

### DIFF
--- a/packages/malloy-db-duckdb/src/duckdb_common.ts
+++ b/packages/malloy-db-duckdb/src/duckdb_common.ts
@@ -39,6 +39,14 @@ import {
 
 const duckDBToMalloyTypes: {[key: string]: AtomicFieldTypeInner} = {
   BIGINT: 'number',
+  INTEGER: 'number',
+  TINYINT: 'number',
+  SMALLINT: 'number',
+  UBIGINT: 'number',
+  UINTEGER: 'number',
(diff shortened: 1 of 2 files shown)


lloydtabb:
> @lloydtabb was there any reason we didn't do this from the start?
> 
> Fixes #1073

Two Possibilities.
1) I'm notoriously lazy and was moving too fast
2) They added types later

No idea which is true but I could believe either.

mtoy-googly-moogly:
As long as there are on objections I will merge on green.

PR #1077, opened 2023-04-04
Fix extending models with from_sql based sources

### LINKED ISSUE #1076: creating a source `from_sql` is broken
Trying to create a source from sql is broken...  Seems pretty urgent unless there is a work around I'm not seeing.

The code below generates the following error.

```
Error(s) compiling model:
FILE: internal://internal.malloy
line 1: Must use 'from_sql()' for sql source 'order_items_table
  | query: order_items_table->by_sequence
  |        ^
```
The code.
```
sql: airports_sql is {
  select: """
    select 
      *
      , ROW_NUMBER() OVER (PARTITION BY state ORDER BY elevation) as rn
    FROM malloy-data.malloytest.airports
  """
  connection: "bigquery"
}

// Broken:  This eroneously reports an error
source: order_items_table is from_sql(airports_sql) {
// Broken:  This eroneously reports an error
  query: by_sequence is {
    group_by: rn
    aggregate: order_count is count()
  }
}
```

mtoy-googly-moogly:
Fixes #1076 

StructDefs which resulted from an `sql:` statement now have a bit on them, in the structdef.

We clear that bit when we alias them into a `source:`

This allows us to treat sql blocks as if they were a thing which is neither a query nor structdef, and properly error when people forget to use from_sql (and not erroneously error when people do)

KInd of a lot of work to maintain a distinction which none of us believe in, but this is the most peaceful thing for me, given that there is no guarantee we are actually going to get to source/query, this PR correctly implements the current intended semantics for SQL blocks.

We will undo this when we do source/query, and that's OK with me too.

### DIFF
--- a/packages/malloy/src/lang/ast/sources/named-source.ts
+++ b/packages/malloy/src/lang/ast/sources/named-source.ts
@@ -22,6 +22,7 @@
  */
 
 import {
+  isSQLBlockStruct,
   isValueParameter,
   paramHasValue,
   StructDef,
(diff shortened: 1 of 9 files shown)


christopherswenson:
Looks good to me!

PR #1081, opened 2023-04-12
Support Explore object serialization with JSON

caleb-c-li:
This patch supports the `Explore` object serialization with JSON:
- encapsulated structDef in an Explore and all its parents, sources into a struct for JSON serialization.
- leave the server side optimization of `_fieldMap` field to future, as it contains self references.

### DIFF
--- a/packages/malloy/src/malloy.ts
+++ b/packages/malloy/src/malloy.ts
@@ -1241,6 +1241,11 @@ abstract class Entity {
 }
 
 export type Field = AtomicField | QueryField | ExploreField;
+export type SerializedExplore = {
+  _structDef: StructDef;
+  sourceExplore?: SerializedExplore;
+  _parentExplore?: SerializedExplore;
+};
 
 export class Explore extends Entity {
   protected readonly _structDef: StructDef;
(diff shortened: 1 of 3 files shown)


packages/malloy/src/malloy.ts
```diff
@@ -1409,6 +1414,34 @@ export class Explore extends Entity {
   public get structDef(): StructDef {
     return this._structDef;
   }
+
+  public toJson(): SerializedExplore {
```
whscullin, line 1418: In this codebase the existing methods are styled `toJSON()` and `fromJSON()`,  in alignment with the standard Javascript `toJSON()` method. This is counter to Google's style guide for acronyms but when in Rome...

packages/malloy/src/malloy.ts
```diff
@@ -1409,6 +1414,34 @@ export class Explore extends Entity {
...
+    };
+  }
+
+  public toJsonString(): string {
+    return JSON.stringify(this.toJson());
+  }
+
+  public static fromJsonString(json_explore: string): Explore {
+    return Explore.fromJson(JSON.parse(json_explore) as SerializedExplore);
+  }
```
whscullin, line 1432: Similarly, these don't have counterparts elsewhere, and I'm not sure they're really necessary.

caleb-c-li: Removed the fromJsonString () and toJsonString()`stringify()/parse()`.

packages/malloy/src/malloy.ts
caleb-c-li: Kept the names as toJSON() and fromJSON(). The toJSON() method is implicitly called when serialize one Explore object.

christopherswenson (approved):
Seems good to me 👍

PR #1084, opened 2023-04-18
Accept table read functions in duckdb as tables

mtoy-googly-moogly:
If we just turn off quoting, we gain the ability to treat a table read function in duckdb as a table ... enabling users to write

```
sql: user_from_json is { connection: "duckdb" select: """
   SELECT * from read_json_auto("some file", various options)
 """}
 
 source: users is from_sql(user_from_json) {
 ```
 
 as
 
 ```
 source: users is table('duckdb:read_json_auto("some file", various options)') {
 ```
 
 What we do is, if the table name starts with WORD OPEN_PAREN and ends with CLOSE PAREN, we just pass it through to duckdb to parse however it likes. This will let us read JSON, CSV and PARQUET files with all the options duckdb has, without having to have an SQL block

### DIFF
--- a/packages/malloy/src/dialect/duckdb.ts
+++ b/packages/malloy/src/dialect/duckdb.ts
@@ -163,6 +163,11 @@ export class DuckDBDialect extends Dialect {
   }
 
   quoteTablePath(tableName: string): string {
+    // If this looks like a function call, don't quote, this lets you
+    // specify read_json_auto(...) for example, as if it were a table name
+    if (tableName.match(/^[a-z_]+\(.*\)$)/)) {
+      return tableName;
+    }
     return tableName.match(/\//) ? `'${tableName}'` : tableName;
   }
 

### CODE REVIEW

**Comment by carlineng** [comment]
@mtoy-googly-moogly one question for you -- all the DuckDB examples in the [JSON Loading docs](https://duckdb.org/docs/data/json/overview) use single-quotes to wrap parameter values, but your example above uses double-quotes; so

```
source: users is table('duckdb:read_json_auto("some file", various options)')
```

would actually be specified as:

```
source: users is table('duckdb:read_json_auto('some file', various options)')
```

Would that cause problems for the parser?

**Comment by carlineng** [comment]
It looks like DuckDB does not like double-quotes... I get a compilation error when trying the following:

```
sql: my_sql_query is {
  select: """
    SELECT
      *
    FROM read_csv(
      ""../data/test.csv",
      header=True,
      columns={'test': 'VARCHAR', 'data': 'VARCHAR'} 
      )
  """
  connection: "duckdb"
}
```

image

carlineng:
Actually, it looks like DuckDB does accept double-quotes for those arguments, even though the examples all specify single-quotes.

mtoy-googly-moogly:
The brain trust didn't approve. Closing.

PR #1088, opened 2023-04-27
Tweak linter   - quote props

srschandan:


whscullin (approved):
LGTM. We can investigate writing a lint rule to enforce string keys.

PR #1097, opened 2023-05-05
Pass projectId and credentials body to BQ auth

carinakoo8:
To specify projectId and credentials dynamically, we can pass `projectId` to BigQuery, and we can pass `credentials` body in lieu of `serviceAccountKeyPath`.


bporterfield (approved):

bporterfield:
thanks @carinakoo8!

PR #1107, opened 2023-05-19
Exposing bytes processed when running a bigquery query through malloy api.

narreola:
Working on adding some tests.


bporterfield (approved):
Looks good to me! Thank you.

Issue #1106, opened 2023-05-13
Scatter charts seem to always start fom a zero axis


lloydtabb:
Reported on Slack.

![image](https://github.com/malloydata/malloy/assets/1093458/934cbdc3-38ce-426c-84a8-3ff4b43ca102)

srschandan:
In your case the chart does seem disoriented and confusing. I can make sure that this behavior will only be localized to scatter charts. However, allowing the user to toggle this behavior may need a bigger discussion so I'll leave it at that.

lloydtabb:
Yes.  I agree.  The non zero axis should only effect scatter charts. Bar
charts should start at 0

PR #1114, opened 2023-05-23
Fix chart always starts at zero issue. 

srschandan:


srschandan:
Fixes: https://github.com/malloydata/malloy/issues/1106

srschandan:
Screenshot 2023-05-22 at 5 00 03 PM

PR #1124, opened 2023-05-25
Nested pipelines are busted in DuckDB

lloydtabb:
Aparently they didn't implement 

> Binder Error: Nested lateral joins or lateral joins in correlated subqueries are not (yet) supported

May have to revert to using array indexes :(


lloydtabb:
Here is where the problem shows itself in our demos...

https://github.com/malloydata/patterns/blob/main/auto_bin.malloynb

PR #1127, opened 2023-05-29
add peggy to devDependencies

mtoy-googly-moogly:
@bporterfield i don't know if it belongs there or in normal dependencies


mtoy-googly-moogly:
dep was alredy existing in sub package ... closing

PR #1135, opened 2023-06-04
Annotation

mtoy-googly-moogly:
The basic idea is to use lines that start with `#` to attach meta data to objects which mean nothing to the compiler but which applications can attach meaning to.

First customer is our rendering library.

Overview:

* `#` to end of line is a tag, are collected and applied to objects in the model by various rules
* There is no restriction on what can be put in these tags
* There is a helper function which parses particular kinds of tags for API users.
  * Tags which start `#"` are considered doc strings
  * Tags which start `# ` are considered render tags
  * All tags (including the above) are available to API users to interpret as the prefer


packages/malloy/src/tags.ts
```diff
@@ -0,0 +1,177 @@
...
+          }
+          tagProp[token] = value;
+          tn += 3;
+          continue;
+        }
+      }
+      return undefined;
+    }
+    if (token.startsWith('-')) {
+      delete tagProp[token.slice(1)];
```
christopherswenson, line 137: Didn't know we were doing this, but that's neat. Seems useful.

christopherswenson (commented):
Looks great! Made a couple of comments. Do what you want with them 😊

packages/malloy/src/tags.ts
christopherswenson: I love that we'll be able to do:

```
# dashboard
query: foo is ...

# -dashboard
query: bar is -> foo ...
```

packages/malloy/src/tags.ts
```diff
@@ -0,0 +1,177 @@
...
+ * A string is " enclosing and ending in " with \" allowed inside
+ */
+function parseTag(
+  src: string,
+  tagProp: MalloyTagProperties
+): MalloyTag | undefined {
+  if (src.startsWith('#" ')) {
+    return {docString: src.slice(3)};
+  }
+  if (!src.startsWith('# ')) {
```
christopherswenson, line 101: Looks like our default parser won't allow `#dashboard` (vs `# dashboard`). I think this might be confusing to people? I guess we can make the syntax highlighter highlight the former in a way that makes it clear it's wrong...

packages/malloy/src/tags.ts
mtoy-googly-moogly: Yeah, I tried to tell lloyd that the usage was "render=barchart" , which would refine properly, but he made sad noises and this is how you do that.

I actually think "render=barchart" might be better, but the hierarchy is

"render-style-name"
can be refined with "refine-render=render-style-name"

But I thought I would give - a spin first

packages/malloy/src/tags.ts
mtoy-googly-moogly: Yeah, because I am, in my mind, imagining the character after the '#' determines who gets the annotation in a world where there are multiple users, and i didn't want to own all '#[a-z]' annotations

christopherswenson: Yep, makes sense. I'm good with that

packages/malloy/src/lang/grammar/MalloyParser.g4
```diff
@@ -32,28 +32,38 @@ malloyStatement
...
+
+isDefine
+  : beforeIs=tags IS afterIs=tags
   ;
 
 runStatement
-  : RUN topLevelAnonQueryDef     # runStatementDef
-  | RUN queryName                # runStatementRef
+  : blockTags=tags RUN noteTags=tags topLevelAnonQueryDef     # runStatementDef
+  | RUN queryName                                             # runStatementRef
```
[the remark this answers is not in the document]
mtoy-googly-moogly: No tags for `run: queryName` because that rule goes away when `queryName` is a legal `topLevelAnonQueryDef` ... it it also 10 lines of code if this bothers you

packages/malloy/src/malloy.ts
```diff
@@ -1785,14 +1798,18 @@ export class Query extends Entity {
...
 
   constructor(turtleDef: TurtleDef, parent: Explore, source?: Query) {
     super(turtleDef, parent, source);
     this.parent = parent;
   }
 
+  getTags(): Tags {
+    return new Tags(undefined);
+  }
+
```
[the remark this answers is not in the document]
mtoy-googly-moogly: because i sprinkled this everywhere ... and then only fixed them when i wrote tests that failed, so i don't know who needs that

packages/malloy/src/lang/ast/query-items/field-declaration.ts
```diff
@@ -46,14 +47,23 @@ import {
...
   defineName: string,
   exprSrc?: string
 ) => FieldDeclaration;
 
-export abstract class FieldDeclaration extends MalloyElement {
+export abstract class FieldDeclaration
+  extends MalloyElement
+  implements Noteable
+{
+  isMeasure?: boolean;
```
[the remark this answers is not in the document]
mtoy-googly-moogly: yup ... gone

packages/malloy/src/malloy.ts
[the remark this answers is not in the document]
mtoy-googly-moogly: but looking at it, i think i am just missing a test

fixed as you suggest

PR #1142, opened 2023-06-07
Return Querytimezone correctly, since it was not properly populated.

narreola:
I feel like there should be a better way to find the final stage, I'm planning to look into it early next week.

### DIFF
--- a/packages/malloy/src/malloy.ts
+++ b/packages/malloy/src/malloy.ts
@@ -412,6 +412,7 @@ export class Malloy {
           // TODO feature-sql-block There is no source explore...
           sourceExplore: '',
           sourceFilters: [],
+          // TODO: Figure out where to grab query timezone from in this case.
         },
         {
           name: 'empty_model',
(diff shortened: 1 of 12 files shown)


packages/malloy/src/malloy.ts
```diff
@@ -1131,7 +1132,9 @@ export class PreparedResult {
    * @return The query timezone.
    */
   public get queryTimezone(): string | undefined {
```
mtoy-googly-moogly, line 1134: The result structdef of a query should already have the correct timezone in it, and more things than the renderer will break of that isn't true.

The logic to make that happen, and this is in `lang/model/` so it is a deep dive for me to to figure out, is that the timezone of a query is known when the query is "loaded" and it is the last one in the list of

```
[
  The structdef that is the source for the query,
  ... each query operation in the query pipeline
]
```

mtoy-googly-moogly (approved):
Looks good, ship it

PR #1143, opened 2023-06-07
Update datetime renderer to use query timezone when available.

narreola:
This PR sits on top of: https://github.com/malloydata/malloy/pull/1142 (still figuring out if there is a way to isolate changes that sit on top of other, or chain PRs)

### DIFF
--- a/packages/malloy-render/src/html/date.ts
+++ b/packages/malloy-render/src/html/date.ts
@@ -56,9 +56,6 @@ export class HTMLDateRenderer implements Renderer {
 
     const element = this.document.createElement('span');
     element.appendChild(this.document.createTextNode(timestring));
-    element.appendChild(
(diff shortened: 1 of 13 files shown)


christopherswenson:
I tend to chain PRs by making branch `feature-a` off of `main`, then branch `feature-b` off of `feature-a`. Then I just make a PR for feature B against main like normal, but wait to merge it until A is merged.

narreola:
@christopherswenson I think that's the setup I have here, timezoneui is off of timezone, and timezone is off of main. The only thing that would be great was if there would be a way to, in PR for branch B, should diffs relative to branch A instead of relative to main.

christopherswenson:
I think you can make a pull request to another branch. So make PR B "merge into branch A" and then when branch A lands, swap PR B to be based off main.

PR #1164, opened 2023-06-19
Surface warnings when there are no errors

christopherswenson:
- Add `problems` to `Model` class (runtime API) to surface warnings when there are no errors
- Change `errors` to `problems` throughout the translator
- Add a warning when a function call does not match case
- Add parameter to `MalloyElement.log` to specify a severity


packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -670,17 +670,17 @@ export abstract class MalloyTranslation {
   fatalErrors(): FatalResponse {
     return {
       final: true,
-      errors: [...this.root.logger.getLog()],
+      problems: [...this.root.logger.getLog()],
```
mtoy-googly-moogly, line 673: filter out warnings?

mtoy-googly-moogly (changes requested):

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -82,7 +82,7 @@ function badModel(s: MarkedSource | string, msg: string): TestFunc {
     if (t.translated) {
       fail(emsg);
     } else {
-      const errList = m.errors().errors;
+      const errList = m.problems().problems;
```
mtoy-googly-moogly, line 85: problems() should return the list and problemResponse() should be protected and return the response.

test/src/databases/all/problems.spec.ts
```diff
@@ -0,0 +1,109 @@
+/*
```
mtoy-googly-moogly, line 1: should probable be part of the api tests, not its own file

packages/malloy/src/lang/parse-malloy.ts
christopherswenson: Well I did this, and it broke everything!

We're using `fatalErrors` to generate the response in the case that there _are_ fatal errors, but even then we still want to show warnings. So I renamed it to `fatalResponse`, to hopefully clarify.

christopherswenson:
Might be worth making `toCompile` fail if there are any warnings? Or make `toCompileWithoutProblems` which ensure there are no warnings? @mtoy-googly-moogly

packages/malloy/src/lang/parse-error-handler.ts
```diff
@@ -46,6 +46,7 @@ export class MalloyParserErrorHandler implements ANTLRErrorListener<Token> {
     const error: LogMessage = {
       message: msg,
       at: {url: this.translator.sourceURL, range},
+      severity: 'error',
```
christopherswenson, line 49: @mtoy-googly-moogly You mentioned you didn't want `severity` to be required in the logging functions. Is this okay, as this logging function takes a `LogMessage` directly? Or would you prefer that I change the signature of this function to either `log(message: string, at?: SourceLocation, severity = 'error')` or `log({ message, at, severity = 'error' }: { message: string; at?: SourceLocation; severity?: string })`

PR #1170, opened 2023-06-20
Add run statement

christopherswenson:
PR is temporarily against `crs-warnings` because it relies on that work.

* Adds `run:` statement at the top level of a Malloy file.
* Makes the `run:` statement accept a query definition, e.g. `run: flights -> by_carrier` or a query name, e.g. `run: flights_by_carrier`. It does not accept a source name, e.g. `run: flights` (and probably never will), and currently does not accept a SQL block name `run: my_sql_block`. It also doesn't accept a SQL block expression, e.g. `run: conn.sql('foo')` because those don't exist yet!
* Adds a warning on anonymous query definitions, e.g. `query: flights -> by_carrier`
* Still uses the same method of returning anonymous queries from models as before
* Updates document symbol walker to pick up the new run statements so code lenses can be added

```malloy
source: flights is table('malloy-data.flights') {
  query: by_carrier is { group_by: carrier; aggregate: flight_count is count() }
}

query: flights_by_carrier is flights -> by_carrier

run: flights -> by_carrier
run: flights_by_carrier
run: -> flights_by_carrier

// Syntax error
query: flights_by_carrier

// Warning: Anonymous `query:` statements are deprecated, use `run:` instead 
query: flights -> by_carrier
query: -> flights_by_carrier
```

The bit that allows `run: query_name` (rather than `run: -> query_name`) will eventually be redundant, but I think until we implement `extend` and `refine` (and possibly not until we drop the old syntax for refinements), this will be nice to allow leaving off the leading arrow.


christopherswenson:
Closes https://github.com/malloydata/malloy/issues/1171

packages/malloy/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@malloydata/malloy",
-  "version": "0.0.43",
+  "version": "0.0.44",
```
christopherswenson, line 3: @whscullin I've been bumping this incrementally (so that my `malloy-link` version doesn't conflict with the current real version), but I don't think I'm actually doing the right thing here. Do we have a strategy for keeping this up-to-date?

packages/malloy/src/lang/ast/query-elements/run-query.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  execute(doc: Document): ModelDataRequest {
+    const modelQuery = this.theQuery.query();
+    doc.queryList.push(modelQuery);
+    return undefined;
+  }
+}
+
+export class RunQueryRef extends RunQuery {
+  constructor(readonly queryName: FieldName) {
```
christopherswenson, line 50: This should probably be a `ModelEntryReference`

christopherswenson:
The warning should be behind a/the new `ENABLE_M4_WARNINGS` flag.

packages/malloy/package.json
whscullin: You shouldn't need to to touch that to use link.

Issue #1178, opened 2023-06-20
Deprecate `join_*:` in queries


christopherswenson:
Add a warning when using `join_*:` in queries, like "`join_one:` in queries is deprecated, use `extend: { join_one: ... }` instead."

Issue #1177, opened 2023-06-20
Add `extend: { ... }`


christopherswenson:
Implement new `extend:` syntax. 

```malloy
query: foo is flights -> {
  extend: {
    dimension: bar is 1
  } 
}
```

Issue #1176, opened 2023-06-20
Deprecate `declare:`


christopherswenson:
Deprecate `declare:` everywhere. In a source, the message should be like "`declare:` is deprecated, use `dimension:` or `measure:` instead." In a query, the message should be like, "`declare:` is deprecated, use `extend: { dimension: ...; measure: ... }` instead."

PR #1185, opened 2023-06-20
Add extend block

christopherswenson:
(Currently basing off of `crs-warnings` because it depends on that work).

Adds new syntax:

```malloy
query: airports -> { 
  extend: {
    measure: baz is count()
    dimension: foo is 1
    measure: bar is count()
    join_one: airports on id = airports.id
  }
  group_by: foo, airports.fac_type
  aggregate: bar, baz
}
```

Adds a constant `ENABLE_M4_WARNINGS`, which when set to `true` makes all uses of `declare:` and any `join_*:` in a query show a warning. Once annotations land, we should replace this with a check for an annotation like `## m4warnings`.


christopherswenson:
Closes https://github.com/malloydata/malloy/issues/1178
Closes https://github.com/malloydata/malloy/issues/1176
Closes https://github.com/malloydata/malloy/issues/1177

Issue #1174, opened 2023-06-20
Named connection objects


christopherswenson:
* Add a new top level model statement `connection: connection_name`
* Add model-level expression `connection_name.sql('sql_string (with interpolation)')`
* Remove `table('connection_name:table_name')`
* Add model-level expression `connection_name.table('table_name')

PR #1186, opened 2023-06-20
Connection object and new table syntax

christopherswenson:
* Adds new syntax: `connection_name.table('table.path')`
* Adds a warning when `ENABLE_M4_WARNINGS` is true for using old `table('connection_name:table.path')`

Connection objects are part of a model, and are exported, but they are defined implicitly by first use. I expect that this strategy won't work in the long term, as we only know something is a connection name by its appearance in a `xxx.table('...')` statement, and in the future we may have arbitrary "connection functions" that return sources (or maybe other things too..., constants, etc...), and so the translator will need to need to somehow know that `xxx` is the name of a connection and not some other thing. Possibly in the future we could add a way for the translator to ask the runtime "is `xxx` a connection name?," which it would do in the case where a `xxx.function_name()` was called and no model entry called `xxx` already existed.

Other changes:
* Changes the interface by which the translator asks for table schemas (from a `string[]` representing a list of "table URIs" which need to be parsed into connection name and table path; to a `Record<string, { connectionName: string, tablePath: string }>` representing a mapping from "table keys," which should not ever be parsed, to the connection name and table path associated with that table key). 
* Changes the interface by which connections report table schemas, from a `string[]` representing a list of table URIs to a `Record<string, string>` representing a mapping from table keys to table names (note: no connection name, because the connection doesn't need that information...).


packages/malloy/src/lang/ast/sources/table-source.ts
```diff
@@ -60,3 +60,41 @@ export class TableSource extends Source {
...
+
+export class TableFunctionSource extends TableSource {
+  elementType = 'tableFunctionSource';
+  constructor(readonly name: string) {
+    super();
+  }
+
+  getFullName(): string | undefined {
+    // TODO return this parsed...
+    return this.name;
```
christopherswenson, line 98: I would like for this method really to be `getTableInfo(): { connectionName: string; tableName: string }`. This would possibly mean having to a) change the translator's API to make it so the request for schemas can have two pieces of info (both the connection name and the table name), rather than one combined string b) accept the split name and then package it back up into one string for use with the translator...

christopherswenson:
Starts to address https://github.com/malloydata/malloy/issues/1174, but does not implement `connection.sql('sql')` yet, as that's a bigger and separable change.

packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -332,20 +331,8 @@ export class BigQueryConnection
     };
   }
 
-  public async structDefFromSQL(sqlCommand: string): Promise<StructDef> {
```
christopherswenson, line 335: This function did not seem to be used anywhere at all, so I removed it....

PR #1187, opened 2023-06-20
cleanup custom lang test matchers

mtoy-googly-moogly:
After some discussion @christopherswenson and I have a smaller, cleaner set of test matchers in lang that we agree on.

Been waiting for this moment to standardize, and fix all the old tests.

Now done.

### DIFF
--- a/packages/malloy/src/lang/test/imports.spec.ts
+++ b/packages/malloy/src/lang/test/imports.spec.ts
@@ -22,6 +22,7 @@
  */
 import './parse-expects';
 import {TestTranslator} from './test-translator';
+import escapeRegEx from 'lodash/escapeRegExp';
 
 describe('import:', () => {
   test('simple source', () => {
@@ -93,8 +94,9 @@ source: botProjQSrc is from(->botProjQ)
       },
     });
     docParse.translate();
-    expect(docParse).not.toParse();
-    expect(docParse.prettyErrors()).toContain(reportedError);
+    expect(docParse).translationFailsWith(
+      new RegExp(escapeRegEx(reportedError))
+    );
   });
   test('chained imports', () => {
     const docParse = new TestTranslator('import "child"');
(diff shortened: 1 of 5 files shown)


packages/malloy/src/lang/test/parse-expects.ts
```diff
@@ -37,11 +37,8 @@ declare global {
...
     interface Matchers<R> {
-      modelParsed(): R;
-      toBeErrorless(): R;
-      toCompile(): R;
-      modelCompiled(): R;
-      expressionCompiled(): R;
+      toParse(): R;
+      toTranslate(): R;
       toReturnType(tp: string): R;
       compileToFailWith(...expectedErrors: ErrorSpec[]): R;
```
christopherswenson, line 43: I think this name doesn't work — maybe `translateToFailWith`?

packages/malloy/src/lang/test/imports.spec.ts
```diff
@@ -93,7 +93,7 @@ source: botProjQSrc is from(->botProjQ)
       },
     });
     docParse.translate();
-    expect(docParse).not.toBeErrorless();
+    expect(docParse).not.toParse();
```
christopherswenson, line 96: I think we mentioned the possibility of `toParse()` and `toTranslate()` to fail if there were warnings or errors. I *think* that `checkForErrors` in the definition of `toParse()` allows warnings — though maybe this is just something for me to fix in my PR...

Also, `.not.toParse()` would then mean "has warnings or errors." Maybe we should try not to use `.not.toParse()` and instead have a `.parseToFailWith(...)`?

christopherswenson (changes requested):
Awesome. Glad we're honing in on the right answer here. Only change I feel strongly about is the name of `compileToFailWith` being something like `translateToFailWith`.

packages/malloy/src/lang/test/parse-expects.ts
mtoy-googly-moogly: I don't know how i missed this, yes `translationFailsWith` or somethiing

packages/malloy/src/lang/test/imports.spec.ts
mtoy-googly-moogly: yeah that's probably a bad test i'll fix it

PR #1204, opened 2023-06-26
update lerna and jsdom

whscullin:
Mitigate an number of vulnerabilities, remove a bunch of redundant dependency declarations.


test/src/dependencies.spec.ts
```diff
@@ -37,25 +37,25 @@ describe('dependencies', () => {
...
       fileExtensions: ['ts'],
-      excludeRegExp: [/packages\/malloy\/src\/lang/, /\.d\.ts/],
+      excludeRegExp: [/malloy\/src\/lang/, /\.d\.ts/],
     });
     expect(deps.circular().length, getMessage(deps.circular())).toBe(0);
   });
 
   it('javascript references should not be circular', async () => {
-    const deps = await madge('.', {
+    const deps = await madge('packages', {
```
whscullin, line 48: @gmaden this test was crashing when run against the whole world, and took oddly long, restricting these to packages sped things up a lot and fixed the crash. I'm not sure if we need any other directories for this, we could add them as an array if so.

PR #1205, opened 2023-06-27
Return defined renderer

whscullin:
Fix case where dataStyles specifies a renderer and there are no tag or magic name overrides.


narreola:
LGTM, thanks for fixing this.

narreola (approved):

PR #1215, opened 2023-07-05
Add getImport API

rmstar:
Added a `getImport` API on the Model which returns the import, if one exists, at the given position in the document.

### DIFF
--- a/packages/malloy/src/lang/parse-malloy.ts
+++ b/packages/malloy/src/lang/parse-malloy.ts
@@ -795,10 +795,6 @@ export abstract class MalloyTranslation {
     // search to find the one we're looking for.
     for (let index = 0; index < this.imports.length; index++) {
       const imp = this.imports[index];
-      // Ignore any import in another file
-      if (imp.location.url !== this.sourceURL) {
-        return;
-      }
       if (
         imp.location.range.start.line <= position.line &&
         imp.location.range.end.line >= position.line &&
@@ -815,6 +811,10 @@ export abstract class MalloyTranslation {
 
   private buildImports(): void {
     for (const [key, value] of Object.entries(this.root.importZone.location)) {
+      // Ignore any import in another file
+      if (value.url !== this.sourceURL) {
+        continue;
+      }
       this.imports.push({importURL: key, location: value});
     }
   }
--- a/packages/malloy/src/lang/parse-malloy.ts
+++ b/packages/malloy/src/lang/parse-malloy.ts
@@ -82,7 +82,7 @@ import {
   isNeedResponse,
 } from './translate-response';
 import {MalloyParserErrorHandler} from './parse-error-handler';
-import { locationContainsPosition } from './utils';
+import {locationContainsPosition} from './utils';
 
 export type StepResponses =
   | DataRequestResponse
--- a/packages/malloy/src/lang/parse-malloy.ts
+++ b/packages/malloy/src/lang/parse-malloy.ts
@@ -791,22 +791,24 @@ export abstract class MalloyTranslation {
   }
 
   importAt(position: DocumentPosition): ImportLocation | undefined {
-    // Here we assume that imports DO NOT overlap. And then we do a binary
+    // Here we assume that imports DO NOT overlap. And then we do a linear
     // search to find the one we're looking for.
-    const index = this.findIndexBefore(position);
-    if (index === this.imports.length) {
-      return undefined;
-    }
-    const imp = this.imports[index];
-    if (
-      imp.location.range.start.line <= position.line &&
-      imp.location.range.end.line >= position.line &&
-      (position.line !== imp.location.range.start.line ||
-        position.character >= imp.location.range.start.character) &&
-      (position.line !== imp.location.range.end.line ||
-        position.character <= imp.location.range.end.character)
-    ) {
-      return imp;
+    for (let index = 0; index < this.imports.length; index++) {
+      const imp = this.imports[index];
+      // Ignore any import in another file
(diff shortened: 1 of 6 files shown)


packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -783,10 +785,80 @@ export abstract class MalloyTranslation {
...
+    return undefined;
+  }
+
+  private buildImports(): void {
+    for (const [key, value] of Object.entries(this.root.importZone.location)) {
+      this.imports.push({importURL: key, location: value});
+    }
+    // Sort list of imports by end position so that we can do a binary
+    // search to find an import at a given position.
+    this.imports.sort((i1, i2) => {
```
christopherswenson, line 820: The binary search stuff doesn't belong in the `parse-malloy.ts` file. Maybe you can make https://github.com/malloydata/malloy/blob/73b05828440fb2ffc1075312283c1e97ef9f9d50/packages/malloy/src/lang/reference-list.ts#L4 generic to support this use case.

christopherswenson, line 820: The binary search stuff doesn't belong in the `parse-malloy.ts` file. Maybe you can make https://github.com/malloydata/malloy/blob/73b05828440fb2ffc1075312283c1e97ef9f9d50/packages/malloy/src/lang/reference-list.ts#L4 generic to support this use case.

packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -783,10 +785,80 @@ export abstract class MalloyTranslation {
...
+    if (p1.character > p2.character) {
+      return 1;
+    }
+    if (p1.character < p2.character) {
+      return -1;
+    }
+    return 0;
+  }
+
+  private findIndexBefore(position: DocumentPosition): number {
```
christopherswenson, line 841: All of this can be removed if you make `ReferenceList` generic

packages/malloy/src/lang/test/parse.spec.ts
```diff
@@ -2772,6 +2772,27 @@ describe('source references', () => {
...
+        'internal://test/langtests/myfile.malloy':
+          'query: ab -> { project: * }',
+      },
+    });
+    const resp2 = m.translate();
+    expect(resp2.final).toBeDefined();
+    const im = m.importAt({line: 0, character: 10});
+    expect(im?.importURL).toEqual('internal://test/langtests/myfile.malloy');
+    expect(im?.location.url).toEqual('internal://test/langtests/root.malloy');
+    expect(im?.location.range.start).toEqual({line: 0, character: 7});
```
christopherswenson, line 2792: You should be able to just do `expect(im.location).toMatchObject(source.locations[0])`

packages/malloy/src/malloy.ts
```diff
@@ -621,6 +628,19 @@ export class Model implements Taggable {
...
+   *
+   * @param position A position within the document.
+   * @return An `ImportLocation` at that position if one exists.
+   */
+  public getImport(
+    position: ModelDocumentPosition
+  ): ImportLocation | undefined {
+    return this._importAt(position);
+  }
+
```
christopherswenson, line 643: This whole file looks great 👍

packages/malloy/src/model/malloy_types.ts
```diff
@@ -57,6 +57,11 @@ export interface DocumentPosition {
   character: number;
 }
 
+export interface ImportLocation {
+  importURL: string;
+  location: DocumentLocation;
+}
```
christopherswenson, line 63: 👍

christopherswenson (changes requested):
I think the overall strategy looks fine to me. I left a couple comments to clean things up a bit. This should probably also get a review from @mtoy-googly-moogly

packages/malloy/src/lang/test/parse.spec.ts
rmstar: fixed

rmstar:
> I think the overall strategy looks fine to me. I left a couple comments to clean things up a bit. This should probably also get a review from @mtoy-googly-moogly

Chatted with @mtoy-googly-moogly about this - he suggested removing the binary search logic for imports as the number of imports is likely be quite small, so it makes more sense to do a linear scan. OTOH, the number of references can be huge, so we still want to do a binary search on the ReferenceList.

packages/malloy/src/lang/parse-malloy.ts
rmstar: removed binary search code; this is obsolete

packages/malloy/src/lang/parse-malloy.ts
rmstar: removed binary search code; this is obsolete

packages/malloy/src/lang/parse-malloy.ts
```diff
@@ -783,10 +785,40 @@ export abstract class MalloyTranslation {
...
     return attempt;
   }
 
+  importAt(position: DocumentPosition): ImportLocation | undefined {
+    // Here we assume that imports DO NOT overlap. And then we do a linear
+    // search to find the one we're looking for.
+    for (let index = 0; index < this.imports.length; index++) {
+      const imp = this.imports[index];
+      if (
+        imp.location.range.start.line <= position.line &&
```
christopherswenson, line 799: Nit: it'd be nice to abstract this logic to a function `locationContainsPosition` or something like that, since this is repeated in the `ReferenceList` class.

rmstar: fixed

christopherswenson (approved):

PR #1269, opened 2023-08-01
Don't catch mid-line delimiter starts

whscullin:
Check that `>>>` is not preceded by `EOL`. Don't include `EOL` in statement body so notebooks don't have extra space.


christopherswenson (approved):
Yep, I think this looks right. 👍

PR #1277, opened 2023-08-08
Add bug report template

carlineng:
@whscullin anything you would like to add to the template?


whscullin (commented):
Contents look good to me. We could also put this in our https://github.com/malloydata/.github repo so that it covers all of our different repos so issues could live with the appropriate repo instead of all under Malloy.

carlineng:
Ah great idea. Didn't know that repo exists. I'll close this and move there.

PR #1289, opened 2023-08-12
Post Tag-pocalypse cleanup

mtoy-googly-moogly:
Now that the new tag langauge and api are working, it is time to delete the old.

Also some cleanup of code which called the old api and moved to the new api without taking advantage of the new api

### DIFF
--- a/packages/malloy-render/src/html/container.ts
+++ b/packages/malloy-render/src/html/container.ts
@@ -21,7 +21,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import {Explore, TagParse} from '@malloydata/malloy';
+import {Explore, Tag} from '@malloydata/malloy';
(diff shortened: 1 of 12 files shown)


mtoy-googly-moogly:
@narreola I would appreciate a review, (of the renderer changes) I looked over the renderer's use of tags just to see how the API was working for you and, mostly because you ported from the old API to the new API, you weren't really using the new API.

So I took a pass at "if this were being implemented now, how would it be written".

Naming was a problem because the data structure that as a whole lines with of properties is called a `Tag` now, not a `Tags`, so I changed the parameter being passed around to be called `tagged`, meaning "the tagged object" so things like `tagged.has('pivot')` reads well.

Also I noticed what a hassle it was to get an array of strings from a tag and so I wrote an accessor for that.

narreola:
LGTM, thanks for taking care of this.

narreola (approved):

Issue #1094, opened 2023-05-04
We need a way to cast to native types.


lloydtabb:
I think there are always going to be times that we want to cast to a native type in a database.  It would be great if there were some syntax for that in Malloy.  Maybe..

We just hit this on slack (again).  @christopherswenson will make it occur less, but I still think we're going to want to be able to control the types of output tables....

```
group_by: foo_int64 is foo::"int64"
```

mtoy-googly-moogly:
Yup .. `::QUOTED_THING` is one idea I've had too. We should talk about `CAST` and `SAFE_CAST` and wrap all three of these up in one "make casting more correct" language change. There's already an item on the roadmap for this.

PR #1308, opened 2023-08-21
SQL Cast syntax

christopherswenson:
Adds `somevalue::"sometype"` where "sometype" is a SQL type. That type is passed directly into the SQL translator.

Some things to think about:
- Yields an `unsupported`-typed value if the SQL type is not known by the dialect.
- Errors on some malformed types, i.e. those that include characters that don't appear in types in that dialect; this could some day be improved on a per-dialect basis to actually validate the types accurately; right now `int64[[[[` is allowed, for instance.


packages/malloy/src/lang/ast/expressions/expr-cast.ts
```diff
@@ -21,31 +21,37 @@
...
-export function isCastType(t: string): t is CastType {
-  return ['string', 'number', 'boolean', 'date', 'timestamp'].includes(t);
-}
-
 export class ExprCast extends ExpressionDef {
   elementType = 'cast';
   constructor(
     readonly expr: ExpressionDef,
-    readonly castType: CastType,
+    readonly castType: CastType | {raw: string},
```
christopherswenson, line 35: Is "raw" the right word here? Maybe `{ sqlType: string }`?

mtoy-googly-moogly:
Fixes #1094

PR #1310, opened 2023-08-21
Typescript 5

whscullin:
Update to Typescript 5. Mostly uneventful except for `tsutils` not having been updated to support Typescript 5 yet, but its functionality wasn't hard to replace. Main issue is that HTML table rendering snapshots are off, and it's fairly hard to determine why.


whscullin:
Closing in favor of https://github.com/malloydata/malloy/pull/1312, some Vega updates slipped in and caused rendering tests to fail, that's a different problem for a different PR.

PR #1314, opened 2023-08-26
Bump Node.js to v18 per Pending v16 EOL

abhillman:
Node.js 16.x will leave maintenance status 2022-09-11. Source: https://github.com/nodejs/release#release-schedule.

Per the OpenJS foundation, the 18.x series is currently recommended. Source: https://nodejs.org/en.

Resolves https://github.com/malloydata/malloy/issues/1313

PR #1317, opened 2023-08-30
Specific nixos/nix:2.15.2

abhillman:
**tl;dr**: this ensures that the latest package definitions are fetched from `nixpkgs`

---

We change the image to `nixos/nix:2.15.2` to ensure that we pull this specific image from the cache, which is populated with a more recent version of `nixpkgs`.

Updating `nixos/nix` references was done with the following:

```bash
git grep --null --files-with-matches "nixos/nix" |
  xargs --null -I{} sed -i 's/nixos\/nix/nixos\/nix:2.15.2/' {}
```

This will allow us to use the latest official package references; for example https://search.nixos.org/packages?channel=23.05&show=nodejs_18.

---

For a concrete example of why this is helpful, see https://github.com/malloydata/malloy/pull/1314#discussion_r1309498349.

### DIFF
--- a/cloudbuild/README.md
+++ b/cloudbuild/README.md
@@ -15,7 +15,7 @@ There are several technologies to be aware of:
 The way that we run tests:
 
 - we have a `cloudbuild.yaml` file that specifies the build -- see `cloudbuild/build-test/cloudbuild.yaml`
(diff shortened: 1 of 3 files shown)


abhillman:
cc/ @whscullin

PR #1332, opened 2023-09-10
Disallow asymmetric aggregate over repeated nested relationships

christopherswenson:
Accidentally pushed directly to `main`. This PR should logically contain: https://github.com/malloydata/malloy/commit/5dc9df70140966258d95e61d0aec4c199dbb3a1d

Expands on https://github.com/malloydata/malloy/pull/1325

Given:

* `nested` is a nested structure with `structRelationship = { type: 'nested', isArray: true }`
* `nested_inline` is a nested structure with `structRelationship = { type: 'nested_inline', isArray: false }`
* `inline` is a nested structure with `structRelationship = { type: 'inline' }`

These are the warnings/errors:

| Example | Warning / Error | Note |
| ---| -|-|
| `sum(nested.column)` | Error: Join path is required for this calculation; use `nested.column.sum()` | `nested.column` doesn't use a join, but it uses a nested column, and we require explicit locality here too; since `nested` is repeated, it's an error, not a warning |
| `nested.column.sum()` |  | |
| `source.sum(nested.column)` |  | Error: Cannot compute `sum` across repeated relationship `nested`; use `nested.column.sum()` |
| `sum(nested_inline.column)` | Warning: Join path is required for this calculation; use `nested_inline.column.sum()` or `source.sum(nested_inline.column)` to get a result weighted with respect to `nested_inline` | `nested_inline` is not repeated |
| `nested_inline.column.sum()` |  | |
| `source.sum(nested_inline.column)` |  | |
| `sum(inline.column)` | Warning: Join path is required for this calculation; use `inline.column.sum()` or `source.sum(inline.column)` to get a result weighted with respect to `inline` | `inline` is not repeated |
| `inline.column.sum()` |  | |
| `source.sum(inline.column)` |  | |


christopherswenson:
Closing in favor of https://github.com/malloydata/malloy/pull/1335

PR #1352, opened 2023-09-21
add * { except: } for wildcard expansion in select: *

mtoy-googly-moogly:
On the way to towards a more full featured star.

`select: * { except: badThing, otherBadThing }`

What's missing is "*" in index doesn't use this, and the user request was also to have rename, which is more code to implement

* [x] @lloydtabb 
* [x] @christopherswenson 
* [x] @carlineng

### DIFF
--- a/packages/malloy/src/lang/ast/field-space/query-spaces.ts
+++ b/packages/malloy/src/lang/ast/field-space/query-spaces.ts
@@ -132,6 +132,9 @@ export abstract class QuerySpace
     }
     const dialect = this.dialectObj();
     for (const [name, entry] of current.entries()) {
+      if (wild.except.has(name)) {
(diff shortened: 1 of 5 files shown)


lloydtabb:
I really think `replace:` is the key thing here not (rename).  I think we should be able to replace a field with an expression including that field without having to do it in two steps.

  https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_replace

as long as you are ok with someday including this functionality I'm good.

```
select: * { 
  except: badThing, otherBadThing 
  replaces:
     foo is FOO
     bar is UPPER(bar)
}
```

mtoy-googly-moogly:
I wrote a sentence about replaces: on the roadmap

PR #1368, opened 2023-09-26
Expose model imports

whscullin:
In order to do better cache invalidation, I need to know what files a model depends on, so when an upstream file changes, the current model recompiles (https://github.com/malloydata/malloy-vscode-extension/issues/258) and that info is not currently exposed.


mtoy-googly-moogly:
I could implement the "give me the whole list" ... do you really need to locations or would just a list of URL's be enough?

whscullin:
Just the URLs would be enough.

whscullin:
Closing in favor of https://github.com/malloydata/malloy/pull/1490

PR #1376, opened 2023-09-29
duckdb 0.9.2

whscullin:
Experiment with updating to duckdb 0.9.0. There was an issue with prebuilt binaries for 0.9.0 so this uses a recent dev build.

@lloydtabb / @christopherswenson - A bunch of UDF tests are failing, like "Multi value to udf". The results from DuckDB are lacking the "t1" property name. Not clear to me if this is a bug or a change in behavior we need to adapt to.

BigQuery:
```json
[
      {
        "fun": [
          {
            "t1": 52
          }
        ]
      }
    ]
```

DuckDB/DuckDB WASM:
```json
 [
      {
        "fun": [
          {
            "": 52
          }
        ]
      }
    ]
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4062,9 +4062,9 @@
       "link": true
     },
     "node_modules/@malloydata/duckdb-wasm": {
-      "version": "0.0.3",
-      "resolved": "https://registry.npmjs.org/@malloydata/duckdb-wasm/-/duckdb-wasm-0.0.3.tgz",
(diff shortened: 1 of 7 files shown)


whscullin:
Another failure is:
```
 modeled having complex - duckdb

    expect(received).toBe(expected) // Object.is equality

    Expected: 102
    Received: 178
```

lloydtabb:
I saw some stuff in their release notes about a change here.  It may be intentional but it sure is weird...

PR #1377, opened 2023-09-30
Take a swing at source/query expressions

mtoy-googly-moogly:
Trying to move to 4.0 has revealed some problems in the query grammar which are difficult to fix.

This moves to a much simplified query grammar, which a "smart" transformer, which is also difficult, but with a bounded difficulty.


mtoy-googly-moogly:
this got folded into #1383

Issue #1379, opened 2023-10-01
##! m4warning not recommending query: -> view:


lloydtabb:
Just testing.  Want to track for m4 release.

PR #1383, opened 2023-10-02
m4 warnings on ... new query grammar

mtoy-googly-moogly:
In preparation for M4, all the test code here has been moved forward to be strictly 4.0 Compatible. Attempting to do that revealed some holes in our pre-4.0 compatibility story, which is why this PR also includes a lot of work on the parsing.

This will be landed with `m4warnings` off, with the intention of a quick release a day or so later with the warnings turned on.


christopherswenson:
Fixes https://github.com/malloydata/malloy/issues/1379

PR #1393, opened 2023-10-04
feat: image properties for width, height, alt text

skokenes:
Adds support for the following properties for an image tag:
- width: set a width value to be applied to the image. Will also be applied to the null value placeholder to maintain consistent size
- height: set a height value to be applied to the image. Will also be applied to the null value placeholder to maintain consistent size
- alt: set alt text for the image

## Examples
Set width to 200px. All images will be 200px wide. The corresponding height will be according to the aspect ratio of the image. The null value placeholder will also be sized to 200px wide to maintain horizontal alignment in dashboards.
```
# image { width=200px }
```

Set height to 200px. All images will be 200px tall. Width will be according to the aspect ratio of the image. Null value placeholders will get the 200px height as well.
```
# image { height=200px }
```

Set width and height to 200px. All images will be given width and height of 200px. Note that this can warp images with the wrong aspect ratio.
```
# image { height=200px width=200px }
```

Set alt text to 'hello'. All images will be given the alt text of 'hello'.
```
# image { alt=hello }
```

### DIFF
--- a/packages/malloy-render/src/html/image.ts
+++ b/packages/malloy-render/src/html/image.ts
@@ -39,16 +39,20 @@ export class HTMLImageRenderer implements Renderer {
       );
     }
 
-    if (data.isNull()) {
-      return createNullElement(this.document);
-    }
+
+    let element: HTMLElement = data.isNull() ? createNullElement(this.document) : this.document.createElement("img");
     const { tag } = data.field.tagParse();
     const height = tag.numeric("image", "height");
     const width = tag.numeric("image", "width");
-    const element = this.document.createElement('img');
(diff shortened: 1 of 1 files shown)


lloydtabb (approved):
Looks great.  Maybe percent width?

whscullin:
Unfortunately that doesn't quite work, the `numeric()` will convert `100em` to `100` and you're unconditionally adding `px`. You probably need to retrieve the tag as a string to determine if it has units, then process accordingly.

skokenes:
Wow good catch, I will open a new PR. Thanks @whscullin

PR #1396, opened 2023-10-06
fix: text not numeric img size props

skokenes:
Previous PR for adding width/height attributes to images claimed you could pass valid CSS text values for size, but that was incorrect. The code was actually pulling out numeric values and hardcoding to pixels.

This PR updates to the desired behavior: user's can define image sizes in pixels, ems, percentages, etc.


whscullin (approved):
You may want to fix up unitless numbers, but this works like your description, so 👍

Issue #1404, opened 2023-10-11
Colab: rendering URLS is broken.


lloydtabb:
![image](https://github.com/malloydata/malloy/assets/1093458/2437ea91-8eed-4dfc-b166-d210061041d4)

PR #1405, opened 2023-10-12
Fix link renderer in Colab

srschandan:
Reference document from `this`.


srschandan:
Fix for https://github.com/malloydata/malloy/issues/1404

PR #1410, opened 2023-10-17
Stop using "project" and "turtle" in error messages

mtoy-googly-moogly:
There were still a few places left using the old words


mtoy-googly-moogly:
I've folded this into #1419

Issue #1364, opened 2023-09-26
native cast strings don't render


mtoy-googly-moogly:
```
run: duckdb.sql("SELECT 'abc'::VARCHAR(3) as abc")->{
  select: abc, also_abc is 'abc'::"VARCHAR(3)"
}
```

produces SQL

```SQL
SELECT 
   base."abc" as "abc",
   CAST('abc' AS VARCHAR(3)) as "also_abc"
FROM (SELECT 'abc'::VARCHAR(3) as abc) as base
```

JSON:

```json
[
  {
    "abc": "abc",
    "also_abc": "abc"
  }
]
```

but the result schema doesn't understand the datatype ... so it doesn't render ...

<img width="377" alt="image" src="https://github.com/malloydata/malloy/assets/66150587/18e82acd-8aa7-4751-a764-dcb82a5e47cc">

PR #1413, opened 2023-10-18
fix: capture VARCHARs with max length as strings

skokenes:
Fixes https://github.com/malloydata/malloy/issues/1364

The issue is that when we convert SQL types to Malloy types, we only check literal values like BOOLEAN, VARCHAR, etc. If a user casts to a VARCHAR with max length like `VARCHAR(3)`, this is not caught by our type conversion and results in an Unsupported Type, which then won't get rendered as a string.

The solution I've added here is to add an additional set of checks that use regex pattern matching so we can capture more complex SQL types than just literals. If this solution is acceptable, then I will implement it in the other connectors besides DuckDB.


mtoy-googly-moogly (approved):
🐑 🇮🇹

PR #1418, opened 2023-10-19
feat: storybook for render package

skokenes:
Adds a Storybook app to the render package for local development and experimentation with the renderer:
CleanShot 2023-10-19 at 15 03 58@2x

The previous workflow for working with the renderer was convoluted: you would essentially run a dev version of the VSCode extension as a proxy for running the latest renderer code. It required several steps across repositories and terminals to get it working properly.

This PR adds a Storybook that can be launched from the `malloy-render` package via `npm run storybook`. This launches an app that can read and render Malloy queries stored in the malloy-render package. It supports hot reloading, so any changes to the renderer will be instantly reflected in your browser.

This PR adds one basic story as an example. The goal is to expand this with stories for all of our chart types and variations on their settings.


packages/malloy-render/.storybook/main.js
```diff
@@ -0,0 +1,46 @@
...
+        preserveSymlinks: true,
+      },
+      define: {
+        'process.env': {},
+      },
+    });
+    return finalConfig;
+  },
+};
+export default config;
```
whscullin, line 46: It would be nice if this were also in typescript, we don't have any tooling that supports Flow. Besides a silly workaround for `'name': '@storybook/html-vite'` being hard-typed it seems trivial.

whscullin (approved):
Generally LGTM. Definitely blows up our node_modules but I guess that's the hazards of living in a monorepo.

packages/malloy-render/.storybook/main.js
skokenes: Done

PR #1421, opened 2023-10-21
4.0 warnings are now errors

mtoy-googly-moogly:


mtoy-googly-moogly:
This was helpful but I decided not to merge this, but fold it into #1422

PR #1429, opened 2023-10-23
feat: dynamic rendering props

skokenes:
Introduces the ability to have dynamic data lookup in the Malloy renderer, with the first implementation being with the `# image.alt` property:
image

The renderer can optionally read tags of `field` as being references to data within the current record or any record above. To indicate reading from a record above, `^` are used to indicate how many record levels in the results tree to go up before looking for a field value. For example, `field=brand` would look at a brand field in the current cell's record (assuming we land https://github.com/malloydata/whatsnext/blob/main/wns/WN-0005/wn-0005.md). `field='^brand'` would look up brand in the parent record of the current cell. `field='^^brand'` would look at the parent's parent record, so on and so forth. When navigating up the tree, array parents are ignored. This PR does not support trying to read from a parent array.

In the case of the `alt` tag, a static value set to `alt` is used as a fallback value for when a dynamic lookup is not provided or does not work. Other props may or may not want to use the static value in the same way.

## Functional changes this PR covers:
- To support looking up parent data, the `DataColumn` types are updated with references to their direct `parent`, which would include `DataArray`s, and their `parentRecord` which skips `DataArray` and provides the next ancestor `DataRecord` of a cell.
- Utility functions are provided in the renderer for resolving a dynamic value lookup
- The Image renderer supports looking up dynamic values for the `alt` tag

## Dev changes this PR covers:
- Modifies the storybook setup so that it will hot reload changes to the Malloy dependencies within the monorepo
- Updates the malloy-db-duckdb package to use a newer version of our duckdb-wasm package that resolves some bugs (thanks @whscullin)

### DIFF
--- a/packages/malloy-render/src/stories/static/data/logos.csv
+++ b/packages/malloy-render/src/stories/static/data/logos.csv
@@ -1,11 +1,11 @@
-id,brand,logo,foo
-1,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,a
-2,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,b
-3,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,c
-4,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,a
-5,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,b
-6,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,c
-7,Malloy,https://malloydata.github.io/img/logo.png,a
-8,Malloy,https://malloydata.github.io/img/logo.png,b
-9,Malloy,https://malloydata.github.io/img/logo.png,c
-10,BrokenBrand,bad_url,a
+id,brand,logo,product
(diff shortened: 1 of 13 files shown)


packages/malloy-render/src/stories/static/data/logos.csv
```diff
@@ -0,0 +1,11 @@
...
+1,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,gViz
+2,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,Plex
+3,Google,https://upload.wikimedia.org/wikipedia/commons/thumb/2/2f/Google_2015_logo.svg/2560px-Google_2015_logo.svg.png,Google Charts
+4,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,Looker
+5,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,Looker Studio
+6,Looker,https://upload.wikimedia.org/wikipedia/commons/thumb/4/4c/Looker.svg/2560px-Looker.svg.png,Looker Studio Pro
+7,Malloy,https://malloydata.github.io/img/logo.png,Malloy
+8,Malloy,https://malloydata.github.io/img/logo.png,Malloy Composer
+9,Malloy,https://malloydata.github.io/img/logo.png,Malloy Py
+10,BrokenBrand,bad_url,Viz
```
whscullin, line 11: I wonder if duckdb was being fussy about the final newline.

whscullin (approved):
LGTM

packages/malloy-render/src/stories/static/data/logos.csv
skokenes: Could be! I _think_ I tried without a final newline, but I'm not certain

christopherswenson:
LGTM.

I would love for this to have actual syntax in the tag language rather than using a property called `field` in the future... maybe we can start a discussion about this with @mtoy-googly-moogly

PR #1457, opened 2023-10-31
WN-0006 Scalar Lenses and support views with literal heads having no fields

christopherswenson:
Implements the experimental parts (scalar lense) of [WN 0006](https://github.com/malloydata/whatsnext/blob/main/wns/WN-0006/WN-0006.md)
* Feature requires experimental tag: `##! experimental { scalar_lenses }`
* Anywhere where you can reference a view name in a query, you can reference a dimension or measure instead
* E.g. `run: flights -> flight_count` is equivalent to `run: flights -> { aggregate: flight_count }`
* Can be used in views and nests as well
* Doesn't work with dotted names (yet)
* Particularly useful for shorthand query "column building" style: `flights -> carrier + flight_count + { where: origin.state = 'CA' }`

Adds support for the first literal view of a refinement chain not specifying any fields.
* Previously, `run: flights -> { where: true } + metrics` would fail with an error indicating that it couldn't tell what kind of query you wanted (index, project, reduce). 
* Now that query runs fine
* But `run: flights -> { where: true }` has the same error as before

### DIFF
--- a/packages/malloy/package.json
+++ b/packages/malloy/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@malloydata/malloy",
-  "version": "0.0.99",
+  "version": "0.0.98",
   "license": "MIT",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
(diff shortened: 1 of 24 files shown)


lloydtabb:
Things broken...  Can't nest a single scalar field.

```
run: airports -> faa_region + {
  # list
  nest: state
}
```

PR #1461, opened 2023-11-02
Fix code that skips tests that use `nest:` on databases that don't support nests

christopherswenson:
This code was clearly meant to be short-lived, but it also caused _all tests that use nests_ to be skipped, _regardless of whether the connection supports nesting_.


mtoy-googly-moogly (approved):
FIx is correct. Embarassed I did this.

PR #1463, opened 2023-11-02
Add flattened measure lists

skokenes:
## Feature

When rendering a nested query with no group_bys, you can get an ugly looking table with unnecessarily embedded headers:
```malloy
{
    group_by: category
    aggregate: avg_retail is retail_price.avg()
    nest:
      mens is {
        aggregate:
          avg_price is retail_price.avg()
          total_cost is cost.sum()
        where: department="Men"
      }
      womens is {
        aggregate:
          avg_price is retail_price.avg()
          total_cost is cost.sum()
        where: department="Women"
      }
 }
```
CleanShot 2023-11-02 at 12 19 50@2x

Using a `# flatten` tag, a nested query with no group_bys will automatically pull the column names of the nested measures up a level like so:
```malloy
{
  group_by: category
  aggregate: avg_retail is retail_price.avg()
  nest:
    # flatten
    mens is {
      aggregate:
        avg_price is retail_price.avg()
        total_cost is cost.sum()
      where: department="Men"
    }
    # flatten
    womens is {
      aggregate:
        avg_price is retail_price.avg()
        total_cost is cost.sum()
      where: department="Women"
    }
 }
```
CleanShot 2023-11-02 at 12 20 58@2x

This has limited support within pivoting; it will work within a pivot as long as none of the flattened measure names are repeated:
```malloy
# pivot
test is {
  group_by: distribution_center_id
  aggregate: avg_retail is retail_price.avg()
  nest:
    # flatten
    mens is {
      aggregate:
        avg_price is retail_price.avg()
        total_cost is cost.sum()
      where: department="Men"
    }
    # flatten
    womens is {
      aggregate:
        avg_price1 is retail_price.avg()
        total_cost1 is cost.sum()
      where: department="Women"
    }
  where: distribution_center_id=1 or distribution_center_id=2
}
```
CleanShot 2023-11-02 at 12 22 20@2x

However, there is a bug where if you try to flatten multiple queries with the same measure names within a pivot, it will not render properly. We will fix this bug later.

## Implementation notes
This isn't the cleanest implementation, but we are going to refactor this code soon so I did not spend much time here perfecting the architecture. We need this feature ASAP for demos.

### DIFF
--- a/packages/malloy-render/src/html/table.ts
+++ b/packages/malloy-render/src/html/table.ts
@@ -113,8 +113,6 @@ export class HTMLTableRenderer extends ContainerRenderer {
       return this.document.createElement('span');
     }
 
-    this.childRenderers;
-
     if (!table.isArray() && !table.isRecord()) {
       throw new Error('Invalid type for Table Renderer');
     }
@@ -249,9 +247,14 @@ export class HTMLTableRenderer extends ContainerRenderer {
         }
         pivotDepth = Math.max(pivotDepth, dimensions!.length);
       } else if (shouldFlatten) {
-        const exField = field as ExploreField;
-        const flattenedFields = exField.allFields.map(
-          f => new FlattenedColumnField(exField, f, `${exField.name} ${f.name}`)
+        const parentField = field as ExploreField;
+        const flattenedFields = parentField.allFields.map(
+          f =>
+            new FlattenedColumnField(
+              parentField,
+              f,
+              `${parentField.name} ${f.name}`
+            )
         );
         for (const flatField of flattenedFields) {
           cells[rowIndex][columnIndex] = this.createHeaderCell(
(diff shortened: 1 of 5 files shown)


packages/malloy-render/src/html/table.ts
```diff
@@ -190,12 +234,40 @@ export class HTMLTableRenderer extends ContainerRenderer {
...
+                childRenderer: nonDimension.flattenedField?.getChildRenderer(
+                  childRenderer.childRenderers
+                ),
+              }
             );
             columnIndex++;
           }
         }
         pivotDepth = Math.max(pivotDepth, dimensions!.length);
+      } else if (shouldFlatten) {
```
narreola, line 249: will you always be able to flatten here? is there any condition you'd like to check to see if flattening is a valid option?

narreola (commented):
I think is a good idea to add a test that makes sure this gets rendered correctly.

narreola (approved):

packages/malloy-render/src/html/table.ts
skokenes: Done

PR #1469, opened 2023-11-02
possible fix for esbuild problem

mtoy-googly-moogly:
```
% npm install @esbuild/darwin-x64
% git checkout package.json
% git add package-lock.json
% git commit -m"possible for for esbuild problem"
```


whscullin (approved):
:shipit:

PR #1480, opened 2023-11-07
WN-0006 Scalar Lenses Part II

christopherswenson:
Updates to the scalar lenses experiment feature —

* `nest: name_of_dimension` is equivalent to `nest: { group_by: name_of_dimension }`
* Using `join_name.field_name` is allowed in view refinement chains

### DIFF
--- a/packages/malloy/src/lang/ast/query-properties/nest.ts
+++ b/packages/malloy/src/lang/ast/query-properties/nest.ts
@@ -80,6 +80,9 @@ abstract class TurtleDeclRoot
         this.log(headEnt.error);
         reportWrongType = false;
       } else if (headEnt.found instanceof QueryField) {
+        if (this.turtleName.list.length > 1) {
+          this.turtleName.log('Cannot use view from join');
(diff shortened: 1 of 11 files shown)


packages/malloy/src/lang/ast/query-properties/nest.ts
```diff
@@ -266,21 +274,68 @@ export class NestReference
...
+        );
       }
-      this.log(
-        `Cannot use ${kind} field in a nest operation, did you mean to use ${useInstead} operation instead?`
-      );
     }
   }
 
+  makeEntry(fs: DynamicSpace): void {
+    if (fs instanceof QuerySpace) {
```
mtoy-googly-moogly, line 300: this looks so much like the code in expand turtle i wonder if some of this logic needs to get pushed up into spacefield or queryfieldspace

mtoy-googly-moogly (approved):
yeah this is good ... lots of work on the re-write but this seems correct to mt.

PR #1483, opened 2023-11-09
feat: renderer next setup

skokenes:
This PR sets up the foundation for the next generation Malloy renderer.

## Changes
- Lit installed for web components, storybook/types added for missing types
- New renderer deployed inside the legacy HTMLView behind a `## renderer_next` document tag
- New renderer wrapper implemented as a Lit web component
- Initial implementation of a basic table matching design specs from @tschomer 
- basic theme control with CSS variables
- Stories showing new renderer as well as ability to custom theme it

In the future, we will probably want to export the web component directly so that developers have flexibility to use it in different ways like so:
```typescript
import "@malloy/render";

// In vanilla js
const el = document.createElement("malloy-render");

// In Lit
html`<malloy-render></malloy-render>`

// In frameworks like React/Solid/Vue/etc, however they support web components
<MyReactApp>
  <malloy-render></malloy-render>
</MyReactApp>
```

### DIFF
--- a/packages/malloy-render/src/html/html_view.ts
+++ b/packages/malloy-render/src/html/html_view.ts
@@ -33,7 +33,7 @@ import {ContainerRenderer} from './container';
 import {createErrorElement} from './utils';
 import {MainRendererFactory} from '../main_renderer_factory';
 import {HTMLListRenderer} from './list';
-import '../component/table';
+import '../component/render';
 
 export class HTMLView {
   constructor(private document: Document) {}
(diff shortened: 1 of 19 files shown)


whscullin (approved):
LGTM

packages/malloy-render/src/renderer_types.ts
```diff
@@ -29,6 +29,7 @@ export interface RendererOptions {
   onDrill?: DrillFunction;
   titleCase?: boolean;
   queryTimezone?: string;
+  target?: HTMLElement;
```
skokenes, line 32: no longer needed

packages/malloy-render/src/html/html_view.ts
```diff
@@ -33,11 +33,19 @@ import {ContainerRenderer} from './container';
...
 import {HTMLListRenderer} from './list';
+import '../component/render';
 
 export class HTMLView {
   constructor(private document: Document) {}
 
   async render(result: Result, options: RendererOptions): Promise<HTMLElement> {
+    const isNextRenderer = result.resultExplore.modelTag.has('renderer_next');
+    if (isNextRenderer) {
+      const el = document.createElement('malloy-render');
```
skokenes, line 44: should be this.document

packages/malloy-render/src/stories/tables.stories.ts
```diff
@@ -0,0 +1,40 @@
+import {Meta} from '@storybook/html';
+import script from './static/tables.malloy?raw';
+import {createLoader} from './util';
+import './themes.css';
+
+const meta: Meta = {
+  title: 'Malloy Next/Tables',
+  render: ({classes}, context) => {
+    const el = document.createElement('malloy-render');
```
skokenes, line 9: is the malloy-render import missing here? How is this working...

PR #1485, opened 2023-11-09
Fix refinement output space issue

christopherswenson:
Before this fix, this query does the wrong thing:

```malloy
source: x is duckdb.sql("SELECT 1 AS n") extend {
  measure: c is count()
  view: v is { aggregate: c }
}
run: x -> v + {
  aggregate: e is c { where: false }
}
```

In particular, `e` should be `0` in the result, but it is `1` instead. This is because `c` (in the definition of `e`) is being treated as an output field, and the compiler gets confused when there is an output field with filters, because that shouldn't normally be possible.

This was due to an issue involving mutating the `SpaceEntry` that is copied from the refinement LHS `exprSpace` into the `QuerySpace` of the refinement RHS in order to specify that in the `QuerySpace` the field is in the output space. With a small refactor of the way that output space is recorded for `QuerySpace` fields, this should be fixed.

### DIFF
--- a/packages/malloy/src/lang/ast/field-space/join-space-field.ts
+++ b/packages/malloy/src/lang/ast/field-space/join-space-field.ts
@@ -21,15 +21,11 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import {FieldSpace} from '../types/field-space';
 import {Join} from '../query-properties/joins';
 import {StructSpaceField} from './static-space';
 
 export class JoinSpaceField extends StructSpaceField {
-  constructor(
-    readonly intoFS: FieldSpace,
-    readonly join: Join
-  ) {
+  constructor(readonly join: Join) {
     super(join.structDef());
   }
 }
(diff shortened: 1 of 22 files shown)


packages/malloy/src/lang/ast/field-space/join-space-field.ts
```diff
@@ -21,15 +21,11 @@
...
 
-import {FieldSpace} from '../types/field-space';
 import {Join} from '../query-properties/joins';
 import {StructSpaceField} from './static-space';
 
 export class JoinSpaceField extends StructSpaceField {
-  constructor(
-    readonly intoFS: FieldSpace,
-    readonly join: Join
-  ) {
```
christopherswenson, line 32: Not related, but noticed that `intoFS` was never used.

packages/malloy/src/lang/ast/query-properties/joins.ts
```diff
@@ -57,7 +57,7 @@ export abstract class Join
   note?: Annotation;
 
   makeEntry(fs: DynamicSpace) {
-    fs.newEntry(this.name.refString, this, new JoinSpaceField(fs, this));
```
christopherswenson, line 60: (Removal of `intoFS` argument)

PR #1498, opened 2023-11-14
Use configured project as default project

whscullin:
Right now if somebody configures a project id, it's used to normalize table paths, but not as a project id for the connection, which is what I think people expect. This uses the default project as the connection projectId, which I think is what is expected, unless somebody knows something I don't...


whscullin:
Meh, the default _billing_ project needs to be able to be set separately.

PR #1505, opened 2023-11-15
defaultProject -> projectId and projectId -> billingProjectId. The pr…

bporterfield:
…ojectId is used for table name normalization and the billingProjectId is now passed to the BQ SDK as its projectId

### DIFF
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -74,12 +74,12 @@ interface CredentialBody {
 }
 
 interface BigQueryConnectionConfiguration {
-  defaultProject?: string;
+  projectId?: string;
   serviceAccountKeyPath?: string;
   location?: string;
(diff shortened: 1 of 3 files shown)


packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -74,12 +74,12 @@ interface CredentialBody {
 }
 
 interface BigQueryConnectionConfiguration {
-  defaultProject?: string;
+  projectId?: string;
```
whscullin, line 77: It might be good to include doc strings like `/** This ID is used for table path normalization */` here, so that consumers of the connection will get this info in their IDE.

whscullin (approved):
LGTM

packages/malloy-db-bigquery/src/bigquery_connection.ts
bporterfield: 👍

PR #1509, opened 2023-11-16
feat: nested tables

skokenes:
Implements nested tables in the next-gen renderer according to our design spec
CleanShot 2023-11-16 at 14 00 40@2x

## Features
- Table cells have left and right gutters, defaulting to 15px side
- Gutter divs used instead of padding, because in the future we need to render content in those gutters while hiding text overflow within the cell content
- Gutter divs also required because in some scenarios, we want to remove the border from the gutter. This is mainly the case for nested tables. We don't want nested tables that are side by side to look like they are connected, so we don't draw borders in the gutters on the bookends of tables (nested or otherwise)
- Nested tables gutters should not cascade, so gutters aren't applied in the cells embedding a nested table
- Column widths must be hardcoded for nested columns to line up across parent rows. For now, we are hardcoding the width until we implement a sampling approach to dynamically determine a width per column

### DIFF
--- a/packages/malloy-render/.storybook/preview.ts
+++ b/packages/malloy-render/.storybook/preview.ts
@@ -3,7 +3,9 @@ import {Preview} from '@storybook/html';
 import registeredData from './registered_data.json';
 
 async function createConnection() {
-  const connection = new DuckDBWASMConnection('duckdb');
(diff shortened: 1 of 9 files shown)


whscullin (approved):
LGTM

PR #1513, opened 2023-11-21
feat: pinned headers

skokenes:
https://github.com/malloydata/malloy/assets/5554373/044345c6-6e42-4502-b48f-c5ce1d907282

## Changes
- Table now has a scroll wrapper around it. It will fit width/height to whatever it is embedded within
- Table uses a self-consuming context to determine if its at root or not.
- If at root, a set of sticky headers is drawn. The sticky headers are just another instance of the malloy-table embedded, but with only 1 row drawn at each level and all non-headers hidden
- On scroll, a class is added so that the sticky headers will change color. In order for this to work, sticky header versions of malloy-table are rendered within the parent root instead of its own ShadowDOM root
- For sticky headers, all pointer events are removed from base wrapper and then explicitly added to th cells. Otherwise, user can't interact with the content visible underneath the sticky header
- Malloy DataArray / DataRecord now caches calls to row() and cell(), rather than returning a new instance on every invocation. This was necessary for rendering tools downstream to have stable references to data points in order to not re-render unnecessarily
- Some typography tweaks with Inter
- Added an example story showing a more compact table, per Lloyd's feedback. This may become the default

### DIFF
--- a/packages/malloy-render/.storybook/preview-head.html
+++ b/packages/malloy-render/.storybook/preview-head.html
@@ -1,0 +1,2 @@
+<link rel="preconnect" href="https://rsms.me/" />
+<link rel="stylesheet" href="https://rsms.me/inter/inter.css" />
--- a/package-lock.json
+++ b/package-lock.json
@@ -4041,6 +4041,14 @@
       "resolved": "https://registry.npmjs.org/@lit-labs/ssr-dom-shim/-/ssr-dom-shim-1.1.2.tgz",
       "integrity": "sha512-jnOD+/+dSrfTWYfSXBXlo5l5f0q1UuJo3tkbMDCYA2lKUYq79jaxqtGEvnRoh049nt1vdo1+45RinipU6FGY2g=="
     },
+    "node_modules/@lit/context": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@lit/context/-/context-1.1.0.tgz",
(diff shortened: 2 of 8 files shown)


packages/malloy-render/.storybook/preview-head.html
```diff
@@ -0,0 +1,2 @@
+<link rel="preconnect" href="https://rsms.me/" />
+<link rel="stylesheet" href="https://rsms.me/inter/inter.css" />
```
christopherswenson, line 2: Wondering if we should package this CSS in our repo somehow so that rendering our CSS doesn't require connecting to their server.

christopherswenson (approved):
Looks great! Left one line comment, feel free to take it or leave it.

Also wondering stylistically whether there should be a vertical border on the sides of the header cells at the edges—it just looks a little odd to have a background color that meets the white background with no border in between. (E.g. on the left of `category`, `brand`, and `sku` in your example video, same on the rightmost columns as well.)

packages/malloy-render/.storybook/preview-head.html
skokenes: Inter is our suggested font for using Malloy, but we don't actually ship it as part of the bundled renderer. It is up to the developer using the renderer to load the font themselves if they want to use it. This is similar to what Google Material 3 does, where Roboto is their default font but it is on the app developer to bring that font in https://github.com/material-components/material-web/blob/main/docs/theming/typography.md

In the code snippet above, that is only executed as part of the Storybook that we use to preview our components, so that won't get shipped as part of Malloy renderer.

In terms of where to recommend loading the font from, it is also available on Google Fonts https://fonts.google.com/specimen/Inter?query=inter. I believe I loaded it directly from source in the Storybook because the latest version had just been released and I wanted to try some of the new number variants.

PR #1517, opened 2023-11-27
Fix noImplicitReturns and no implicitAny issues.

narreola:


narreola:
For context, google complains about these rules not being enabled, stories files are ignored so I was not forced to fix it for now.

PR #1518, opened 2023-11-27
fix drill for joined in tables.

srschandan:
1. Fix column name for joined in tables.
2. Move select to down below.


srschandan:
> It would be great if you could add some tests for this.

Done.

PR #1519, opened 2023-11-27
feat: number alignment

skokenes:
Adds right alignment for numeric fields in the new renderer

CleanShot 2023-11-27 at 13 44 53@2x

### DIFF
--- a/packages/malloy-render/src/component/table.ts
+++ b/packages/malloy-render/src/component/table.ts
@@ -86,19 +86,22 @@ const renderFieldContent = (
       .rowLimit=${options.pinnedHeader ? 1 : Infinity}
     ></malloy-table>`;
   }
-  if (options.pinnedHeader)
-    return renderCell(f, '', {
-      hideStartGutter: isFirstChild(f),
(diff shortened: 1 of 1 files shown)


packages/malloy-render/src/component/table.ts
```diff
@@ -86,19 +86,23 @@ const renderFieldContent = (
...
   }
-  if (options.pinnedHeader)
-    return renderCell(f, '', {
-      hideStartGutter: isFirstChild(f),
-      hideEndGutter: isLastChild(f),
-    });
-  return renderCell(f, row.cell(f).value, {
+  let value = row.cell(f).value;
+  if (options.pinnedHeader) value = '';
+  if (f.isAtomicField() && f.isNumber())
```
whscullin, line 91: Should this be an `else if`? Is it possible for `options.pinnedHeader` and `f.isAtomicField() && f.isNumber()`, in which case you'd be asserting `'' is number` (which still works because string has a `toLocaleString()` method, but might eventually have other side effects?)

whscullin (approved):
One pedantic comment otherwise LGTM.

PR #1523, opened 2023-11-29
feat: styling prep for VS Code ext

skokenes:
Adds a few style changes to support VS Code extension integration and testing

- Adds and exposes `container` part so that table wrapper can be styled with CSS; needed to control max height of table in a notebook
- Adds a compact mode setting so that we can easily test some variations on the table design with real data
- Cleans up some of the number typography


whscullin (approved):
LGTM

PR #1525, opened 2023-11-30
Add lens support for imports

whscullin:
We want to enable lenses to preview sources to `import` statements, so add that information to the list of symbols returned from the document symbol walker.

Also enable removing a hack in the extension to retrieve `NamedQuery` objects for a model for schema views.


whscullin:
@srschandan this is the parts I'm adding to support `import` lenses.

PR #1527, opened 2023-11-30
feat: theme overrides

skokenes:
Adds the ability to override renderer themes directly from a malloy file, either with model or view tags.

Examples:
```malloy
-- every table in model will have blue text
## renderer_next theme.tableBodyColor=blue

source: products is duckdb.table('products.parquet') extend {
  ...
  
  -- This table will have red text
  # theme.tableBodyColor=red
  view: local_override is { select: * }
}
```

The order of precedence for a theme setting is:
1. View theme override with `# theme.prop=x`
2. Model theme override with `## theme.prop=x`
3. CSS override with `malloy-render { --malloy-theme--prop: x; }`
4. The default style value for the given prop, defined internally by malloy


whscullin (approved):
LGTM

PR #1528, opened 2023-12-01
feat: number formatting

skokenes:
Adds number formatting from old renderer to new renderer: `number`, `currency`, `percent`, and `duration`
image

For the number formatting logic in render-numeric-field.ts, the logic is just copied out of the legacy renderer code. I opted not to create separate renderer components for number formats for now; it seems wasteful to create a web component for every numeric cell in a table if it doesn't have any need for state, lifecycle events, etc.

So for now, our formatting just takes a numeric value and field def and returns a string of what it should look like. If our requirements grow such that these numbers need more complex features, we can refactor into a component.

### DIFF
--- a/packages/malloy-render/src/component/table.ts
+++ b/packages/malloy-render/src/component/table.ts
@@ -27,6 +27,7 @@ import {customElement, eventOptions, property, state} from 'lit/decorators.js';
 import {classMap} from 'lit/directives/class-map.js';
 import {createContext, provide, consume} from '@lit/context';
 import {isFirstChild, isLastChild} from './util';
+import {renderNumericField} from './render-numeric-field';
 
 type TableContext = {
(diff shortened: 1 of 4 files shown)


packages/malloy-render/src/component/table.ts
```diff
@@ -88,10 +89,11 @@ const renderFieldContent = (
       .rowLimit=${options.pinnedHeader ? 1 : Infinity}
     ></malloy-table>`;
   }
-  let value = row.cell(f).value;
+  let value: number | string = row.cell(f).value as number;
```
whscullin, line 92: I wonder if it's possible to write a guard for these values so that it's not necessary to make assumptions about their types.

whscullin (approved):
LGTM

PR #1533, opened 2023-12-02
Add the ability to do INNER, RIGHT and FULL joins.

lloydtabb:
There are a number of problems we keep encountering because we can't do these types of joins.  They are not usually important in semantic models, but seem to be pretty important when querying models.

The UNREVIEWED syntax for this feature.  Defaults to a left join.

```
##! experimental.join_types
join_one: right_table [left|inner|right|full] on x=right_table.x
```

Because the behavior is kind of weird and unexpected in a `source:`, perhaps we only allow this syntax in a `query:`.  But maybe would also be weird.


mtoy-googly-moogly:
We talked about maybe adding tests to the parser tests for this, ( which would be in packages/malloy/src/lang/test/source.spec.ts ) ... but actually don't do that. The syntax is going to change, while it is experimental the test you have written are great.

Michael says merge is fine with me. Let's get to learning about this stuff.

christopherswenson:
👍🏼

PR #1539, opened 2023-12-06
Implement a walker for finding table paths

srschandan:
Implement a walker to get the full table path.
Enable connection object to provide a browsable table source.


packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -769,6 +778,12 @@ export class BigQueryConnection
     return url;
   }
 
+  public getTableSourceUrl(tablePath: string) {
+    const tablePathInfo = tablePath.split('.');
+    const url = `https://pantheon.corp.google.com/bigquery?mods=allow_workbench_image_override&project=${this.projectId}&ws=!1m5!1m4!4m3!1s${tablePathInfo[0]}!2s${tablePathInfo[1]}!3s${tablePathInfo[2]}`;
```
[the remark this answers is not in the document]
srschandan: Publicly accessible, of course it'll redirect to auth. However, BigQuery doesn't provide a standardized URL for pointing to tables, this was a trial and error effort.
Configurable would make sense too but there's not a lot that can be done (or configured) in the sense that here's a URL that'll take you the table definition.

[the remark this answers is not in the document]
whscullin: It redirects you to google corp auth, not public corp auth. This doesn't work on my personal machine.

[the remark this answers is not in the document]
srschandan: Oh yes, thanks. Updated.

packages/malloy/src/lang/parse-tree-walkers/find-table-path-walker.ts
```diff
@@ -0,0 +1,66 @@
...
+import {getId, getPlainString} from '../parse-utils';
+import {MalloyTranslation} from '../parse-malloy';
+import {CommonTokenStream} from 'antlr4ts';
+import {DocumentRange} from '../../model/malloy_types';
+import * as parser from '../lib/Malloy/MalloyParser';
+import {MalloyParseInfo} from '../malloy-parse-info';
+import {ParseTreeWalker} from 'antlr4ts/tree/ParseTreeWalker';
+
+export interface PathInfo {
+  connId: string;
```
christopherswenson, line 34: I would prefer a full word here, `connectionId`.

packages/malloy/src/lang/parse-tree-walkers/find-table-path-walker.ts
```diff
@@ -0,0 +1,66 @@
...
+class FindTablePathWalker implements MalloyParserListener {
+  pathInfos: PathInfo[] = [];
+  constructor(
+    readonly translator: MalloyTranslation,
+    readonly tokens: CommonTokenStream
+  ) {}
+
+  enterTableMethod(pcx: parser.TableMethodContext): void {
+    const connId = getId(pcx.connectionId());
+    const tablePath = getPlainString(pcx.tablePath());
```
christopherswenson, line 48: It seems like the functionality of `getPlainString` is not super clear to the caller. It looks like it returns the non-interpolated parts of a string, which might be strange behavior in this particular context.

If you write `duckdb.table("""foo.bar.baz%{foo}""")` you will get a parser error, but then we will attempt to show you table path info for `"foo.bar.baz"` which seems wrong.

Maybe you should rework this to just exclude table calls that have interpolations.

packages/malloy/src/lang/test/find-table-path-walker.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import {TestTranslator, markSource} from './test-translator';
+
+test('Table path can be retrieved', () => {
+  const source = markSource`source: flights is DB.table('my.table.flights')
+  source: flights2 is flights extend { dimension: a is astr }`;
+  const doc = new TestTranslator(source.code);
+  const {pathInfo} = doc.tablePathInfo();
+  expect(pathInfo?.length).toBe(1);
```
christopherswenson, line 31: Can you modify this test to actually check the relevant parts of the returned table info?

I would check:
- Connection ID text
- Table Path text
- Position in document

For the position in the document, you can change the `markSource` call to be 
```
const source = markSource`
  source: flights is ${"DB.table('my.table.flights')"}
  source: flights2 is flights extend { dimension: a is astr }
`;
``` 
then compare `source.locations[0]` to `pathInfo[0].range`

packages/malloy/src/malloy.ts
```diff
@@ -978,6 +991,39 @@ export class Parse {
...
+  /**
+   * @return The range of characters in the source Malloy document that defines
+   * this table.
+   */
+  public get range(): DocumentRange {
+    return this._range;
+  }
+
+  /** @return The Connection Id for this table. */
+  public get connId(): string {
```
christopherswenson, line 1017: Same as above, `connectionId`

packages/malloy/src/runtime_types.ts
```diff
@@ -193,6 +196,10 @@ export interface StreamingConnection extends Connection {
   ): AsyncIterableIterator<QueryDataRow>;
 }
 
+export interface BrowsableSourceConnection extends Connection {
```
christopherswenson, line 199: This feels to me like a bit of a strange thing to include in our official API.

Is this something that other users of the Malloy library (other than VSCode) care about using?

If not, I would say this should be implemented in the VSCode extension. To do that, just move this interface into the VSCode extension and then instead of having the `Conection.browsableSource` method, have a function `isBrowsableSourceConnection(connection: Connection): connection is BrowsableSourceConnection`.

christopherswenson, line 199: This feels to me like a bit of a strange thing to include in our official API.

Is this something that other users of the Malloy library (other than VSCode) care about using?

If not, I would say this should be implemented in the VSCode extension. To do that, just move this interface into the VSCode extension and then instead of having the `Conection.browsableSource` method, have a function `isBrowsableSourceConnection(connection: Connection): connection is BrowsableSourceConnection`.

packages/malloy-db-bigquery/src/bigquery_connection.ts
```diff
@@ -769,6 +778,12 @@ export class BigQueryConnection
     return url;
   }
 
+  public getTableSourceUrl(tablePath: string) {
+    const tablePathInfo = tablePath.split('.');
+    const url = `https://console.cloud.google.com/bigquery?ws=!1m5!1m4!4m3!1s${tablePathInfo[0]}!2s${tablePathInfo[1]}!3s${tablePathInfo[2]}`;
+    return url;
+  }
+
```
christopherswenson, line 786: This implementation also sort of feels like it shouldn't be in our official API. Maybe the whole `BrowsableSourceConnection` stuff should just entirely be moved into the VSCode extension...

christopherswenson (changes requested):
Great work, this looks pretty close. I left a few comments.

Also I thought maybe a good way to get the behavior you wanted (where you hover over a source and get this information) is to combine this with the info from the document symbols to make it so that when you hover over a source, you get information about all the table calls that appear inside that source definition; so in the normal case of `source: foo is conn.table('foo')` you get just the table you care about, and for a complicated case where the source joins in other tables it just shows you all of them.

packages/malloy/src/runtime_types.ts
srschandan: I was thinking about users who would want to build their own connection, this is definitely doable in VS Code, which I did in my first iteration of this. However if a user builds their own connection then VS Code has to be changed to include that connection to support this.

packages/malloy/src/lang/parse-tree-walkers/find-table-path-walker.ts
srschandan: Done.

packages/malloy/src/malloy.ts
srschandan: Done.

packages/malloy/src/runtime_types.ts
srschandan: removed

srschandan: removed

packages/malloy-db-bigquery/src/bigquery_connection.ts
srschandan: removed

packages/malloy/src/lang/parse-tree-walkers/find-table-path-walker.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+class FindTablePathWalker implements MalloyParserListener {
+  pathInfos: PathInfo[] = [];
+  constructor(
+    readonly translator: MalloyTranslation,
+    readonly tokens: CommonTokenStream
+  ) {}
+
+  protected getPlainString(cx: HasString): string | undefined {
```
christopherswenson, line 52: It might be nice to abstract this out into a utility function in `parse-utils`, since `malloy-to-ast` has almost exactly the same function.

christopherswenson (commented):
This looks great apart from the one suggested refactor!

packages/malloy/src/lang/parse-tree-walkers/find-table-path-walker.ts
srschandan: Done

whscullin (approved):

PR #1545, opened 2023-12-06
New drill icon in the legacy renderer.

srschandan:
VS Code light:
Screenshot 2023-12-06 at 1 52 54 PM

VS Code dark:
Screenshot 2023-12-06 at 1 53 13 PM

### DIFF
--- a/packages/malloy-render/src/html/table.ts
+++ b/packages/malloy-render/src/html/table.ts
@@ -413,7 +413,7 @@ export class HTMLTableRenderer extends ContainerRenderer {
         const drillIcon = createDrillIcon(this.document);
         drillCell.appendChild(drillIcon);
         drillCell.style.cssText = `
-          padding: 8px;
+          padding: 2px;
           vertical-align: top;
           border-bottom: 1px solid var(--malloy-border-color, #eaeaea);
           width: 25px;


srschandan:
> I wonder if it would be good to have this externally configurable, since the action defined by drilling is configurable and won't likely be copy in the long run.

Agree. This was supposed to go into the new renderer but that the design and functionality is not finalized yet. This is more like get a better icon in there right now to make it look better.

skokenes (changes requested):
Code looks good, but this implementation causes the text in the table cells to no longer be vertically centered. Can you fix that?

srschandan:
> Code looks good, but this implementation causes the text in the table cells to no longer be vertically centered. Can you fix that?

Our cell contents are always aligned to top. https://github.com/malloydata/malloy/blob/49c4b9b69ef557358d2a4653e3b8658abc112796/packages/malloy-render/src/html/table.ts#L417

I reduced the padding on the drill icon to make sure when there's only one line of data in the table, everything looks center aligned.

skokenes:
It's weird, on multiline cells in VS Code extension it looks aligned with top line:
CleanShot 2023-12-08 at 09 39 03@2x

But on single line rows, it looks slightly off:
CleanShot 2023-12-08 at 09 40 14@2x

I don't think this really matters for now; we don't do solid control over the row height, font size etc in the old renderer so that will make it difficult to have it aligned properly to the first line in every situation.

This looks good to me

PR #1547, opened 2023-12-08
Refactor query AST

christopherswenson:
Expecting this to get about a million test failures still.

Removes some pre Malloy 4.0 features:
- `refine` is no longer a keyword and does not work in place of `+`
- Implicit (old-style) source extension no longer works (e.g. `source: flights is duckdb.table('flights.parquet') { ... }`)
- Using `+` to mean source extension no longer works
- Implicit (old-style) refinement no longer works (e.g. `flights -> by_carrier { where: state = 'CA' }`)
- `from` is no longer legal
- `from_sql` is no longer legal
- Leading arrow in front of query names is illegal (e.g. `run: -> some_query`)
- Views with leading arrows are illegal (e.g. `view: foo is -> { ... }`)

Known issues:
- Using a `ViewRefine` as a refinement for a multi-stage query is not allowed, even if you could write that as a single-operation refinement by hand, e.g. `run: x -> two_stage + ({ where: true } + { limit: 3 })` vs `run: x -> two_stage + { where: true; limit: 3 }`

Gets rid of `pipeHead` in the IR for `TurtleDef`s.


packages/malloy/src/malloy.ts
```diff
@@ -891,24 +890,10 @@ export class PreparedQuery implements Taggable {
...
-      name:
-        'as' in this._query ? this._query.as || this._query.name : defaultName,
-    });
-    return new PreparedQuery(
-      {...turtleDef, structRef, type: 'query'},
-      this._modelDef,
-      this.problems,
-      this.name || turtleDef.as || turtleDef.name
-    );
+  public getFlattenedQuery(_defaultName: string): PreparedQuery {
```
christopherswenson, line 895: So it turned out that `flattenQuery` from `malloy_query.ts` also copied filters on the source into the turtle def. I don't think that's desired here, or used at all by the users of this function, but I'm not 100% sure. I'll have to check that the VSCode usage of this function still works.

christopherswenson:
@mtoy-googly-moogly I don't understand why but removing `refine` as a keyword in the lexer is breaking all my lang tests locally: https://github.com/malloydata/malloy/pull/1547/commits/f55558916a35f6fa77252987eecdeda397b2218b

packages/malloy/src/malloy.ts
christopherswenson: Looks like it's only used by Composer actually

packages/malloy/src/lang/ast/field-space/view-field.ts
```diff
@@ -0,0 +1,51 @@
...
+import {QueryFieldDef, TurtleDef} from '../../../model/malloy_types';
+import {ViewDefinition} from '../source-properties/view';
+import {FieldSpace} from '../types/field-space';
+import {QueryField} from './query-space-field';
+
+// TODO naming: There's two subclasses of QueryField: ViewField and QueryFieldStruct
+// All three names should be changed. Maybe:
+// QueryField -> ViewField
+// ViewField -> ASTViewField
+// QueryFieldStruct -> IRViewField
```
christopherswenson, line 33: Naming discussion here

christopherswenson:
Reminder to self to look at `nestParent` mutation again.

PR #1566, opened 2024-01-04
feat: basic nested bar chart

skokenes:
This PR adds a very basic nested bar chart to the new renderer.

## Limitations of this PR
In order to keep this PR from being massive, many necessary features and implementation details have been left out as future TODOs. Those include the following:
- this PR only implements bar chart as nested within a table cell; root level bar chart renderings do not work yet
- the bar chart implementation has its own bespoke Vega spec. We will rearchitect this to use a more generalized XY Plot components like the old renderer does. Starting just with bar here to keep things simple
- the extraction of which fields are X and Y in the bar chart is rudimentary
- many of the styles needed to calculate size and render the bar charts are hardcoded, instead of referencing the theme
- only supports strings in X axis and numbers in Y axis
- Vega-Lite specs are typed as `any`. [Vega-Lite does not currently export the sub-types of their spec](https://github.com/vega/vega-lite/issues/9222). We are working towards a resolution on this but in the meantime need to rely on `any`

## PR Summary
- Adds the ability to render nested bar charts within a table using `# bar` annotations
- Bar charts attempt to read the first column as a string for X and the second column as a number for Y
- Bar charts can be sized using preset dimensions via tags with `# size=spark|sm|md|lg|xl|2xl`. Alternatively, a user can provide a custom size in pixels using `# size.width=200 size.height=200`. The size is not implemented as a sub-property of the `bar` tag because we assume that all charts, cells, etc will use this common size property. But we can change that if needed.
- Bar charts share numeric y axis domains, but not x axis domains.
- Bar charts are expected to closely align with the surrounding rows of a table. See the Specs section below. When calculating chart layout settings, we do some math to make sure that top and bottom axis lines will always align with adjacent table row lines and that overall chart sizes snap to the row grid of a table
- X axis text can either be rotated 0 degrees or 90 degrees. Whether to rotate or not is determined by the max size of text and the available space. When rotated, we limit how big the axis can grow in relation to the total height of the chart and use truncation otherwise.

## Spec
image

## Demo
[The render book is updated here with some examples.](http://go/malloy-render-book)

## Snapshots
CleanShot 2024-01-04 at 11 22 12@2x
CleanShot 2024-01-04 at 11 22 29@2x

### DIFF
--- a/packages/malloy-render/src/component/bar-chart.ts
+++ b/packages/malloy-render/src/component/bar-chart.ts
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files
+ * (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge,
+ * publish, distribute, sublicense, and/or sell copies of the Software,
+ * and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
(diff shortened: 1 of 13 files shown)


packages/malloy-render/src/component/bar-chart.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  @consume({context: resultContext})
+  @property({attribute: false})
+  metadata!: RenderResultMetadata;
+
+  override render() {
+    const field = this.data.field as ExploreField;
+    const keys = field.allFields;
+    const {tag} = field.tagParse();
+    const isSpark = tag.text('size') === 'spark';
```
narreola, line 47: NIT: sounds like a good candidate to have some utils for... containsTag?

narreola:
LGTM, can you please make sure to make any required changes to google3 BUILD files, I suspect since there are new files we need to include them.

packages/malloy-render/src/component/bar-chart.ts
skokenes: I don't know that we need wrapper utils for the tag APIs themselves, they are pretty robust if you look at what they can do.

But I do find it annoying to have to do `tagParse()` to get the tag, wish I could just call `field.tag.whatever` directly. I'll check with @mtoy-googly-moogly

packages/malloy-render/src/component/bar-chart.ts
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright 2023 Google LLC
```
whscullin, line 2: Happy New Year!

whscullin (approved):
LGTM

Issue #1560, opened 2023-12-21
Support something similar to PARTITION BY in SQL window functions


carlineng:
Right now, analytic functions do not have the ability to contain `PARTITION BY` clauses to define the window on which they operate. Traditionally, we've recommended that people use nesting in place of `PARTITION BY` clauses, but this can result in some very awkward-looking Malloy queries. Take the following example:

I have a table that contains sales data over several years:
![image](https://github.com/malloydata/malloy/assets/79773/caa8dc37-d73e-4a3e-97bd-d81a9e83826a)

I want to calculate sales for each month, and compare to sales for that month in the prior year (i.e., a simple year-over-year analysis).

In SQL, I can write this like so:

```sql
select
  yr
  , mo
  , sum(rev) as this_year_rev
  , lag(sum(rev)) over (partition by mo order by yr) as last_year_rev
from 'output.parquet'
group by 1,2;
```

In Malloy, the same query must be written like so:
```
run: duckdb.table('output.parquet') -> {
  group_by: mo
  nest: n is {
    group_by: yr
    aggregate: this_year_rev is sum(rev)
    calculate: last_year_rev is lag(this_year_rev)
    order_by: yr asc
} -> {
  select: n.yr, mo, n.this_year_rev, n.last_year_rev
}
```

This is a really awkward query because it's forcing the user to think about nesting, but the desired output isn't a nested resultset. Instead, I'd like to be able to express something like the following (imagined syntax):

```
run: duckdb.table('output.parquet') -> {
  group_by: yr, mo
  aggregate: total_rev is sum(rev)
  calculate: l is lag(total_rev) { partition_by: mo, order_by: yr asc }
}
```

This fits my mental model of the query much better. The aggregate is happening at the year/month level, and the calculate is subsequently applied.

lloydtabb:
I'm 100%, we should do this.  @christopherswenson @mtoy-googly-moogly

Issue #1557, opened 2023-12-15
Support STRING_AGG aggregation function


carlineng:
One of the most common requests for missing functions that we do not yet support is `STRING_AGG`.

See BQ docs for behavior: https://cloud.google.com/bigquery/docs/reference/standard-sql/aggregate_functions#string_agg

PR #1568, opened 2024-01-04
Add `string_agg` function, `partition_by` and `order_by` for analytic functions, and `partition_by` and `limit` for `string_agg`

christopherswenson:
Changes:
- Adds a `string_agg` function with two overloads: `string_agg(field)` and `string_agg(field, separator)`. If `separator` is not specified, it will be `,`. Also adds a `string_agg_distinct` which is the same as `string_agg`, but distinct.
- Adds, currently for use only with `string_agg`, the ability to specify an `order_by` and `limit` for an aggregate function, similar to the way you specify filters for an aggregate expression: `string_agg(field, sep) { order_by: field, limit: 10 }`. These are under experiments `function_order_by` and `aggregate_limit` respectively. (These names were chosen because `order_by` also applies to window functions, whereas `limit` only applies to aggregates.)
- Adds, for all window functions, the ability to specify `order_by` and `partition_by` in the same way as above. This is under the experiment `partition_by`.

For both aggregates and window functions, `order_by` is a comma-required-separated list of expressions followed by an optional `asc` or `desc`. This is different from a query's `order_by`, which is a comma-optional-separated list of field names or integers (also with `asc`/`desc`). This difference comes from the fact that `order_by` in a query needs to be a field in the output space, whereas `order_by` in an aggregate or window function can be an expression.

Specifically, `order_by` in an aggregate (i.e. `string_agg`) can be any input-space expression, and `order_by` in an analytic function must either be an aggregate expression or an output field reference. These requirements are correctly detected by the parser.

`partition_by` is a comma-required-separated list of non-dotted identifiers (referring to fields in the output space). The comma is required because for `select` views, expressions should actually be allowed in `partition_by`, so commas will be required down the line anyway.

`limit` must be a literal integer.

Known limitations:
- When using `string_agg_distinct` with an `order_by`, the order by must be exactly the same as the first parameter; this is currently a SQL error not caught by Malloy
- A problem for future me and Michael is that currently `limit` is only allowed for BQ, and there's some funkiness with the way functions are parsed/compiled to make sure there's a good error message in the other dialects. There is an existing similar issue that if an overload doesn't exist for one dialect, it gets a gross internal error. We need to give this some more thought.

Known bugs:
- `string_agg`'s `order_by` currently doesn't get added to the array agg unnest, meaning `order_by` doesn't work properly when there's a fanout

Future changes:
- Change the `string_agg` functions to use a custom fragment to produce a custom SQL compile with the following rules:
  - A regular `string_agg` with no `order_by` and no fanout just uses the `STRING_AGG` function
  - If a `string_agg` has a fanout or an `order_by`, ordering expression and/or the fanout distinct key are put into a `concat("MAGIC_PREFIX", ordering_expression, "MAGIC SEPARATOR", distinct_key, "MAGIC_SUFFIX", thing_you_wanted_to_string_agg)`, then that expression is `STRING_AGG(DISTINCT)`ed, then the `MAGIC_PREFIX.*?MAGIC_SUFFIX`es are removed with `REGEXP_REPLACE`. This allows BigQuery to do fanout `string_agg` calls and removes the bug with `string_agg`'s `order_by` not being added to the array agg unnest.
- Add the syntax `{ order_by: asc }` ( and `{ order_by: desc }`) for `string_agg_distinct` to solve the issue of "the order by must be exactly the same as the first parameter". For `string_agg` and analytic functions, `order_by: asc` is an error, and for `string_agg_distinct`, `order_by: anything_else` is an error. 

Fixes https://github.com/malloydata/malloy/issues/1557
Fixes https://github.com/malloydata/malloy/issues/1560


christopherswenson:
Probably will not merge without implementing [WN-0011](https://github.com/malloydata/whatsnext/blob/main/wns/WN-0011-field-properties/wn-0011.md), assuming it is accepted.

packages/malloy/src/dialect/dialect_map.ts
```diff
@@ -121,6 +121,8 @@ export function getDialectFunction(name: string): FunctionDef | undefined {
             needsWindowOrderBy: overload.needsWindowOrderBy,
             between: overload.between,
             isSymmetric: overload.isSymmetric,
+            supportsOrderBy: overload.supportsOrderBy,
+            supportsLimit: overload.supportsLimit,
```
christopherswenson, line 125: TODO remove

packages/malloy/src/dialect/functions/util.ts
```diff
@@ -37,6 +37,8 @@ export interface DialectFunctionOverloadDef {
   e: Expr;
   needsWindowOrderBy?: boolean;
   isSymmetric?: boolean;
+  supportsOrderBy?: boolean;
+  supportsLimit?: boolean;
```
christopherswenson, line 41: TODO remove

packages/malloy/src/dialect/functions/util.ts
```diff
@@ -213,6 +215,9 @@ export function overload(
   options?: {
     needsWindowOrderBy?: boolean;
     between?: {preceding: number | string; following: number | string};
+    isSymmetric?: boolean;
+    supportsOrderBy?: boolean;
```
christopherswenson, line 219: TODO remove

packages/malloy/src/dialect/functions/util.ts
```diff
@@ -221,5 +226,8 @@ export function overload(
     e,
     needsWindowOrderBy: options?.needsWindowOrderBy,
     between: options?.between,
+    isSymmetric: options?.isSymmetric,
+    supportsOrderBy: options?.supportsOrderBy,
```
christopherswenson, line 230: TODO remove

packages/malloy/src/dialect/postgres/functions/string_agg.ts
```diff
@@ -0,0 +1,54 @@
...
+  makeParam,
+  literal,
+} from '../../functions/util';
+
+export function fnStringAgg(): DialectFunctionOverloadDef[] {
+  const value = makeParam('value', maxScalar('string'));
+  const separator = makeParam('separator', literal(maxScalar('string')));
+  // const ob: Fragment = {type: 'function_order_by'};
+  // const lim: Fragment = {type: 'function_limit'};
+  return [
```
christopherswenson, line 40: TODO update this with new params

packages/malloy/src/model/malloy_query.ts
```diff
@@ -401,6 +402,18 @@ class QueryField extends QueryNode {
     );
   }
 
+  private generateSQLAggregateOrderBy(orderBy: OrderBy[]): string {
+    const o: string[] = [];
+    for (const f of orderBy) {
+      // TODO not sure how to validate field?
```
christopherswenson, line 408: TODO del

christopherswenson:
Fixes https://github.com/malloydata/malloy/issues/1560

test/src/databases/all/functions.spec.ts
```diff
@@ -1117,6 +1119,388 @@ expressionModels.forEach((expressionModel, databaseName) => {
...
+          `run: aircraft -> {
+            where: name ~ r'.*RUTHERFORD.*'
+            aggregate: f is string_agg(name)
+          }`
+        )
+        .run();
+      expect([
+        'RUTHERFORD PAT R JR,RUTHERFORD JAMES C',
+        'RUTHERFORD JAMES C,RUTHERFORD PAT R JR',
+      ]).toContain(result.data.path(0, 'f').string.value);
```
mtoy-googly-moogly, line 1138: I really want this to be

```
await expect(`run: aircraft -> {
            where: name ~ r'.*RUTHERFORD.*'
            aggregate: f is string_agg(name)
          }`).malloyResultMatches(expressionModel, {f: [
        'RUTHERFORD PAT R JR,RUTHERFORD JAMES C',
        'RUTHERFORD JAMES C,RUTHERFORD PAT R JR',
      ]);
 ```
 
 Is there a reason it doesn't do that?

PR #1576, opened 2024-01-10
feat: renderer can produce result

skokenes:
Enable the renderer to optionally receive `queryResult` and `modelDef` and reconstruct the Result object internally, as an override to providing an already constructed Result object. This makes it easier to use the web component in apps like notebooks that are receiving the serialized queryResult and modelDef but don't have malloy code for constructing the Result.


whscullin (approved):
LGTM.

PR #1578, opened 2024-01-10
WIP: SQL expression functions

christopherswenson:
TODO:
- [x] Add experiment `sql_functions` for this feature
- [x] Log an error if the user tries to interpolate a field from a join (for now)
- [x] Log an error if the user tries to interpolate a join itself (for now)

(The join stuff is going to be disabled for now because it doesn't fully work with the compiler's join finder yet. We were thinking first step would be to disable it with an error, second step would be to fix it and re-enable.) 

Need to think about, for the future:
- Whether we want a real syntax for this rather than functions, e.g. `sql'MY SQL EXPRESSION'!number`
- Whether a user can specify an aggregate or analytic expression

### DIFF
--- a/packages/malloy/src/lang/ast/expressions/expr-func.ts
+++ b/packages/malloy/src/lang/ast/expressions/expr-func.ts
@@ -207,7 +207,7 @@ export class ExprFunc extends ExpressionDef {
       );
       return errorFor('cannot call with source');
     }
-    const funcCall: Expr = [
+    let funcCall: Expr = [
       {
         type: 'function_call',
         overload,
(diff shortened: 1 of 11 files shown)


packages/malloy/src/lang/ast/expressions/expr-func.ts
```diff
@@ -226,6 +226,12 @@ export class ExprFunc extends ExpressionDef {
         'sql_boolean',
       ].includes(func.name)
     ) {
+      if (!this.inExperiment('sql_functions')) {
```
amarkowitz, line 229: @christopherswenson - what do you think?

christopherswenson: Yep, I like adding this as an experiment. Though you should pass in `true` as a second parameter to `inExperiment` — otherwise it logs an error, as you can see in your test (it logs a generic error in addition to the custom one you're logging).

packages/malloy/src/lang/ast/expressions/expr-func.ts
```diff
@@ -369,3 +435,34 @@ function findOverload(
     }
   }
 }
+
+type InterpolationPart =
+  | {type: 'string'; value: string}
+  | {type: 'interpolation'; name: string};
+
+function parseSQLInterpolation(template: string): InterpolationPart[] {
```
christopherswenson, line 443: It doesn't look like either of us added a way to include `${` literally in the SQL string. Not sure how important that is at this stage, but it might be worth doing a `.replace('\$', '$')` or something in the `{type: 'string'}` case to allow for that.

christopherswenson (commented):
Looks good @amarkowitz, thanks for taking this over! Just the two comments I left to address (the one on this review about escaping `${` and the one about the experiment error message).

christopherswenson:
Also looks like some tests are failing... looking.

test/src/databases/all/expr.spec.ts
```diff
@@ -340,6 +340,170 @@ describe.each(runtimes.runtimeList)('%s', (databaseName, runtime) => {
...
+    });
+
+    it('sql_boolean', async () => {
+      await expect(`
+      ##! experimental { sql_functions }
+      source: a is ${databaseName}.table('malloytest.aircraft_models') extend { where: aircraft_model_code ? '0270202' }
+
+      run: a -> {
+          group_by: boolean_1 is sql_boolean("\${seats} > 20")
+          group_by: boolean_2 is sql_boolean("\${engines} == 2")
```
christopherswenson, line 379: Looks like this should just be `sql_boolean("\${engines} = 2")` with one `=` rather than two.

amarkowitz: Ah, interesting.  I was using duckdb for local testing and this seems to pass 🤷   Clearly fails for postgres and bq so I'll be sure to run those locally too on future pulls to double check.

packages/malloy/src/lang/ast/expressions/expr-func.ts
amarkowitz: Great call out.  I'll take a look at this today and update.  Thanks!

christopherswenson: Per discussion with you in chat, since the Looker parser doesn't do this, no need to do it for now.

amarkowitz (approved):
LGTM

christopherswenson:
LGTM, thanks for bringing this up to date and across the finish like @amarkowitz

PR #1586, opened 2024-01-18
Update render doc

srschandan:


packages/malloy-render/README.md
```diff
@@ -21,3 +21,22 @@ Stories are written in the `src/stories` directory. To add more data and Malloy
 ### On Reloading Changes
 
 When running `npm run storybook`, only changes in the malloy-render package will hot reload properly. Changes to dependencies like the core `malloy` package may require a browser reload to work properly.
+
+## Using the Bundled Renderer
+
+Any JS platform with JSON results from a DB query execution and a PreparedResult object from Malloy library (or from Malloy Service) should be able to use the bundled renderer to create an HTML Malloy Result. The renderer bundle is avialable from v0.0.118 onwards. Example usage:
```
skokenes, line 27:
```suggestion
Any web browser with JSON results from a DB query execution and a PreparedResult object from the Malloy library should be able to use the bundled renderer to create an HTML Malloy Result. The renderer bundle is available from v0.0.118 onwards. Example usage:
```
Small typo, removing reference to internal only stuff...

packages/malloy-render/README.md
```diff
@@ -21,3 +21,22 @@ Stories are written in the `src/stories` directory. To add more data and Malloy
...
+  'https://cdn.jsdelivr.net/npm/@malloydata/render@0.0.118/dist/bundle/bundled_renderer.min.js';
+document.head.appendChild(script);
+var resultElement = document.getElementById('result_div');
+renderMalloyResults(result, total_rows, preparedResult).then(
+  function (malloyResElement) {
+    resultElement.appendChild(malloyResElement);
+  }
+);
+```
+
```
skokenes, line 41: We might also want to include an example of how to just use the new malloy renderer as a web component? That can come later though

skokenes (approved):
LGTM

packages/malloy-render/README.md
srschandan: Ack.

PR #1588, opened 2024-01-23
max can be null

whscullin:
...when computing column width. Fixes
Screenshot 2024-01-22 at 5 03 54 PM

### DIFF
--- a/packages/malloy-render/src/component/table-layout.ts
+++ b/packages/malloy-render/src/component/table-layout.ts
@@ -83,8 +83,8 @@ function getColumnWidth(f: Field, metadata: RenderResultMetadata) {
       width =
         Math.max(getTextWidth(fieldMeta.maxString!, font), titleWidth) +
         COLUMN_BUFFER;
-    } else if (f.isAtomicField() && f.isNumber()) {
-      const formattedValue = renderNumericField(f, fieldMeta.max!);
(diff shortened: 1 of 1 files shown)


packages/malloy-render/src/component/table-layout.ts
```diff
@@ -83,8 +83,8 @@ function getColumnWidth(f: Field, metadata: RenderResultMetadata) {
       width =
         Math.max(getTextWidth(fieldMeta.maxString!, font), titleWidth) +
         COLUMN_BUFFER;
-    } else if (f.isAtomicField() && f.isNumber()) {
-      const formattedValue = renderNumericField(f, fieldMeta.max!);
+    } else if (f.isAtomicField() && f.isNumber() && fieldMeta.max !== null) {
+      const formattedValue = renderNumericField(f, fieldMeta.max);
```
skokenes, line 87: We probably still want to measure the null value display here? Which is '∅', and I guess should be exported as a const somewhere...
 
For now, could be this?
```suggestion
    } else if (f.isAtomicField() && f.isNumber()) {
      const formattedValue = fieldMeta.max === null ? '∅' : renderNumericField(f, fieldMeta.max);
    }
```
or should renderNumericField be changed to allow a number | null as input and handle the null display there?

skokenes (approved):
LGTM

Issue #1605, opened 2024-02-03
Scalar Lenses: order by field not available in the output space


lloydtabb:
> Unknown field in the output space

```
run: duckdb.table('data/state_facts.parquet') extend {
  measure: c is count()
}
-> popular_name + c + {
  nest: state + c + {order_by: state}
}
```

PR #1607, opened 2024-02-05
Fix bug where `nest: a + b + { order_by: a }` could not look up `a`

christopherswenson:
Fixes https://github.com/malloydata/malloy/issues/1605

Bug: `nest: a + b + { order_by: a }` failed to compile with an error that `a` could not be found in the output space (when evaluating the `order_by`.

This was actually an unfortunate parsing error. The rule for nests was:
```g4
nestEntry
  : tags fieldPath (PLUS vExpr)?    # nestExisting
  | tags queryName isDefine vExpr   # nestDef
  ;
```
But this meant that `nest: a + b + { order_by: a }` would be parsed the same as `nest: a + (b + { order_by: a })`, which should make it clear why "`a` is not in the output space."

My solution is to change the rule for nests to:
```g4
nestEntry
  : tags (queryName isDefine)? vExpr   # nestDef
  ;
```
This makes the `name is` part syntactically optional. Then, in the AST builder for the `nestDef`, we traverse the `View` tree to see if there's a valid inferable name (refinement `X + Y` asks `X` for its name, field reference `a` has a name, arrow `X -> Y` asks `Y` for its name, and query properties `{ ... }` has no name). If there is no inferable name, we yield an error telling you to add a name.

This has a nice side effect that now if you write `nest: { group_by: foo }` you no longer get a nasty parse error, but instead get a nice error telling you that you need a name.

### DIFF
--- a/packages/malloy/src/lang/ast/view-elements/qop-desc-view.ts
+++ b/packages/malloy/src/lang/ast/view-elements/qop-desc-view.ts
@@ -120,4 +120,8 @@ export class QOpDescView extends View {
     }
     return pipeline;
   }
+
+  getImplicitName(): string | undefined {
+    return undefined;
+  }
 }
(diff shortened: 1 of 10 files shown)


mtoy-googly-moogly (approved):
Excellent

PR #1618, opened 2024-02-07
WIP allow unknown function calls to specify an expression type

christopherswenson:


christopherswenson:
Not sure about `measure`/`calculation` vs `aggregate`/`analytic`...

PR #1621, opened 2024-02-08
CSV writer for nested data

srschandan:
Creates  CSV for Malloy Results where nested data is converted to CSV as well.

### DIFF
--- a/packages/malloy/src/malloy.ts
+++ b/packages/malloy/src/malloy.ts
@@ -66,6 +66,7 @@ import {
   ImportLocation,
   Annotation,
   NamedModelObject,
+  QueryValue,
 } from './model';
 import {
   Connection,
@@ -3857,6 +3858,26 @@ export class JSONWriter extends DataWriter {
   }
 }
 
+type CellMatrix = {
+  rows: string[];
+  length: number;
+  width: number;
+};
+
+/**
+ * CSV writer class that handles nested data.
+ * This writer creates CSV using a DFS traversal of the result dataset.
+ * Each trivial column value is converted to a CSV of 1x1 matrix and all the
+ * columns are merged together to create a CSV that represents 1 QueryDataRow.
+ * Since this follows DFS, each non trivial data is rendered into a NxM matrix
+ * where N is the number of rows in the nested data and M is the number of
+ * columns it has.
+ * For any row with X number of columns, we end up with X number of NxM matrices
+ * where the value of N,M pair may be different for each column.
+ * We then merge the matrices so that we end up with a larger matrix of size
+ * Max(N)xSum(M) by taking one row of csv from each matric at a time. For any
+ * matrix with N<Max(N), we add a row of empty CSV cells of size N.
+ */
 export class CSVWriter extends DataWriter {
   private readonly columnSeparator = ',';
   private readonly rowSeparator = '\n';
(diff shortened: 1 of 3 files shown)


packages/malloy/src/malloy.ts
```diff
@@ -3882,49 +3904,152 @@ export class CSVWriter extends DataWriter {
...
+  private mergeMatrices(matrices: CellMatrix[]): CellMatrix {
+    const maxLength = Math.max(...matrices.map(matrix => matrix.length));
+    const matrixWidth = matrices.reduce((sum, matrix) => sum + matrix.width, 0);
+    const csvMatrix: string[] = [];
+    for (let i = 0; i < maxLength; i++) {
+      let csvRow = '';
+      for (const matrix of matrices) {
+        if (i < matrix.length) {
+          csvRow = csvRow + matrix.rows[i];
+          if (!csvRow.endsWith(this.columnSeparator)) {
```
whscullin, line 3967: Is the goal here to not have two separators in a row? That seems like it could get confused by empty cells. In general, the more efficient and predictable way to separate a series of cells is to create an array and then do `.join()`. That prevents trailing separators like the ones that are failing tests now.

If you can compute the length of the array in advance you can save some memory allocations, but even if you can't `.push()` will overall be more performant and memory efficient.

packages/malloy/src/malloy.ts
```diff
@@ -3882,49 +3904,152 @@ export class CSVWriter extends DataWriter {
...
+            csvRow = csvRow + this.columnSeparator;
+          }
+        } else {
+          // Add empty cells.
+          for (let w = 0; w < matrix.width; w++) {
+            csvRow = csvRow + this.columnSeparator;
```
whscullin, line 3973: This could be `this.columnSeparator.repeat(matrix.width)` but I would recommend looking into the `.join()` strategy

packages/malloy/src/malloy.ts
srschandan: Done

packages/malloy/src/malloy.ts
srschandan: Done.

packages/malloy/src/malloy.ts
```diff
@@ -3882,49 +3905,148 @@ export class CSVWriter extends DataWriter {
...
-    } else if (cell.isDate() || cell.isTimestamp()) {
-      return cell.value.toISOString();
-    } else if (cell.isString()) {
-      return cell.value;
+  // Re-using the old stringify method for sanity.
+  private stringify(value: QueryValue) {
+    if (value === null) {
+      return this.emptyCell;
+    } else if (value instanceof Date) {
+      return value.toISOString();
```
whscullin, line 3913: Is it OK that this behavior has changed from converting dates and timestamps to ISO strings to just dates?

whscullin (commented):
One question, and I would love some tests, otherwise LGTM.

packages/malloy/src/malloy.ts
srschandan: This is the tradeoff if we want to be able to support pivoted tables in the CSVWriter. When we have pivoted data, the re-constructed result object (TODO) would not have structdef/modeldef. Hence we have to inject the data as a raw JSON object in the writer, where we lose the ability to identify timestamps 
```
export type QueryScalar = string | boolean | number | Date | Buffer | null;
```

Unless we want to have a separate CSV writer for pivoted case, which adds to the complexity in VSCode where we have to constantly look for pivoted data, this is inevitable.

srschandan: @carlineng

PR #1634, opened 2024-02-29
Pavanka snowflake malloy

christopherswenson:


lloydtabb:
Snowflake has this weirdness where they uppercase all unquoted identifiers.  Our renderer won't work with this strategy.   The first attempt lowercase all variables and set a flag, but I don't want all snowflake users to have to set that flag (and it introduced other problems).   

Currently all identifiers are quoted and this works.  the downside is that many tables in Snowflake will have UPPERCASE columns and have to be written that way in Malloy.  The upside is that it will be easy to understand.

lloydtabb:
@christopherswenson in snowflake window functions in a GROUP BY query require an ORDER BY in the OVER () clause.  Bummer.

PR #1645, opened 2024-03-07
Fix support for running individual dbs

whscullin:
This introduces `onlyIf()`, a utility function to allow a test to be skipped by conditional, while still allowing test runners to pick up the test signature. 

It also hardens `describeIfDatabaseAvailable()` so that the `describe()` returned can handle `describe.each()`s with empty input arrays.

Database tests in packages don't have access to the utils script so a mini skip header is used for those.

### DIFF
--- a/packages/malloy-db-bigquery/src/bigquery.spec.ts
+++ b/packages/malloy-db-bigquery/src/bigquery.spec.ts
@@ -31,7 +31,7 @@ import {fileURLToPath} from 'url';
 const envDatabases = (
   process.env['MALLOY_DATABASES'] ||
   process.env['MALLOY_DATABASE'] ||
-  ''
+  'bigquery'
 ).split(',');
 
 let describe = globalThis.describe;
--- a/packages/malloy-db-bigquery/src/bigquery.spec.ts
+++ b/packages/malloy-db-bigquery/src/bigquery.spec.ts
@@ -28,6 +28,17 @@ import * as util from 'util';
 import * as fs from 'fs';
 import {fileURLToPath} from 'url';
 
+const envDatabases = (
+  process.env['MALLOY_DATABASES'] ||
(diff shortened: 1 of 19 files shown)


packages/malloy-db-bigquery/src/bigquery.spec.ts
```diff
@@ -28,6 +28,17 @@ import * as util from 'util';
 import * as fs from 'fs';
 import {fileURLToPath} from 'url';
 
+const envDatabases = (
```
mtoy-googly-moogly, line 31: Since this gesture is so common maybe it could be

```Typescript
envDatabases = getEnvDatabases(default | undefined)
```

?

packages/malloy-db-bigquery/src/bigquery.spec.ts
```diff
@@ -28,6 +28,17 @@ import * as util from 'util';
...
+  process.env['MALLOY_DATABASES'] ||
+  process.env['MALLOY_DATABASE'] ||
+  'bigquery'
+).split(',');
+
+let describe = globalThis.describe;
```
mtoy-googly-moogly, line 37: may this as a utility function too?

mtoy-googly-moogly:
awesome thanks for doing this, left a couple of ignorable comments!

packages/malloy-db-bigquery/src/bigquery.spec.ts
whscullin: There's a bunch of repeated code in our connections that might benefit from some sort of core utilities library. Packages aren't suppose to reach out and pull in code from other directories without a package dependency (although we may not care that much, I guess), but the only package they share is @malloydata/malloy, and I feel like adding test helpers there might just be bloat.

PR #1656, opened 2024-03-13
export dist/dialect in malloy

pavanka:
- export `dist/dialect` as "dialect" in malloy.
this enables unblocks importing from malloy service
```
>> malloy-service/src/server.ts
import {Dialect, registerDialect} from '@malloydata/malloy/dialect'
```
- also fix the snowflake_tables script {just for posterity} of how to infer types for parquet format


pavanka:
not needed

PR #1657, opened 2024-03-13
cleanup sql: and from_sql()

mtoy-googly-moogly:
Some cleanup that should have happened when 4.0 finished was missed. Doing it now because other changes were touching this code, better to delete it than to keep it compiling for no reason.


christopherswenson (approved):
This all looks good to me. The bit about removing the `sql` keyword from the grammar will make it no longer highlight this way:
Screenshot 2024-03-18 at 12 17 14 PM

But instead like `table` here:
Screenshot 2024-03-18 at 12 17 20 PM

And it is probably preferable for them to be the same. Not sure which is more helpful highlighting, but a win for consistency, and I think the "function"-like highlighting is better than the "keyword"-like highlighting in this case. 👍

PR #1667, opened 2024-03-18
Figutierrez/trino2

narreola:


narreola:
This pr provides a basic version of trino connection/dialect.

PR #1669, opened 2024-03-19
Add initial trino connection to iterate on.

narreola:


packages/malloy-db-trino/src/trino_connection.ts
```diff
@@ -0,0 +1,467 @@
...
+  }*/
+
+  public async estimateQueryCost(_sqlCommand: string): Promise<QueryRunStats> {
+    /*const dryRunResults = await this.dryRunSQLQuery(sqlCommand);
+    return {
+      queryCostBytes: Number(
+        dryRunResults.metadata.statistics.totalBytesProcessed
+      ),
+    };*/
+    throw new Error('Not impld 6');
```
whscullin, line 451: I would stub this, otherwise VS Code will be unhappy.

packages/malloy-db-trino/src/trino_executor.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (server) {
+      const user = process.env['TRINO_USER'];
+
+      if (!user) {
+        throw Error(
+          'Trino server specified but no user was provided. Set TRINO_USER and TRINO_PASSWORD environment variables'
+        );
+      }
+
+      const password = process.env['TRINO_PASSWORD'];
```
whscullin, line 40: I would still have a way to pass in a password so we can store that in the VS Code secret mananger.

whscullin (approved):
Couple minor things to work better in VS Code, otherwise 🚢 🇮🇹

packages/malloy-db-trino/src/trino_executor.ts
narreola: will fix a bit later.

PR #1687, opened 2024-03-27
feat: basic plot and bar

skokenes:
Implements basics of plot API and migrates our existing basic bar chart to it


whscullin (approved):
LGTM

PR #1704, opened 2024-04-03
Trino: wasn't naming doubly nested struts.

lloydtabb:


packages/malloy/src/dialect/trino/trino.ts
```diff
@@ -138,7 +138,7 @@ export class TrinoDialect extends Dialect {
         if (f.isArray) {
           s = `array(${s})`;
         }
-        fields.push(s);
```
lloydtabb, line 141: @whscullin this

PR #1717, opened 2024-04-10
Fix double nested schema parsing

whscullin:
Handle the case of arrays in arrays - `aa array(row(a double, b array(row(c integer, d varchar(60)))))` was getting parsed with a name of `aa array(row(a double, b ` and type of `array(row(c integer, d varchar(60)))))`, this makes sure we stop at the first `array(`


whscullin:
Handle the case of arrays in arrays - aa array(row(a double, b array(row(c integer, d varchar(60))))) was getting parsed with a name of aa array(row(a double, b  and type of array(row(c integer, d varchar(60))))), this makes sure we stop at the first array(

Also fixes timestamps being returned as strings.

PR #1718, opened 2024-04-10
Trino: Nest pipelines

lloydtabb:


lloydtabb:
make nested pipelines work.

PR #1721, opened 2024-04-15
[snowflake] add support for safe cast in snowflake dialect

pavanka:
- ensure output col names from snowflake are of same case as the malloy identifiers
- snowflake connection and executor tests succeed now
- add support for safe cast
- ensure safe cast expr test passes for snowflake

### DIFF
--- a/packages/malloy-db-snowflake/src/snowflake_executor.ts
+++ b/packages/malloy-db-snowflake/src/snowflake_executor.ts
@@ -170,10 +170,12 @@ export class SnowflakeExecutor {
 
   private async _setSessionParams(conn: Connection) {
     // set some default session parameters
+    // ensure we do not ignore case for quoted identifiers
+    await this._execute("ALTER SESSION SET QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE;", conn);
     // set utc as the default timezone which is the malloy convention
     await this._execute("ALTER SESSION SET TIMEZONE = 'UTC';", conn);
     // ensure week starts on Sunday which is the malloy convention
-    await this._execute('ALTER SESSION SET WEEK_START = 7;', conn);
+    await this._execute("ALTER SESSION SET WEEK_START = 7;", conn);
     // so javascript can parse the dates
     await this._execute(
       "ALTER SESSION SET TIMESTAMP_NTZ_OUTPUT_FORMAT='YYYY-MM-DDTHH24:MI:SS.FF3TZH:TZM';",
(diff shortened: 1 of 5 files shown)


packages/malloy-db-snowflake/src/snowflake_executor.ts
```diff
@@ -170,10 +170,12 @@ export class SnowflakeExecutor {
 
   private async _setSessionParams(conn: Connection) {
     // set some default session parameters
+    // ensure we do not ignore case for quoted identifiers
```
lloydtabb, line 173: Can you explain why you added this back?  Have you experimented with this?

pavanka: hey @lloydtabb yes I changed this to essentially get tests to pass, commented underneatb

test/src/databases/all/expr.spec.ts
```diff
@@ -388,12 +388,12 @@ describe.each(runtimes.runtimeList)('%s', (databaseName, runtime) => {
...
   });
 
-  test.when(!['postgres', 'snowflake'].includes(runtime.connection.name))(
+  test.when(!['postgres'].includes(runtime.connection.name))(
     'sql safe cast',
     async () => {
       await expect(`
       run: ${databaseName}.sql('SELECT 1 as one') -> { select:
-        bad_date is '123':::date
+        bad_date is '12a':::date
```
pavanka, line 396: originally in executor we set QUOTED_IDENTIFIERS_IGNORE_CASE =  TRUE, so we always ignore case even if quoted, so whether we accessed it as flight."id" or flight.id, we were able to read from the id/ID field.

here I am just asserting that QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE, which is the behavior we finalized for malloy.

what I found is that if I do not set this to FALSE explicitly, for some reason output would look like
BAD_DATE, BAD_NUMBER, GOOD_NUMBER, though in generated SQL we have "bad_date", "bad_number", "good_number", etc.

I also ran tests under databases/all and all succeeded for me:

```
 MALLOY_DATABASE=snowflake npx jest --runInBand test/src/databases/all
 
Test Suites: 10 passed, 10 total
Tests:       15 skipped, 354 passed, 369 total
Snapshots:   0 total
Time:        397.491 s
Ran all test suites matching /test\/src\/databases\/all/i.
 ```

pavanka (commented):
thank you for the review!

pavanka:
synced with the latest malloy and updated pr

PR #1754, opened 2024-07-02
presto docker

skokenes:


lloydtabb:
Trino is busted and I want to start building the presto dialect so I'm merging

PR #1771, opened 2024-07-12
feat: duration formatting

skokenes:
## Changes
Adds:
- ability to format a duration when the `# number` tag is included
- ability to list durations in a terse format using `# duration.terse`

Fixes:
- if bad duration name is supplied, it is ignored and the default duration (seconds) is used

Cleanup:
- extracts duration functions into 1 place and exports them for re-use across both old and new renderer

## Examples
Old renderer
CleanShot 2024-07-12 at 11 07 13@2x

New renderer
CleanShot 2024-07-12 at 11 07 49@2x

### DIFF
--- a/packages/malloy-render/src/component/render-numeric-field.ts
+++ b/packages/malloy-render/src/component/render-numeric-field.ts
@@ -36,12 +36,13 @@ const multiplierMap = new Map<DurationUnit, number>([
   [DurationUnit.Days, Number.MAX_VALUE],
 ]);
 
-function formatTimeUnit(value: number, unit: DurationUnit) {
+function formatTimeUnit(value: number, unit: DurationUnit, numFormat?: string) {
   let unitString = unit.toString();
   if (value === 1) {
     unitString = unitString.substring(0, unitString.length - 1);
   }
-  return `${value} ${unitString}`;
+  const formattedValue = numFormat ? format(numFormat, value) : value;
+  return `${formattedValue} ${unitString}`;
 }
 
 export function renderNumericField(f: AtomicField, value: number) {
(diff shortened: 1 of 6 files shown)


packages/malloy-render/src/component/render-numeric-field.ts
```diff
@@ -24,6 +24,7 @@
 import {AtomicField} from '@malloydata/malloy';
 import {Currency, DurationUnit} from '../data_styles';
 import {format} from 'ssf';
+import { getText } from '../html/duration';
```
nachoarreola, line 27: nit: run linter?

packages/malloy-render/src/html/duration.ts
```diff
@@ -21,16 +21,95 @@
...
 } from '../data_styles';
 import {RendererOptions} from '../renderer_types';
 import {Renderer} from '../renderer';
 import {RendererFactory} from '../renderer_factory';
+import { format } from 'ssf';
+
+export function formatTimeUnit(value: number, unit: DurationUnit, options: { numFormat?: string; terse?: boolean } = {}) {
+  let unitString = unit.toString();
+    if(options.terse) unitString = terseDurationUnitsMap.get(unit) ?? unitString;
+    else if (value === 1) {
```
nachoarreola, line 40: nit: can we add parenthesis here? I feel it would be more readable.

nachoarreola (approved):

packages/malloy-render/src/component/render-numeric-field.ts
skokenes: DOne

packages/malloy-render/src/html/duration.ts
skokenes: done

Issue #1764, opened 2024-07-11
When using refinements, unsupported type appears in JSON but not in HTML


bporterfield:
### What happens?

Some fields that are of unsupported type (uuid in postgres, in this case) will show up as a column in a query, as expected. However, when using refinements and adding an existing view, the fields appear in the JSON response but do not appear in the HTML.

```
    view: chat_detail is {
        group_by: 
            `chat id` is id,
            model_message_text, 
            chat_date, 
            `user name` is user.name,
            user.email
    }
```

Running the view above shows "chat id" as a column. However, running the query below does **not** show "chat id" as a column:

```
run: chat -> { 
  where: 
    user.name = 'Ben',
} + chat_detail
```

When running this, `chat id` is not shown in HTML, but it does appear in the JSON response:
<img width="742" alt="Screenshot 2024-07-11 at 9 13 59 AM" src="https://github.com/malloydata/malloy/assets/108260/3813fd4f-6e07-4210-9330-5882d7216cbc">

<img width="111" alt="Screenshot 2024-07-11 at 9 13 35 AM" src="https://github.com/malloydata/malloy/assets/108260/1db2f1ec-f70e-4663-aaff-0e7e7c9083dd">

However, if I change `chat id` definition in the view to:
```
`chat id` is id::string,
```
Now `chat id` appears in the HTML response as well as JSON

### To Reproduce

see above

### OS:

macOS

### Malloy Client:

local VSCode

### Malloy Client Version:

latest

### Database Connection:

Postgres

Issue #1761, opened 2024-07-11
Support uuid type in Postgres


bporterfield:
### What happens?

It's not uncommon to use uuid for primary keys, should be simple to support but not supporting means some annoying casting and confusing error messages

### To Reproduce

create table with uuid, note that it is not a supported type in malloy, encounter various error messages

### OS:

any

### Malloy Client:

any

### Malloy Client Version:

any

### Database Connection:

Postgres

PR #1790, opened 2024-07-22
fix: add uuid to duckdb, postgres

skokenes:
Adds uuid support to DuckDB and Postgres. I believe this fixes https://github.com/malloydata/malloy/issues/1764 and https://github.com/malloydata/malloy/issues/1761, but need some manual verification still

### DIFF
--- a/packages/malloy/src/dialect/duckdb/duckdb.ts
+++ b/packages/malloy/src/dialect/duckdb/duckdb.ts
@@ -67,6 +67,7 @@ const duckDBToMalloyTypes: {[key: string]: FieldAtomicTypeDef} = {
   'TIME': {type: 'string'},
   'DECIMAL': {type: 'number', numberType: 'float'},
   'BOOLEAN': {type: 'boolean'},
+  'UUID': {type: 'string'},
 };
 
 export class DuckDBDialect extends Dialect {
--- a/packages/malloy/src/dialect/postgres/postgres.ts
+++ b/packages/malloy/src/dialect/postgres/postgres.ts
@@ -89,6 +89,7 @@ const postgresToMalloyTypes: {[key: string]: FieldAtomicTypeDef} = {
   'bytea': {type: 'string'},
   'pg_ndistinct': {type: 'number', numberType: 'integer'},
   'varchar': {type: 'string'},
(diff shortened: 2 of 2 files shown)


mtoy-googly-moogly (changes requested):
I do not approve this change. We used to do this, I backed it out when I added unsupported (now called unsupported native sql) types.

This is not "supporting" uuid. This is pretending that a uuid is a string.

I am open to argument about how wrong I am, but I need the argument.

leandro-hermes:
I don't think (in my poor knowledge) that we need some kind of support for uuid in Malloy - I don't see any use for this.

But on the other hand, I see benefits in considering them as string fields, because currently Malloy doesn't allow us to create expressions with unsupported types:

```
line 4: Unsupported type not allowed in expression
  |   where: id = '6ffae411-6852-4131-9ba9-e022838991b7'
  |          ^ 
  ```

The need to cast the field as string causes the database to not properly use the indexes created on these fields, degrading the performance.

skokenes:
> I do not approve this change. We used to do this, I backed it out when I added unsupported (now called unsupported native sql) types.
> 
> This is not "supporting" uuid. This is pretending that a uuid is a string.
> 
> I am open to argument about how wrong I am, but I need the argument.

Ok. This came across my desk because the legacy renderer isn't showing uuid columns at all, they just disappear in the output tables. I'll look into why. The "Unsupported" renderer is supposed to render the unsupported data type as a string if it can. And it would seem like we should be able to do that with a uuid column, so I'm not sure whats going on there.

mtoy-googly-moogly:
@leandro-hermes this is a great example, the idea of unsupported native types is newish and we definitely need to listen to people trying to use them and work on them a little

the workaround for that might be to write, (i think)

    where: id = '6ffae411-6852-4131-9ba9-e022838991b7'::"uuid"
    
 how does that feel?

leandro-hermes:
@mtoy-googly-moogly I don't like the idea of casting the string, since we don't need to do this in SQL (at least in PostgreSQL and DuckDB). Since Malloy is supposed to simplify SQL, by adding this, we would hurt this idea.

But, it is not a thing that would make me stop using Malloy at all.

PR #1820, opened 2024-08-09
Prettier errors when opening these files

mtoy-googly-moogly:
Apparently it is possible to check in files that prettier doesn't like. These are the ones I found by clicking all the files in packages/malloy/lang


mtoy-googly-moogly:
Closed --- @whscullin is doing a better and more global fix.

PR #1824, opened 2024-08-13
fix nullsafe not

lloydtabb:


test/src/databases/all/nomodel.spec.ts
```diff
@@ -63,6 +63,18 @@ afterAll(async () => {
 
 runtimes.runtimeMap.forEach((runtime, databaseName) => {
   const q = runtime.getQuoter();
+  // Issue #1824
+  it(`not boolean field with null - ${databaseName}`, async () => {
+    await expect(`
+      run: ${databaseName}.sql("""
+          SELECT
+            CASE WHEN 1=1 THEN NULL ELSE false END as ${q`n`}
```
daniel-ohayon, line 71: you can't just do `SELECT NULL AS ...` ? (doesn't work with all dialects i assume?)

lloydtabb: That would fail in bigquery.   You can't select a NULL literal.

lloydtabb: The case statement forces it's type to be a boolean.

PR #1828, opened 2024-08-13
Move NOT in nullsafe not

mtoy-googly-moogly:
change `NOT X`

from `COALESCE(NOT X, FALSE)` to `NOT COALESCE(X, FALSE)`

Truth table should be ...

| X | NOT X |
| ---- | ---- |
| true | false |
| false | true |
| null | true |

- [ ] I don't there there was ever a set of tests for this, needs to be some.


mtoy-googly-moogly:
Llloyd already didi this in #1824

PR #1846, opened 2024-08-19
Revert lodash version.

nachoarreola:


whscullin:
FWIW we're actually using 4.17.21 by virtue of using compatible versioning (^4.17.20)

PR #1880, opened 2024-09-04
Be more friendly with LIKE and IS NULL

mtoy-googly-moogly:
The expressions _expr_ `IS NULL`, _expr_ `IS NOT NULL`, _expr_ `LIKE` _expr_ , and _expr_ `NOT LIKE` _expr_ are all parsed and generated correct code, and return a compile time warning telling the user to write in Malloy-speak.

Also this changed how null safe not and comparisons are handled.


nachoarreola:
I'd love to experiment with code replacement suggestions in warnings, would it be easy to add them for this particular erorr? ```replacement: { textRange: DocumentRange, replaceWith: string }```

mtoy-googly-moogly:
@nachoarreola I think that is possible. I was imagining there was a separate linter process which made all kinds of replacement suggestions, including reformatting and code re-writing for old versions of Malloy, stuff that might not be doable by the compiler.

mtoy-googly-moogly:
Since an error already has a range, I just added `replacement?: string` to the `LogMessage`

PR #1882, opened 2024-09-06
Function Blueprint DSL

christopherswenson:
Improves the experience of authoring new Malloy standard functions, dialect overrides, and dialect functions by creating a function "blueprint" system. This system is a typescript DSL which can succinctly describe simple functions, e.g. 
```ts
{ takes: { value: 'number' }, returns: 'number', impl: { function: 'SIN' } }
```
and is also expressive enough to represent much more complicated functions, like `string_agg`. 

See the [README](https://github.com/malloydata/malloy/blob/f9a524a769c39561379c43f57a8ed93981d7d0ea/packages/malloy/src/dialect/functions/README.md) for details on the blueprint DSL.


packages/malloy/src/dialect/functions/malloy_standard_functions.ts
```diff
@@ -0,0 +1,778 @@
...
+    },
+    returns: {calculation: {generic: 'T'}},
+    impl: {function: 'LAG', needsWindowOrderBy: true},
+  },
+  'with_default': {
+    generic: ['T', LAG_TYPES],
+    takes: {
+      'value': {measure: {generic: 'T'}},
+      'offset': {literal: 'number'},
+      'default': {constant: {generic: 'T'}}, // TODO needs to be constant max aggregate?
```
christopherswenson, line 539: `LAG` theoretically supports `lag(value, null, count())` (a "constant aggregate" for the default value). This currently doesn't work, because aggregates are currently treated as "output space" expressions, but that conflicts with the "constant" requirement. I haven't worked out exactly what this implies about the Malloy type system. In the IR, this is expressible as `{ type: T, exprType: 'aggregate', evalSpace: 'constant' }`, but but type-checkable in `expr-func`. In the blueprint system, it is not expressible. Since this doesn't change current behavior, I feel okay making this change while I figure out what the right long term answer is, and if necessary, update the blueprint system to allow for it.

PR #1891, opened 2024-09-11
Add Some Trino Functions

christopherswenson:
Branched off https://github.com/malloydata/malloy/pull/1882


christopherswenson:
I haven't figured out yet what to do about these "inline" tests for aggregate or window functions. Scalar functions can just use constants for their arguments, but that doesn't really make sense for aggregate or window functions. Additionally, window functions need additional fields beyond just the one expression.

One way this could work:

```ts
// scalar test
{ 
  group_by: 'to_timestamp(1725555835) = @2024-09-05 17:03:55',
  result: true
}
// aka
['to_timestamp(1725555835) = @2024-09-05 17:03:55', true]

// aggregate test
{
  data: [{col: 1}, {col: 2}, {col: 3}],
  aggregate: `col.sum()`,
  result: 6
}

// window function test
{
  data: [{col: 1}, {col: 2}, {col: 3}],
  view: `{
    group_by: col
    order_by: col desc
    calculate: prev_col is lag(col)
  }`,
  result: [{col: 3, prev_col: null}, {col: 2, prev_col: 3}, {col: 1, prev_col: 2}],
}
```

For tests with `data`, we would generate an SQL source which created that data inline. The downside would be that each test would generate its own query against the database. To make this better, we could allow multiple aggregate tests with the same data:

```ts
{ 
  tests: [
    {
      data: [{col: 1}, {col: 2}, {col: 3}],
      cases: [
        {aggregate: `col.sum()`, result: 6},
        {aggregate: `col.sum(col + 1)`, result: 9},
      ]
    }
  ]
}
```

For really complex tests, we could allow multiple data sources and model definitions:

```ts
{ 
  tests: [
    {
      data: {
        table_one: [{id: 1, val: 6}, {id: 2, val: 7}, {id: 3, val: 2}],
        table_two: [{one_id: 1}, {one_id: 2}, {one_id: 3}, {one_id: 3}],
      },
      model: `
        source: one is table_one 
        source: two is table_two extend {
          join_one: one on one_id = one.id
        }
      `,
      source: 'two',
      cases: [
        {aggregate: `source.avg(one.val)`, result: 4.25},
        {aggregate: `one.val.avg()`, result: 5},
      ]
    }
  ]
}
```

Maybe this is overkill, and these complex tests should just be included in the test package...

PR #1902, opened 2024-09-13
Add some new Presto dialect functions

carlineng:
Adds the following functions, and associated tests:

- date_format
- from_unixtime
- regexp_replace
- to_unixtime
- approx_percentile
- date_parse
- bool_or
- corr
- bitwise_and
- bitwise_or
- min_by
- max_by
- json_extract_scalar
- regexp_like
- bool_and
- variance
- arbitrary
- bitwise_or_agg
- bitwise_and_agg
- bitwise_xor_agg

I was having trouble with local tests on `date_parse` and `to_unixtime`; I think something related to timezones and converting between Javascript datetime objects and Presto timestamps... any advice on how to resolve those would be appreciated.


packages/malloy/src/dialect/trino/dialect_functions.ts
```diff
@@ -62,9 +145,117 @@ const string_agg_distinct: OverloadedDefinitionBlueprint = {
...
+};
+
+const from_unixtime: DefinitionBlueprint = {
+  takes: {'unixtime': 'number'},
+  returns: 'timestamp',
+  impl: {function: 'FROM_UNIXTIME'},
+};
+
+const json_extract_scalar: DefinitionBlueprint = {
+  takes: {'json_val': 'string', 'json_path': 'string'},
```
christopherswenson, line 195: Seems like this can be `'json_val': ['string', 'json']`. Looking at the docs, it seems like the function supports both a string or JSON as the first argument (e.g. `JSON_EXTRACT_SCALAR(JSON '{"a": 1, "b": 2}', '$.a')`)

packages/malloy/src/dialect/trino/dialect_functions.ts
```diff
@@ -62,9 +145,117 @@ const string_agg_distinct: OverloadedDefinitionBlueprint = {
...
+  returns: 'string',
+  impl: {sql: 'JSON_EXTRACT_SCALAR(${json_val}, ${json_path})'},
+};
+
+const regexp_like: DefinitionBlueprint = {
+  takes: {'str': 'string', 'pattern': 'string'},
```
christopherswenson, line 201: Can probably be `'pattern': ['string', 'regular expression']` to support the Malloy regular expression syntax (`regexp_like("this contains foo in it", r'foo')`).

packages/malloy/src/dialect/trino/dialect_functions.ts
```diff
@@ -62,9 +145,117 @@ const string_agg_distinct: OverloadedDefinitionBlueprint = {
...
+  returns: 'boolean',
+  impl: {sql: 'REGEXP_LIKE(${str}, ${pattern})'},
+};
+
+const regexp_replace: OverloadedDefinitionBlueprint = {
+  remove_matches: {
+    takes: {'input_val': 'string', 'regexp_pattern': 'string'},
```
christopherswenson, line 208: Can also probably be `'pattern': ['string', 'regular expression']`

packages/malloy/src/dialect/trino/dialect_functions.ts
```diff
@@ -62,9 +145,117 @@ const string_agg_distinct: OverloadedDefinitionBlueprint = {
...
+    impl: {
+      sql: 'REGEXP_REPLACE(${input_val}, ${regexp_pattern})',
+    },
+  },
+
+  replace_matches: {
+    takes: {
+      'input_val': 'string',
+      'regexp_pattern': 'string',
+      'replace_pattern': 'string',
```
christopherswenson, line 219: I don't think this should be called `replace_pattern`, as that suggests it's a regular expression. In the presto docs, it's a string called `replacement`.

packages/malloy/src/dialect/trino/dialect_functions.ts
```diff
@@ -62,9 +145,117 @@ const string_agg_distinct: OverloadedDefinitionBlueprint = {
...
+    returns: 'string',
+    impl: {
+      sql: 'REGEXP_REPLACE(${input_val}, ${regexp_pattern})',
+    },
+  },
+
+  replace_matches: {
+    takes: {
+      'input_val': 'string',
+      'regexp_pattern': 'string',
```
christopherswenson, line 218: Also `'pattern': ['string', 'regular expression']`

test/src/databases/presto/presto.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+ *  LICENSE file in the root directory of this source tree.
+ */
+
+/* eslint-disable no-console */
+
+import {RuntimeList} from '../../runtimes';
+import {describeIfDatabaseAvailable} from '../../util';
+import '../../util/db-jest-matchers';
+
+const [describe] = describeIfDatabaseAvailable(['presto']);
```
christopherswenson, line 14: Since you're adding these functions to both Presto and Trino, these tests should run against both. Probably rename this file `presto-trino/presto-trino.spec.ts`, and change this line to  
```ts
const [describe, databases] = describeIfDatabaseAvailable(['presto', 'trino']);
```

and the beginning of the tests to

```ts
const runtimes = new RuntimeList(databases);
describe.each(runtimes.runtimeList)(
  'Presto/Trino dialect functions - %s',
  (databaseName, runtime) => {
```

test/src/databases/presto/presto.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+
+describe('Presto tests', () => {
+  const runtimeList = new RuntimeList(['presto']);
+  const runtime = runtimeList.runtimeMap.get('presto');
+  if (runtime === undefined) {
+    throw new Error("Couldn't build runtime");
+  }
```
christopherswenson, line 21: See above comment

test/src/databases/presto/presto.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+      }`).malloyResultMatches(runtime, {correlation: -0.9911108});
+  });
+
+  it('runs the json_extract_scalar function', async () => {
+    await expect(`run: presto.sql("""
+                SELECT 1 as n
+                """) -> {
+      select:
+        json_arr is json_extract_scalar('[1, 2, 3]', '$[2]')
+        json_obj is json_extract_scalar(
```
christopherswenson, line 177: If you allow the JSON type as mentioned above, you should probably test it here.

christopherswenson (approved):
This is super great work! I left some comments on some tweaks that are probably worth making.

carlineng:
Updated PR to address feedback, PTAL

carlineng:
Trino test is failing because the function signature for APPROX_PERCENTILE is different between Presto and Trino.

Trino is `approx_percentile(x, w, percentage)`, where "w" is a weight parameter that must be > 1
Presto is `approx_percentile(x, percentage, accuracy) `, where accuracy is a parameter between 0 and 1

carlineng:
MotherDuck tests are failing for an unrelated reason... looks like `count_approx` in the `all/functions.spec.ts` is not returning the expected result. I'm going to merge this anyway since I didn't alter those functions in this PR, so I think that failure is unrelated to the changes here.

PR #1909, opened 2024-09-17
WIP Remove `type: 'fieldref'`

christopherswenson:


christopherswenson:
Note to self that this preserves timeframe/numberType, etc.

PR #1939, opened 2024-09-27
Move `replaceMaterializedReferences` outside of `PreparedQuery`

christopherswenson:
The refactor I was trying to describe is really hard to explain so I just did it...

The philosophy: the runtime API has two kinds of things, "Malloy entities" and "Malloy entity materializers" (e.g. `Model` vs `ModelMaterializer`). The Malloy entities contain nothing more than the result of compiling that entity. The materializers contain runtime information needed to re-compile, run, etc.. Therefore, it didn't fit the philosophy for a `PreparedQuery` (which just represents the internal `Query` IR object) have a `replaceMaterializedReferences` property, since this is a description of _how_ you want to compile the SQL, not _what_ the query is.

I will add comments to all the code changes to describe in more detail the philosophy.

### DIFF
--- a/packages/malloy/src/malloy.ts
+++ b/packages/malloy/src/malloy.ts
@@ -102,9 +102,9 @@ export interface CompileOptions {
   noThrowOnError?: boolean;
 }
 
-type PrepareQueryOptions = {
+interface PrepareResultOptions {
   replaceMaterializedReferences?: boolean;
-};
+}
 
 export class Malloy {
   // TODO load from file built during release
@@ -223,12 +223,14 @@ export class Malloy {
     model,
     refreshSchemaCache,
     noThrowOnError,
+    replaceMaterializedReferences,
   }: {
     urlReader: URLReader;
     connections: LookupConnection<InfoConnection>;
     parse: Parse;
     model?: Model;
-  } & CompileOptions): Promise<Model> {
+  } & CompileOptions &
+    PrepareResultOptions): Promise<Model> {
     let refreshTimestamp: number | undefined;
     if (refreshSchemaCache) {
       refreshTimestamp =
@@ -350,7 +352,8 @@ export class Malloy {
             const conn = await connections.lookupConnection(connectionName);
             const expanded = Malloy.compileSQLBlock(
               result.partialModel,
-              toCompile
+              toCompile,
+              {replaceMaterializedReferences}
             );
             const resolved = await conn.fetchSchemaForSQLBlock(expanded, {
               refreshTimestamp,
@@ -380,7 +383,8 @@ export class Malloy {
 
   static compileSQLBlock(
     partialModel: ModelDef | undefined,
-    toCompile: SQLBlockSource
+    toCompile: SQLBlockSource,
+    options?: {replaceMaterializedReferences?: boolean}
   ): SQLBlock {
     let queryModel: QueryModel | undefined = undefined;
     let selectStr = '';
@@ -399,7 +403,11 @@ export class Malloy {
           }
           queryModel = new QueryModel(partialModel);
         }
-        const compiledSql = queryModel.compileQuery(segment, false).sql;
+        const compiledSql = queryModel.compileQuery(
+          segment,
+          false,
+          options?.replaceMaterializedReferences
+        ).sql;
         selectStr += parenAlready ? compiledSql : `(${compiledSql})`;
         parenAlready = false;
       }
@@ -726,19 +734,10 @@ export class Model implements Taggable {
    * @param queryName Name of the query to retrieve.
    * @return A prepared query.
    */
-  public getPreparedQueryByName(
-    queryName: string,
-    options?: PrepareQueryOptions
-  ): PreparedQuery {
+  public getPreparedQueryByName(queryName: string): PreparedQuery {
     const query = this.modelDef.contents[queryName];
     if (query?.type === 'query') {
-      return new PreparedQuery(
-        query,
-        this.modelDef,
-        this.problems,
-        options?.replaceMaterializedReferences ?? false,
-        queryName
-      );
+      return new PreparedQuery(query, this.modelDef, this.problems, queryName);
     }
 
     throw new Error('Given query name does not refer to a named query.');
@@ -750,10 +749,7 @@ export class Model implements Taggable {
    * @param index The index of the query to retrieve.
    * @return A prepared query.
    */
-  public getPreparedQueryByIndex(
-    index: number,
-    options?: PrepareQueryOptions
-  ): PreparedQuery {
+  public getPreparedQueryByIndex(index: number): PreparedQuery {
     if (index < 0) {
       throw new Error(`Invalid index ${index}.`);
     } else if (index >= this.queryList.length) {
@@ -762,8 +758,7 @@ export class Model implements Taggable {
     return new PreparedQuery(
       this.queryList[index],
       this.modelDef,
-      this.problems,
-      options?.replaceMaterializedReferences ?? false
+      this.problems
     );
   }
 
@@ -803,7 +798,7 @@ export class Model implements Taggable {
    * @return A prepared query.
    */
   public get preparedQuery(): PreparedQuery {
-    return this.getPreparedQuery({});
+    return this.getPreparedQuery();
   }
 
   /**
@@ -811,15 +806,14 @@ export class Model implements Taggable {
    *
    * @return A prepared query.
    */
-  public getPreparedQuery(options?: PrepareQueryOptions): PreparedQuery {
+  public getPreparedQuery(): PreparedQuery {
     if (this.queryList.length === 0) {
       throw new Error('Model has no queries.');
     }
     return new PreparedQuery(
       this.queryList[this.queryList.length - 1],
       this.modelDef,
-      this.problems,
-      options?.replaceMaterializedReferences ?? false
+      this.problems
     );
   }
 
@@ -885,7 +879,6 @@ export class PreparedQuery implements Taggable {
     query: InternalQuery,
     model: ModelDef,
     public problems: LogMessage[],
-    public replaceMaterializedReferences: boolean,
     public name?: string
   ) {
     this._query = query;
@@ -908,11 +901,15 @@ export class PreparedQuery implements Taggable {
    * @return A fully-prepared query (which contains the generated SQL).
    */
   public get preparedResult(): PreparedResult {
+    return this.getPreparedResult();
+  }
+
+  public getPreparedResult(options?: PrepareResultOptions): PreparedResult {
     const queryModel = new QueryModel(this._modelDef);
     const translatedQuery = queryModel.compileQuery(
       this._query,
       false,
-      this.replaceMaterializedReferences
+      options?.replaceMaterializedReferences ?? false
     );
     return new PreparedResult(
       {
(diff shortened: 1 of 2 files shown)


packages/malloy/src/malloy.ts
```diff
@@ -223,12 +223,14 @@ export class Malloy {
...
     noThrowOnError,
+    replaceMaterializedReferences,
   }: {
     urlReader: URLReader;
     connections: LookupConnection<InfoConnection>;
     parse: Parse;
     model?: Model;
-  } & CompileOptions): Promise<Model> {
+  } & CompileOptions &
+    PrepareResultOptions): Promise<Model> {
```
christopherswenson, line 233: When asking the Malloy object to compile something might want to pass in `{ replaceMaterializedReferences: true }`, because an SQL block may contain queries which should have their heads materialized.

packages/malloy/src/malloy.ts
```diff
@@ -380,7 +383,8 @@ export class Malloy {
 
   static compileSQLBlock(
     partialModel: ModelDef | undefined,
-    toCompile: SQLBlockSource
+    toCompile: SQLBlockSource,
+    options?: {replaceMaterializedReferences?: boolean}
```
christopherswenson, line 387: Same reason, a SQL block might contain queries that need to have stages materialized.

packages/malloy/src/malloy.ts
```diff
@@ -726,19 +734,10 @@ export class Model implements Taggable {
...
     const query = this.modelDef.contents[queryName];
     if (query?.type === 'query') {
-      return new PreparedQuery(
-        query,
-        this.modelDef,
-        this.problems,
-        options?.replaceMaterializedReferences ?? false,
-        queryName
-      );
+      return new PreparedQuery(query, this.modelDef, this.problems, queryName);
```
christopherswenson, line 740: Calling `getPreparedQuery` is "leaving the runtime world," and thus loses any runtime-specific settings, like `replaceMaterializedReferences`.

packages/malloy/src/malloy.ts
```diff
@@ -750,10 +749,7 @@ export class Model implements Taggable {
    * @param index The index of the query to retrieve.
    * @return A prepared query.
    */
-  public getPreparedQueryByIndex(
-    index: number,
-    options?: PrepareQueryOptions
-  ): PreparedQuery {
+  public getPreparedQueryByIndex(index: number): PreparedQuery {
```
christopherswenson, line 752: Same.

packages/malloy/src/malloy.ts
```diff
@@ -803,23 +798,22 @@ export class Model implements Taggable {
...
+  public getPreparedQuery(): PreparedQuery {
     if (this.queryList.length === 0) {
       throw new Error('Model has no queries.');
     }
     return new PreparedQuery(
       this.queryList[this.queryList.length - 1],
       this.modelDef,
-      this.problems,
-      options?.replaceMaterializedReferences ?? false
+      this.problems
```
christopherswenson, line 816: Same.

packages/malloy/src/malloy.ts
```diff
@@ -908,11 +901,15 @@ export class PreparedQuery implements Taggable {
...
+    return this.getPreparedResult();
+  }
+
+  public getPreparedResult(options?: PrepareResultOptions): PreparedResult {
     const queryModel = new QueryModel(this._modelDef);
     const translatedQuery = queryModel.compileQuery(
       this._query,
       false,
-      this.replaceMaterializedReferences
+      options?.replaceMaterializedReferences ?? false
```
christopherswenson, line 912: Getting a `PreparedResult` from a `PreparedQuery` involves running the compiler, so you may want to pass in `replaceMaterializedReferences`.

packages/malloy/src/malloy.ts
```diff
@@ -1536,10 +1533,7 @@ export class Explore extends Entity implements Taggable {
     return this.structDef.as || this.structDef.name;
   }
 
-  public getQueryByName(
-    name: string,
-    options?: PrepareQueryOptions
-  ): PreparedQuery {
+  public getQueryByName(name: string): PreparedQuery {
```
christopherswenson, line 1536: Again, getting a `PreparedQuery` loses the settings.

packages/malloy/src/malloy.ts
```diff
@@ -2258,7 +2246,7 @@ export class Runtime {
    */
   public loadModel(
     source: ModelURL | ModelString,
-    options?: ParseOptions & CompileOptions
+    options?: ParseOptions & CompileOptions & PrepareResultOptions
   ): ModelMaterializer {
```
christopherswenson, line 2250: When you load a model, you might want to pass in `replaceMaterializedReferences` immediately because of SQL blocks. To make this less confusing, I've added `replaceMaterializedReferences` onto `ModelMaterializer` so that if you do:

```ts
const model = runtime.loadModel(..., { replaceMaterializedReferences: true });
const query = model.loadQuery(...);
const sql = query.getPreparedResult();
```
The `replaceMaterializedReferences` will get stored in the `ModelMaterializer` in the first call, then passed through to the `QueryMaterializer` in the second call, and then used by `getPreparedResult` when running the compiler.

packages/malloy/src/malloy.ts
```diff
@@ -2268,36 +2256,48 @@ export class Runtime {
...
   //      `ModelDef` to the world formally. For now, this should only
   //      be used in tests.
-  public _loadModelFromModelDef(modelDef: ModelDef): ModelMaterializer {
-    return new ModelMaterializer(this, async () => {
-      return new Model(modelDef, [], [], [], []);
-    });
+  public _loadModelFromModelDef(
+    modelDef: ModelDef,
+    options?: PrepareResultOptions
+  ): ModelMaterializer {
```
christopherswenson, line 2293: Same.

packages/malloy/src/malloy.ts
```diff
@@ -2309,9 +2309,9 @@ export class Runtime {
    */
   public loadQuery(
     query: QueryURL | QueryString,
-    options?: ParseOptions & CompileOptions & PrepareQueryOptions
+    options?: ParseOptions & CompileOptions & PrepareResultOptions
   ): QueryMaterializer {
```
christopherswenson, line 2313: Allows `runtime.loadQuery({ replaceMaterializedReferences: true }).run()`.

packages/malloy/src/malloy.ts
```diff
@@ -2326,9 +2326,9 @@ export class Runtime {
   public loadQueryByIndex(
     model: ModelURL | ModelString,
     index: number,
-    options?: ParseOptions & CompileOptions
+    options?: ParseOptions & CompileOptions & PrepareResultOptions
   ): QueryMaterializer {
-    return this.loadModel(model, options).loadQueryByIndex(index);
+    return this.loadModel(model, options).loadQueryByIndex(index, options);
```
christopherswenson, line 2331: Same.

packages/malloy/src/malloy.ts
```diff
@@ -2343,9 +2343,9 @@ export class Runtime {
   public loadQueryByName(
     model: ModelURL | ModelString,
     name: string,
-    options?: ParseOptions & CompileOptions
+    options?: ParseOptions & CompileOptions & PrepareResultOptions
   ): QueryMaterializer {
-    return this.loadModel(model, options).loadQueryByName(name);
+    return this.loadModel(model, options).loadQueryByName(name, options);
```
christopherswenson, line 2348: Same.

packages/malloy/src/malloy.ts
```diff
@@ -2360,7 +2360,7 @@ export class Runtime {
   public loadSQLBlockByName(
     model: ModelURL | ModelString,
     name: string,
-    options?: ParseOptions & CompileOptions
+    options?: ParseOptions & CompileOptions & PrepareResultOptions
   ): SQLBlockMaterializer {
```
christopherswenson, line 2364: Allows `runtime.loadSQLBlockByName(..., { replaceMaterializedReferences: true })`. This is probably not super relevant, as SQL blocks are deprecated in the language.

packages/malloy/src/malloy.ts
```diff
@@ -2563,15 +2563,17 @@ class FluentState<T> {
...
   }
 
   protected makeExploreMaterializer(
-    materialize: () => Promise<Explore>
+    materialize: () => Promise<Explore>,
+    options?: PrepareResultOptions
   ): ExploreMaterializer {
-    return new ExploreMaterializer(this.runtime, materialize);
+    return new ExploreMaterializer(this.runtime, materialize, options);
   }
```
christopherswenson, line 2577: `ExploreMaterializer` can also store a `replaceMaterializedReferences` so that you can say 
```ts
await runtime
  .loadModel(..., { replaceMaterializedReferences: true })
  .loadExploreByName("foo")
  .loadQueryByName("bar")
  .run()
```

Note that `.loadExploreByName()` doesn't have the `replaceMaterializedReferences` option. I could go either way on this, but noting in loading an explore actually needs the setting...

packages/malloy/src/malloy.ts
```diff
@@ -2593,16 +2595,33 @@ class FluentState<T> {
...
+  public loadFinalQuery(options?: PrepareResultOptions): QueryMaterializer {
+    return this.makeQueryMaterializer(
+      async () => {
+        return (await this.materialize()).getPreparedQuery();
+      },
+      {
+        replaceMaterializedReferences: this.replaceMaterializedReferences,
+        ...options,
+      }
+    );
```
christopherswenson, line 2624: Can say `modelMaterializer.loadFinalQuery({ replaceMaterializedReferences: true })`

Being able to do this this late might lead to some weird results... If you created a `ModelMaterializer` without passing in `{ replaceMaterializedReferences: true }`, then passed it in late, you would end up with something that had SQL blocks not use materialization, but queries using it:

```
# materialize
query: step_one is duckdb.table("foo") -> {
  select: a, b, c
}

# materialize
query: step_two is duckdb.sql("""
  SELECT * FROM (%{ step_one }) WHERE ...
""")

# materialize
query: step_three is step_two -> { ... }

run: step_three
```

In this case, you'd end up with `step_one` not being materialized.

test/src/databases/duckdb/materialization.spec.ts
```diff
@@ -33,12 +33,12 @@ describe.each(allDucks.runtimeList)('duckdb:%s', (dbName, runtime) => {
     `;
 
     const qm = runtime.loadQuery(query, {replaceMaterializedReferences: true});
-    const pq = await qm.getPreparedQuery();
+    const preparedResult = await qm.getPreparedResult();
 
-    expect(pq.preparedResult.sql).toBe(
+    expect(preparedResult.sql).toBe(
```
christopherswenson, line 38: Note here we get the prepared result still in runtime world.

Doing `qm.getPreparedQuery().preparedResult` would end up not using the setting.

nachoarreola: should we deprecate the getter at some point?

christopherswenson: Yeah probably. We could even just add a `@deprecated` to the doc string right now. My event logging PR also adds a setting to `get preparedResult`/`getPreparedResult()`.

PR #1954, opened 2024-10-08
Implement case with warning

christopherswenson:
Adds the SQL case syntax:
```
case when some_condition then some_value else default_value end
case lhs_value when rhs_value then some_value else default_value end
```
This receives a warning like "Use a `pick` statement instead of `case`" with a replacement of `pick some_value when some_condition else default_value`.


mtoy-googly-moogly:
I would check in malloy/lang/test/expressions.spec to make sure that the correct expr is being generated there is a nifty matcher for you `compilesTo` because you want to generate the correct Expr ... and then you want to unsure that Expr is being translated properly in the live database test

PR #1956, opened 2024-10-09
Various changes for "IN"

mtoy-googly-moogly:
Here is the plan

- [x] Make `expr = a|b|c` generate the SQL `expr IN (a,b,c)`
- [x] Make `expr = ANY OTHER TREE OR PARTIAL` warn, and still evaluate
- [x] Make `~` and `!~` also accept or-trees and warn on other trees, but still expand to boolean expressions
- [x] Parse `expr IN (a,b,c)` warn, and suggest the Malloy syntax


packages/malloy/src/lang/malloy-to-ast.ts
```diff
@@ -1996,4 +1996,27 @@ export class MalloyToAST
...
+      ),
+      pcx
+    );
+    this.warnWithReplacement(
+      'sql-in',
+      `Use = (a|b|c) instead of${isNot ? ' NOT' : ''} IN (a,b,c)`,
+      this.parseInfo.rangeFromContext(pcx),
+      `${this.getSourceCode(pcx.fieldExpr())} ${isNot ? '!=' : '='} (${from
+        .map(f => this.getSourceCode(f))
+        .join('|')})`
```
christopherswenson, line 2018: I might format this with `.join(' | ')` to generate `x != (a | b | c)` rather than `x != (a|b|c)`, and honestly I would probably remove the parens, and just suggest `x != a | b | c`

packages/malloy/src/lang/test/expressions.spec.ts
```diff
@@ -198,8 +198,8 @@ describe('expressions', () => {
       );
     });
     test('apply granular-literal alternation uses all literals for range', () => {
-      expect('ad ? @2020 | @2022').compilesTo(
-        '{{{ad >= @2020-01-01} and {ad < @2021-01-01}} or {{ad >= @2022-01-01} and {ad < @2023-01-01}}}'
+      expect('ad ? @2020').compilesTo(
+        '{{ad >= @2020-01-01} and {ad < @2021-01-01}}'
```
christopherswenson, line 202: Why is this test changing?

packages/malloy/src/lang/test/parse-expects.ts
```diff
@@ -240,6 +240,11 @@ function eToStr(e: Expr, symbols: ESymbols): string {
       return e.node;
     case 'regexpMatch':
       return `{${subExpr(e.kids.expr)} regex-match ${subExpr(e.kids.regex)}}`;
+    case 'in': {
+      return `{${subExpr(e.kids.e)} ${e.not ? 'not in' : 'in'} ${e.kids.oneOf
+        .map(o => `{${subExpr(o)}}`)
```
christopherswenson, line 245: I might put the `{}` around the whole list rather than each element? Just for consistency with the other formatters here, normally single values don't get `{}` around them: `x in (1, 2, 3)` would then be `{x in {1,2,3}}` rather than `{x in {1},{2},{3}}`. And maybe `.join(', ')` 😊 

Not that it matters...

packages/malloy/src/lang/test/expressions.spec.ts
```diff
@@ -1116,6 +1116,54 @@ describe('expressions', () => {
...
+  });
+  test('a=b|>d', () => {
+    expect('ai=1|>2').compilesTo('{{ai = 1} or {ai > 2}}');
+    expect(expr`ai=1|>2`).toLog(
+      warningMessage(
+        'Only | seperated values are legal when used with = operator'
+      )
+    );
+  });
+  test('a ? = (b | c)', () => {
```
christopherswenson, line 1155: Test name does not match tested expression.

packages/malloy/src/lang/test/expressions.spec.ts
```diff
@@ -1116,6 +1116,54 @@ describe('expressions', () => {
...
+  });
+  test('legacy in', () => {
+    const inExpr = expr`ai in (1,2,3)`;
+    expect(inExpr).compilesTo('{ai in {1},{2},{3}}');
+    expect(inExpr).toLog(warningMessage('Use = (a|b|c) instead of IN (a,b,c)'));
+  });
+  // mtoy todo this hits "alternation tree has no value"
+  test.todo('a ? (= 1) | 2');
+});
+describe('rigor around ? and =', () => {});
```
christopherswenson, line 1166: Are you planning on filling this in?

christopherswenson (approved):
Looks great to me! A few small comments which you may ignore if you want.

packages/malloy/src/lang/malloy-to-ast.ts
mtoy-googly-moogly: I was worried about "x in (a,b,c) and y = z" type of expressions turning into x = a | b | c and y = z so i left the parens in.

I went back and forth on spaces, went with "|" but have no strong attachment, you want me to change it?

packages/malloy/src/lang/test/parse-expects.ts
mtoy-googly-moogly: ok except maybe not comma space

packages/malloy/src/lang/test/expressions.spec.ts
mtoy-googly-moogly: Yup, at one point this was being generated as an IN and I was toying with different ways to address it and left the test change in, i'll put it back.

mtoy-googly-moogly: however i changed it because the test was "does an apply turn into a range computation" not "does an apply of two range computations turn into an "IN". I was intending to add the second test and forgot

packages/malloy/src/lang/malloy-to-ast.ts
mtoy-googly-moogly: added space

packages/malloy/src/lang/test/expressions.spec.ts
mtoy-googly-moogly: resolved as a long standing bug in our grammar, noted this and left it skipped

PR #1969, opened 2024-10-16
Add `alwaysJoins` to `PipeSegment`; joins in queries are allowed again, and generate always joins

christopherswenson:


christopherswenson:
This works for views that are not nested. Issue for nested source extensions not working: https://github.com/malloydata/malloy/issues/1970

PR #1988, opened 2024-10-30
Add percent_rank presto function

christopherswenson:


carlineng (approved):
Thanks. I see there's a parameter specifically for `needsWindowOrderBy`. For partitioning, do all `calculate` functions allow `partition_by` clauses by default?

christopherswenson:
Yes, `partition_by` should be legal on all window functions

PR #1993, opened 2024-11-01
Fix issue where leading non `isCastType` arguments would both a) not set inferred type of native SQL function calls and b) not generate a comma in the function call SQL

christopherswenson:


packages/malloy/src/lang/ast/expressions/expr-func.ts
```diff
@@ -127,23 +127,21 @@ export class ExprFunc extends ExpressionDef {
...
-            collectType = expr.type;
-          }
         }
         funcCall.push(expr.value);
-      }
+      });
       funcCall.push(')');
 
-      const dataType = this.rawType ?? collectType ?? 'number';
+      const inferredType = argExprsWithoutImplicit[0] ?? {type: 'number'};
```
christopherswenson, line 139: This is a little strange to set the result type to just the whole expression of the first argument... but the type system is happy with it because a `TypeDesc` _is_ a `ExpressionValueTypeDef`. @mtoy-googly-moogly if there were a way to extract only the `ExpressionValueTypeDef` parts of a `TypeDesc`, then I would use that here.

Issue #1927, opened 2024-09-25
wrong sorting when nesting views


wwasilak:
### What happens?

I am nesting two views in query. Each of those views is sorted by date field and when run separately the sorting works fine. The problem is, when nested in a query the sorting is broken (it also happens when only 1 view is nested):

![image](https://github.com/user-attachments/assets/b9af39b3-fc36-4064-8ee0-5e78a709502b)

### To Reproduce

view: base is {
    where: order_date >= @2024-07-31
        and (product_id = 441 | 493)
    group_by: 
            order_date
            flag is 1
    aggregate: 
            metric is total_quantity
    }
    
view: from_base1 is base ->
{
    group_by: order_date
    aggregate: total is  metric.sum()
    order_by: order_date asc
}

view: from_base2 is base ->
{
    group_by: order_date
    aggregate: flag.sum()
    order_by: order_date asc
}

view: nested is {
    where: order_date >= @2024-07-31
        and (product_id = 441 | 493)
    group_by: product_id
    aggregate: total_quantity
    nest: from_base1, from_base2
}

### OS:

Windows 10

### Malloy Client:

VSCode

### Malloy Client Version:

v0.3.1727197487 (pre-release)

### Database Connection:

DuckDB

carlineng:
@wwasilak I'm having trouble reproducing this with our sample data. Are you willing to share the compiled SQL from when you run the `nested` view in the model above?

carlineng:
@wwasilak thanks for the additional info. I was able repro in this query:

https://github.com/malloydata/malloy-samples/blob/7b1bfd6d50b3a4c11617e3f0ab9e998dc1936513/ecommerce/ecommerce.malloy#L237

carlineng:
To add some color on what's happening -- it appears the ORDER BY clause in the nested view is not producing correct results, specifically when the nested view is a pipelined view, based off another view. In this example, `base` is a view, and `total_orders_v` is a view that starts with `base` and adds a pipeline segment.

carlineng:
Super strange -- I can reproduce this behavior locally on the VS Code extension pre-release v0.3.1727197487; however, when I try to run it in the DuckDB-WASM version (also on pre-release v0.3.1727197487), I cannot reproduce.

carlineng:
The query produces SQL that looks like the following:

```sql
WITH __stage0 AS (
  SELECT
    group_set,
    inventory_items_0."product_category" as "product_category__0",
[...]

PR #2022, opened 2024-11-26
nested pipelines generate bad order by

lloydtabb:
in nested pipelines, we didn't add order by in duckdb.


lloydtabb:
Fixes https://github.com/malloydata/malloy/issues/1927

PR #2045, opened 2024-12-12
DO NOT MERGE: hack to save code

lloydtabb:


lloydtabb:
uses code like this to find the cube elements

```
source: cube_build is presto_xxx.table('peu_cube_agg_level:xxx') extend {
  view: agg_levels is {
    where: ds = '2024-12-08'
    group_by: agg_level
    aggregate: row_count is count()
    limit: 10000
    order_by: row_count
  }
}
```

PR #2067, opened 2024-12-28
Expand null safety

mtoy-googly-moogly:


lloydtabb:
> x = y is null when x and y are null so it looks falsey when it should be truthy

I believe this is wrong for Malloy (and SQL).  These comparisons are vector operations, not generally tests like in other languages.  The question is 'should I include these records in the result set'.  When you are testing for  x = y, you want records where they match, if a null is involved you generally don't want the records.  I have never, ever written, in SQL

     WHERE X = Y or (X IS NULL and y IS NULL)

IMO it would be a mistake to try and do this.

lloydtabb:
Generally, in SQL is there is a null in an expression, then the results are null.   In SQL 'OR' is an exception to this rule.

image

I don't think we need to 'fix' this, unless we want to make all relational operators 'truthy' and in that, it is a `coalese(x, false)`

mtoy-googly-moogly:
We have decided NOT to expand null safety.

PR #2069, opened 2024-12-31
Added the windows equivalent command to the unix storybook script. 

anagrd-ai:
Eg "npm run storybook-windows".  The unix version fails on a windows machine.

### DIFF
--- a/packages/malloy-render/package.json
+++ b/packages/malloy-render/package.json
@@ -24,6 +24,7 @@
     "clean": "tsc --build --clean",
     "prepublishOnly": "npm run build",
     "storybook": "rm -rf ./node_modules/.cache && storybook dev -p 6006",
+    "storybook-windows": "del /s /q .\\node_modules\\.cache && storybook dev -p 6006",
     "build-storybook": "storybook build",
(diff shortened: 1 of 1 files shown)


packages/malloy-render/package.json
```diff
@@ -24,6 +24,7 @@
     "clean": "tsc --build --clean",
     "prepublishOnly": "npm run build",
     "storybook": "rm -rf ./node_modules/.cache && storybook dev -p 6006",
+    "storybook-windows": "del /s /q .\\node_modules\\.cache && storybook dev -p 6006",
```
srenatus, line 27: just a bystander comment: [rimraf](https://www.npmjs.com/package/rimraf) is what often gets used for cross-platform removal of old stuff. You'd use `rimraf` as CLI commands like [in this random example](https://github.com/observablehq/framework/blob/main/package.json#L25)

mtoy-googly-moogly: @srenatus thanks for the tip! wish you had beat me to the review.

srenatus: It was merely a chance find on my end. I hadn't been looking at Malloy in a long while 😅

PR #2071, opened 2025-01-02
Make CI Faster

mtoy-googly-moogly:
This is basically three things

1) Divide tests into "core" and "dialect"
2) Run "core" tests only once
3) For all dialects where possible, use jest workers to test in parallel

I need to

- [x] Add a test to make sure that all tests in "jest --listTests" exist in at least one `jest.*.config`


.github/workflows/run-tests.yaml
```diff
@@ -11,35 +25,30 @@ jobs:
...
+      - db-duckdb
+      - db-snowflake
+      - db-duckdb-wasm
       - db-motherduck
       - db-mysql
       - db-postgres
-      - db-presto
-      - db-snowflake
-      - db-trino
-      - main
```
whscullin, line 42: Curious why the order change here?

mtoy-googly-moogly: per conversation, longest to shortest, updated list and added comment.

PR #2072, opened 2025-01-02
fix case where the nested field can have multiple types

amitaggarwal1:
fix case where the nested field can have multiple types (based on the 100 sample).

In most cases, we should fail but in cases where the field types are integer and decimal we should not fail and make the field decimal. this can happen when there is a nested field and sometimes has a float value and sometimes an integer value


mtoy-googly-moogly (approved):
LGTM

mtoy-googly-moogly (commented):
Actually I need to think about this.

Issue #1481, opened 2023-11-08
Error using a dimension w/ a nested field reference in a join


carlineng:
### What happens?

I have a Source with a nested field, and I create a dimension on that Source that references the nested field:

```
source: raw_vals is bigquery.table(`mytable`) extend {
  dimension:  myval is nested_field.nested_val
}
```

I create another Source and attempt to use this dimension as a join key:

```
source: grouped_tbl is raw_vals -> {
  group_by: myval
} extend {
  join_many: raw_vals on myval = raw_vals.myval
}
```

The SQL generated by this query fails to compile:

```
Unrecognized name: nested_field_0; Did you mean nested_field? at [13:43]
```

See the example in [this file](https://github.com/malloydata/malloy-samples/blob/a194dcea7c51d986b2d195729dbc6e32cb14b69e/bigquery/ga4/join_nest.malloy#L12-L15)

### To Reproduce

Run the example here: https://github.com/malloydata/malloy-samples/blob/a194dcea7c51d986b2d195729dbc6e32cb14b69e/bigquery/ga4/join_nest.malloy#L12-L15

In the malloy-samples repo in the [ce/unnest_join](https://github.com/malloydata/malloy-samples/tree/ce/unnest_join) branch

### OS:

macOS

### Malloy Client:

Local VSCode

### Malloy Client Version:

v0.3.1699380172

### Database Connection:

BigQuery

christopherswenson:
Related: though this issue is probably easier to fix than mine https://github.com/malloydata/malloy/issues/1482

lloydtabb:
It is the same bug.  The problem is that the join predicate implies an UNNEST and that UNNEST has to happen before the the table is joined in.  I don't have an elegant solution to solve this just yet.

lloydtabb:
So this is what the SQL looks like when you try and join on a nested field.  The problem is that the joins are all performed in order.

When we get to the on clause, a.value is not yet defined.

```
SELECT 
   ....
 FROM base
 LEFT JOIN table_with_nest ON base.id = a.value
 LEFT JOIN UNNEST (table_with_nest.some_array) as a(value)
```

We could, when unnesting, unnest all the related data into a single alias.  

```
SELECT 
   ....
 FROM base
 LEFT JOIN (
   SELECT t.*,  nest.nest_value as unique_name FROM  table_with_nest t
      LEFT JOIN UNNEST (table_with_nest.some_array) as nest(nest_value)
[...]

PR #2074, opened 2025-01-03
Fix snowflake error : "Lateral cannot be on the left side of join"

amitaggarwal1:
For nested fields we were using lateral flatten. The issue with it is that if the

nested field is used in a join, snowflake complains that lateral flatten can't be on left side of join (https://stackoverflow.com/questions/63397022/snowflake-lateral-cannot-be-on-the-left-side-of-join)

fix is to always convert lateral flatten to a left join

Also simplify the array[scalar] vs array[record] treatment. earlier, at the unnesting step we were creating a object for array[scalar] case to make it consistent with the array[record] case. Now at the reference time, based on the case, use the right expression
- array[record] : parent.value:sql_field
- array[scalar] : parent.value

### DIFF
--- a/packages/malloy/src/dialect/snowflake/snowflake.ts
+++ b/packages/malloy/src/dialect/snowflake/snowflake.ts
@@ -204,7 +204,7 @@ export class SnowflakeDialect extends Dialect {
   ): string {
     const as = this.sqlMaybeQuoteIdentifier(alias);
     if (isArray) {
-      return `,LATERAL FLATTEN(INPUT => ${source}) AS ${alias}_1, LATERAL (SELECT ${alias}_1.INDEX, object_construct('value', ${alias}_1.value) as value ) as ${as}`;
+      return `LEFT JOIN lateral flatten(input =>  ${source}) as ${as}`;
     } else {
       // have to have a non empty row or it treats it like an inner join :barf-emoji:
       return `LEFT JOIN LATERAL FLATTEN(INPUT => ifnull(${source},[1])) AS ${as}`;
@@ -266,7 +266,16 @@ export class SnowflakeDialect extends Dialect {
       parentType === 'array[scalar]' ||
       parentType === 'array[record]'
     ) {
-      const arrayRef = `"${parentAlias}".value:${sqlName}`;
+      // Case 1: this is an array of scalar. We can reference the field
+      // simply as parent.value. in fact the sqlName will be value always
+      // because there is nothing else to reference.
+      // Case 2: this is an array of records. In this case, parent.value
+      // gives the record and then parent.value:sql_name gives the
+      // right field in the record.
+      let arrayRef = `"${parentAlias}".value:${sqlName}`;
+      if (parentType === 'array[scalar]') {
+        arrayRef = `"${parentAlias}".value`;
+      }
       switch (childType) {
         case 'record':
         case 'array':
--- a/test/src/databases/all/compound-atomic.spec.ts
+++ b/test/src/databases/all/compound-atomic.spec.ts
@@ -91,6 +91,13 @@ describe.each(runtimes.runtimeList)(
     """)`;
 
     describe('simple arrays', () => {
+      test('join on scalar array field', async () => {
+        const nestedArrayJoin = `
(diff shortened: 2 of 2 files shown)


packages/malloy/src/dialect/snowflake/snowflake.ts
```diff
@@ -204,7 +204,7 @@ export class SnowflakeDialect extends Dialect {
   ): string {
     const as = this.sqlMaybeQuoteIdentifier(alias);
     if (isArray) {
-      return `,LATERAL FLATTEN(INPUT => ${source}) AS ${alias}_1, LATERAL (SELECT ${alias}_1.INDEX, object_construct('value', ${alias}_1.value) as value ) as ${as}`;
```
lloydtabb, line 207: I would imagine some tests are failing.  The INDEX is the distinct key and necessary for deduping.

It would also be great to have a test that shows what this fixes..

amitaggarwal1: Will work on fixing the tests. 

Re: INDEX, INDEX is already part of the flatten structure. This was taking something of the form {INDEX: i, value: v} and converting into {INDEX: i, value: {value: v}}

amitaggarwal1: @lloydtabb added test case. this test fails without the fix with the following error:

```
 query.run failed: SQL compilation error:
    Lateral View cannot be on the left side of join
    SQL: SELECT
       a_0."a_two" as "a_two"
    FROM (
          SELECT [2,4,6,8] AS "evens"
        ) as base
    ,LATERAL FLATTEN(INPUT => base."evens") AS evens_0_1, LATERAL (SELECT evens_0_1.INDEX, object_construct('value', evens_0_1.value) as value ) as "evens_0"
     LEFT JOIN ( SELECT 2 as "a_two" ) AS a_0
      ON a_0."a_two"="evens_0".value:"value"::DOUBLE

    OperationFailedError: SQL compilation error:
    Lateral View cannot be on the left side of join

```

lloydtabb:
So this is a bigger (and long standing) problem. Malloy, generally, can't join on a nested field.  I think I finally know now to fix it, but it is a pretty major rework of code generation.  It's probably worth scheduling the work.

I'm curious to see how your test fares on the other dialects. 

https://github.com/malloydata/malloy/issues/1481

We're working on figuring out how to get tests running against external pull requests (that's the reason for the delay in merging).

lloydtabb:
You can see that it fails in other dialects.

https://github.com/malloydata/malloy/actions/runs/12605890168/job/35135261018?pr=2074#step:4:223

I'll write up what I think the solution is in the https://github.com/malloydata/malloy/issues/1481

test/src/databases/all/compound-atomic.spec.ts
```diff
@@ -91,6 +91,13 @@ describe.each(runtimes.runtimeList)(
     """)`;
 
     describe('simple arrays', () => {
+      test('join on scalar array field', async () => {
+        const nestedArrayJoin = `
```
mtoy-googly-moogly, line 95:
```TypeScript
        const nestedArrayJoin = `
           source: a is ${conName}.sql(""" SELECT 2 as ${quote('a_two')} """) extend {}
           source: b is ${evens} extend { join_one: a is a on a.a_two = evens.value}
           run: b -> {select: a_two is a.a_two}`;

File: packages/malloy/src/dialect/snowflake/snowflake.ts [hunk 3]
```diff
@@ -266,7 +266,16 @@ export class SnowflakeDialect extends Dialect {
       parentType === 'array[scalar]' ||
       parentType === 'array[record]'
     ) {
-      const arrayRef = `"${parentAlias}".value:${sqlName}`;
+      // Case 1: this is an array of scalar. We can reference the field
```
mtoy-googly-moogly, line 269: I'm not 100% sure I know what is going because I don't have the un-nested state of the array exactly in my head at this moment, but something like this would be preferable, if this is the correct solution.

```TypeScript
      const arrayRef = `"${parentAlias}".value` + (parentType === 'array[record]' ? `:${sqlName }`: '');
```

mtoy-googly-moogly:
#2086 has these changes in it

PR #2079, opened 2025-01-03
Replace pull_request with pull_request_target

whscullin:
Take 2, replacing instead adding. With some opinionated linting from ... somewhere. 

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks


mtoy-googly-moogly (approved):
👍

whscullin:
Not a good idea.

PR #2088, opened 2025-01-09
Snowflake connector: handle timeout and abortSignal

carinakoo8:
Support timeout and abortSignal in Snowflake connector.


mtoy-googly-moogly:
#2086 has these changes in it

PR #2113, opened 2025-01-31
Improved handling of null or non-numeric values.

anagrd-ai:
This line was erroring when queries returned null numerics:
  Cannot read properties of undefined (reading 'toLocaleString')

See T214044108 for details

### DIFF
--- a/packages/malloy-render/src/component/render-numeric-field.ts
+++ b/packages/malloy-render/src/component/render-numeric-field.ts
@@ -54,13 +54,8 @@ export function renderNumericField(f: AtomicField, value: number): string {
   } else if (tag.has('number'))
     displayValue = format(tag.text('number') ?? '#', value);
   else {
-    const numericValue = Number(value ?? '');
-    if (!isNaN(numericValue)) {
-      displayValue = numericValue.toLocaleString();
-    } else {
-      // handle the case where value is not a number
-      displayValue = String(value ?? '');
-    }
+    const nonNullValue = value ?? '';
+    displayValue = (nonNullValue as number).toLocaleString();
   }
   return displayValue;
 }
--- a/packages/malloy-render/src/component/render-numeric-field.ts
+++ b/packages/malloy-render/src/component/render-numeric-field.ts
@@ -25,8 +25,12 @@ import {AtomicField} from '@malloydata/malloy';
 import {Currency, DurationUnit} from '../html/data_styles';
 import {format} from 'ssf';
 import {getText} from '../html/duration';
+import {NULL_SYMBOL} from './apply-renderer';
 
-export function renderNumericField(f: AtomicField, value: number): string {
+export function renderNumericField(f: AtomicField, value: number | null | undefined): string {
+  if (value === null || value === undefined) {
+    return NULL_SYMBOL;
+  }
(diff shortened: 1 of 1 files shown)


packages/malloy-render/src/component/render-numeric-field.ts
```diff
@@ -53,6 +53,14 @@ export function renderNumericField(f: AtomicField, value: number): string {
     );
   } else if (tag.has('number'))
     displayValue = format(tag.text('number') ?? '#', value);
-  else displayValue = (value as number).toLocaleString();
+  else {
+    const numericValue = Number(value ?? '');
```
nachoarreola, line 57: Is Number a native Node method?

packages/malloy-render/src/component/render-numeric-field.ts
```diff
@@ -53,6 +53,14 @@ export function renderNumericField(f: AtomicField, value: number): string {
...
+    if (!isNaN(numericValue)) {
+      displayValue = numericValue.toLocaleString();
+    } else {
+      // handle the case where value is not a number
+      displayValue = String(value ?? '');
```
nachoarreola, line 62: when its NaN but not null... should we actually show NaN or something like that? its probably better than blowing the whole thing but theres def something wrong, this way we can debug.

anagrd-ai: Agreed.  See the updated version.

packages/malloy-render/src/component/render-numeric-field.ts
anagrd-ai: Agreed.  I was over thinking this.  Replaced with a much simpler approach.

packages/malloy-render/src/component/render-numeric-field.ts
```diff
@@ -53,6 +53,9 @@ export function renderNumericField(f: AtomicField, value: number): string {
     );
   } else if (tag.has('number'))
     displayValue = format(tag.text('number') ?? '#', value);
-  else displayValue = (value as number).toLocaleString();
+  else {
+    const nonNullValue = value ?? '';
+    displayValue = (nonNullValue as number).toLocaleString();
+  }
```
whscullin, line 59: If we're making this function handle non-numeric values, then we should update the signature to indicate that. I'd be more curious to know where in the chain of events something we're convinced is a number is coming through as a null.

If we are handling `null` here, we should probably do that up top, and return a `∅`.

anagrd-ai: Thanks!  Now returns the null symbol.

anagrd-ai: Also created [T214125835] to add tests for the render package.  There are some tests of the charts already, but no tests of these "render*.ts" functions.  I am also curious where it fails.  The query that originally yielded this error started returning results as expected shortly after, and we are no longer able to reproduce.   Fernando has suggested setting up probers against tables with null values to monitor these sorts of issues going forward.  Indeed probers of any sort would help us detect outages abnd possibly find add'l bugs in this code.

whscullin (approved):
LGTM

PR #2121, opened 2025-02-04
Really switch to pull_request_target

whscullin:
After all the excitement of completely restructuring our test builds I didn't make the actual final cut over to pull_request_target


whscullin:
Note none of these tests will run because the main branch doesn't have pull_request_target as the trigger.

PR #2122, opened 2025-02-04
Implement the test functionality for trino connections

mbooth4:
In order to provide a good user experience in the new Trino and Presto connection editors, we need the `test` functionality to actually work.

Implement that for the `TrinoPrestoConnection`.

Associated with https://github.com/malloydata/malloy-vscode-extension/issues/589

Example:
image

### DIFF
--- a/packages/malloy-db-trino/src/trino_connection.ts
+++ b/packages/malloy-db-trino/src/trino_connection.ts
@@ -22,7 +22,6 @@
  */
 
 import {
-  Connection,
   ConnectionConfig,
   MalloyQueryData,
   PersistSQLResults,
@@ -44,6 +43,7 @@ import {
   FieldDef,
   TinyParser,
   isRepeatedRecord,
+  TestableConnection,
 } from '@malloydata/malloy';
 
 import {BaseConnection} from '@malloydata/malloy/connection';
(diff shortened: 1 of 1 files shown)


packages/malloy-db-trino/src/trino_connection.ts
```diff
@@ -422,7 +422,7 @@ export abstract class TrinoPrestoConnection
   }
 
   public async test(): Promise<void> {
-    // await this.dryRunSQLQuery('SELECT 1');
+    await this.runSQL('SELECT 1');
```
whscullin, line 425: One issue we had (I think it was with BigQuery) is that you can do a `SELECT 1` on an unauthenticated connection. Curious if that could be an issue here.

mbooth4: That doesn't appear to be a problem here, but that could be a result from only testing against our local `docker` containers, which seem very permissive in terms of authentication.

mbooth4: It also looks like our existing Presto connection doesn't actually pass in authorization information at all... I'm looking into it further

whscullin (approved):
LGMT

whscullin:
🚢 🇮🇹

PR #2123, opened 2025-02-07
Add two minimal .malloy test files for presto and trino

mbooth4:
While testing the connection editors for trino and presto, I built small malloy files I could use for manual testing. I figured it would be useful to check in for future use cases, because I could not find any other `trino` or `presto` connections represented in our repos or documentation.


whscullin:
Presto and yrino do go through the full database test suite, although there's not specifically presto or trino test files. I don't have an issue with having some sample malloy files, curious what @lloydtabb thinks...

PR #2128, opened 2025-02-11
First pass at new duckdb node api.

whscullin:


Alex-Monahan:
Howdy folks! I am playing around with Malloy and I wanted to integrate it with some other code using the Node Neo library. Is this branch viable to use? Any plans to move to Node Neo?

PR #2153, opened 2025-02-21
Node 20

whscullin:
Node 18.x is EOL on April 30th, move to supported release.


whscullin:
Will coordinate with other repos.

PR #2154, opened 2025-02-21
Added malloy-filter as a package under malloy

anagrd-ai:
Copied from:
https://github.com/malloydata/malloy-filter/

Verified that all files build correctly, tests pass, sample outputs generated as expected.


packages/malloy-filter/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@malloydata/malloy-filter",
+  "version": "1.0.0",
```
whscullin, line 3: This should match the version of other packages.

packages/malloy-filter/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "sourceMap": true,
+    "rootDir": "src",
+    "outDir": "dist",
+    "strict": true,
+    "esModuleInterop": true,
+    // Prevent tsc from picking up modules from higher directories.
+    "moduleResolution": "node"
+  },
+  "include": ["src/**/*", "jest.config.js"],
+}
```
whscullin, line 14: Should match the tsconfig from other packages so we're generating consistent output.

packages/malloy-filter/package.json
anagrd-ai: Done

packages/malloy-filter/tsconfig.json
anagrd-ai: Thanks for pointing this out.  I was copying other tsconfigs I found under malloy/packages/ but there were a few lingering inconsistencies.  I removed those.  It should (I hope) match what is found in other packages now.

christopherswenson:
I think you might be missing a few places to list this new package. If you do a code search for `malloy-interfaces`, you should see a few more places where all the packages are listed, where you'll need to add this one, e.g. `/Users/cswenson/Documents/malloy/.github/workflows/prerelease.yaml`

whscullin:
The release scripts pull from package.json as of a couple weeks ago

christopherswenson:
@whscullin Oh cool. I guess I should merge main into my WIP branch... 😊

anagrd-ai:
I added the malloy-filter package to additional config files.  Holler if I missed any.  Otherwise looks green test-wise.  Is there anything else I can double-check to ensure the new package is correctly integrated?

PR #2177, opened 2025-03-05
API: Handle race condition with multiple sessions compiling the same source/model/query

christopherswenson:
I feel pretty silly for overlooking this... 

If two clients tried to compile the same model, they would share the same session and stomp on each other.

So now the methods all accept a `session_id` parameter. This is required to reuse a session. In the first call, don't pass a `session_id` (or if you do and it doesn't correspond to a session, it will be ignored). The first response (and all subsequent responses) will have a `session_id` that you can use to keep the session state.

---

Also changed: `ttl` in subsequent requests is no longer ignored, and will cause the expiry date to update. This allows TTL to be used like a "maximum allowed time between requests" feature by always passing e.g. `{ seconds: 5 }`.

### DIFF
--- a/packages/malloy/src/api/sessioned.spec.ts
+++ b/packages/malloy/src/api/sessioned.spec.ts
@@ -315,6 +315,65 @@ ORDER BY 1 asc NULLS LAST
         // New session
         expect(result.session_id).not.toBe(session_id);
       });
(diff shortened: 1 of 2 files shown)


nachoarreola:
Don't feel silly you are doing a lot :) thanks for fixing this!

PR #2178, opened 2025-03-07
add extra fields to the trino connector

Vadrrrr:
Add extra field to Trino connector for passing extra values

This field allows us to pass extra values like extraHeaders, SSL, and can support additional fields in the future.

### DIFF
--- a/packages/malloy-db-trino/src/trino_connection.ts
+++ b/packages/malloy-db-trino/src/trino_connection.ts
@@ -56,7 +56,7 @@ import {
   PrestoQuery,
 } from '@prestodb/presto-js-client';
 import {randomUUID} from 'crypto';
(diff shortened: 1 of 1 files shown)


packages/malloy-db-trino/src/trino_connection.ts
```diff
@@ -146,6 +147,7 @@ class TrinoRunner implements BaseRunner {
       server: config.server,
       schema: config.schema,
       auth: new BasicAuth(config.user!, config.password || ''),
+      ...config.extraConfig,
```
rvalek, line 150: perhaps it'd be better to put this at the top of the object construction so that it couldn't overshadow the named arguments

whscullin:
Thanks you for your contribution!

PR #2180, opened 2025-03-10
Flow interfaces

whscullin:


christopherswenson:
If we can find a way to put the gen_flow code all in one place (maybe a top level scripts directory), I think that would be good. Otherwise, seems fine to me.

PR #2191, opened 2025-03-13
Add experimental support for r// regexes

mtoy-googly-moogly:


mtoy-googly-moogly:
The committee has voted and decided not to do this.

PR #2200, opened 2025-03-14
Add `convertToThrift` and `convertFromThrift` functions

christopherswenson:


packages/malloy-interfaces/src/nest_unions.ts
```diff
@@ -66,11 +106,65 @@ export function unnestUnions(obj: unknown, type: string): unknown {
...
+export function convertToThrift(obj: unknown, type: string): unknown {
+  if (obj === null || obj === undefined) {
+    return obj;
+  } else if (typeof obj === 'number') {
+    return obj;
+  } else if (typeof obj === 'string') {
+    if (type === 'string') return obj;
+    const typeDefinition = getType(type);
+    if (typeDefinition.type === 'enum') {
+      return typeDefinition.values.indexOf(obj) + 1;
```
nachoarreola, line 135: Are you able to grab the id that you define in the thrift? not just the position in the array. I think that'd be the correct way.

This works because right now you have:
```
enum NumberSubtype {
  INTEGER = 1,
  DECIMAL = 2,
}
```

what if you had:
```
enum NumberSubtype {
  INTEGER = 35,
  DECIMAL = 55,
}
```

PR #2243, opened 2025-04-02
parameters can be filter expressions

mtoy-googly-moogly:
@lloydtabb requested this. It is what he wants to use filter expressions for.

If you are declaring a filter expression parameter, you must give it a type.

```
source: x(numFilter::filter<number> is f'> 10 or = 100') is y extend { ... }
```


packages/malloy/src/lang/ast/source-elements/named-source.ts
```diff
@@ -196,6 +197,14 @@ export class NamedSource extends Source {
           parameterSpace ?? new ParameterSpace(parametersOut ?? []);
         const pVal = argument.value.getExpression(paramSpace);
         let value = pVal.value;
+        if (
+          pVal.type === 'filter expression' &&
+          parameter.type === 'filter expression' &&
+          parameter.filterType
+        ) {
+          // mtoy todo ask chris how to get a better error reporting object
```
mtoy-googly-moogly, line 205: For sure want to talk to chris about this before I merge

packages/malloy/src/lang/malloy-to-ast.ts
```diff
@@ -382,7 +419,13 @@ export class MalloyToAST
   ): ast.HasParameter[] {
     if (pcx === undefined) return [];
     this.inExperiment('parameters', pcx);
-    return pcx.sourceParameter().map(param => this.getSourceParameter(param));
+    function notNullParam(p: ast.HasParameter | null): p is ast.HasParameter {
```
mtoy-googly-moogly, line 422: Actually this is another thing I would like to talk to chris about

If there is an error in a parameter declaration, is it better to NOT define the parameter, or to somehow make it a "parameter with type 'error'"

packages/malloy/src/lang/test/parameters.spec.ts
```diff
@@ -696,4 +696,38 @@ describe('parameters', () => {
...
+      source: single_digits is ab_new(goodNumbers is f'>=0 and <= 9')
+    `).toTranslate();
+  });
+  test('filter expression parameters must have types with default values', () => {
+    expect(`
+      ##! experimental.parameters
+      source: ab_new(goodNumbers is f'[25 to 50]') is ab
+    `).toLog(
+      errorMessage(
+        "Filter expression parameters must have expicit filter type, for example 'goodNumbers::filter<string>'"
```
mtoy-googly-moogly, line 712: Would also like chris to agree to this

packages/malloy/src/model/malloy_query.ts
```diff
@@ -1406,6 +1400,66 @@ class QueryField extends QueryNode {
...
+  generateAppliedFilter(
+    context: QueryStruct,
+    filterMatchExpr: FilterMatchExpr
+  ): string {
+    let filterExpr = filterMatchExpr.kids.filterExpr;
+    while (filterExpr.node === '()') {
+      filterExpr = filterExpr.e;
+    }
+    if (filterExpr.node === 'parameter') {
+      const name = filterExpr.path[0];
```
mtoy-googly-moogly, line 1412: This is also one of the uncomfortable places, happy to get correction here.

packages/malloy/src/lang/ast/source-elements/named-source.ts
christopherswenson: I think the thing you want to log to here is `argument`

christopherswenson: Or perhaps `argument.value`

packages/malloy/src/lang/ast/parameters/has-parameter.ts
```diff
@@ -84,6 +90,22 @@ export class HasParameter extends MalloyElement {
...
+        }
+        if (this.filterType) {
+          checkFilterExpression(this, this.filterType, constant.value);
+        }
+        return {
+          value: constant.value,
+          name: this.name,
+          type: constant.type,
+        };
+      }
```
christopherswenson, line 108: I would sort of expect, like how you can say `source: x(foo is null::string)` to also be able to say `source: x(foo is f'foo'::filter<string>)` (as well as `source: x(foo::filter<string> is f'foo')`). It doesn't seem like this allows for that?

packages/malloy/src/lang/malloy-to-ast.ts
christopherswenson: Probably better to make it a parameter with type error, so that if you have

```
source: x(foo is r'asdf') ...
run: x(foo is r'fdas') -> ...
```
You get "parameter cannot be regular expression" but not also "x has no parameter foo"

packages/malloy/src/lang/test/parameters.spec.ts
christopherswenson: Agree. Though as noted above, I would prefer if the use could choose to put the cast after the filter, as you can with null-valued defaults.

packages/malloy/src/model/malloy_query.ts
christopherswenson: I think this is correct given the way namespaces work today.... when we fix namespaces this will probably change.

packages/malloy/src/model/malloy_query.ts
```diff
@@ -1406,6 +1400,66 @@ class QueryField extends QueryNode {
...
+        break;
+      case 'number':
+        fParse = NumberFilterExpression.parse(filterSrc);
+        break;
+      case 'boolean':
+        if (context.dialect.booleanAsNumbers) {
+          throw new Error(
+            'Boolean filter expression not supported on this connection type'
+          );
+        }
```
christopherswenson, line 1442: Is this just a temporary thing where eventually we will compile boolean filter expressions as number comparisons in SQL?

mtoy-googly-moogly: Someone who cares about filter expressions could. The will to live drained out of me at this point. Boolean filter expressions, which are stupid anyway, don't work on MySQL which is stupid also. It's a that I am proposing is part of the MVP of fitlers.

packages/malloy/src/lang/malloy-to-ast.ts
mtoy-googly-moogly: The problem is that 'error' is not a legal type for creating a `HasParameter` and I didn't know if that was intentional for some reason.

packages/malloy/src/lang/ast/source-elements/named-source.ts
mtoy-googly-moogly: I was thinking I wanted the error location to be the location of the source of the filter expression.

I guess putting the actual filter expression in the error and making the location still be the entire source is ok for MVP

?

packages/malloy/src/model/malloy_query.ts
mtoy-googly-moogly: The problem is, when I read a schema, the type will be number, so even if I have thought about that value as a boolean forever, I can't apply a boolean filter to it uniess it is a local dimension to the input source I am processing. You would have to write 

```
dimension: hotOrCold is isCold::boolean ~ f'true' or isHot::boolean ~ f'true'
```

if isCold and isHot came from a table

mtoy-googly-moogly: better to just fail hard and make them write

```
dimension: hotOrCol is isHot or isCold`
```

mtoy-googly-moogly: One of the many reasons explorer should be writing stable lambdas not filter expressions

christopherswenson: I agree that the case where you are reading from the schema is problematic... But there are also cases where you define dimensions which are booleans, that you would want to use filter expressions on.

```
dimension: x is distance > 1000
dimension: filterthing is x ~ f`true`
```

Though I agree boolean filter expressions are not that useful... fine to leave it for now...

I also wonder whether someday we'll support:

```
source: s is conn.table('foo') include {
  some_integer_field::boolean
}
```
and sorta auto-cast `some_integer_field` to a boolean handwave handwave

PR #2245, opened 2025-04-02
Next iteration of custom jest matchers for executing queries

mtoy-googly-moogly:
`malloyResultMatches` has a few problems with compound data types, and the interface to "query result matching" still isn't quite right. Time to fix that maybe?


mtoy-googly-moogly:
I've taken the good bits of this and folded them into main elsewhere.

PR #2246, opened 2025-04-03
Fix reference of undefined result

whscullin:


packages/malloy-render/src/html/html_view.ts
```diff
@@ -48,7 +48,7 @@ export class HTMLView {
   ): Promise<HTMLElement> {
     let malloyResult: Malloy.Result;
     // TODO this check is bad, but I can't get VSCode to work linked without it...
-    if (result instanceof Result || 'modelDef' in result) {
+    if (result && (result instanceof Result || 'modelDef' in result)) {
```
whscullin, line 51: BTW, this is because `instanceof` is a strict equality check and the renderer bundles its own copy of Malloy so it's often going to have different implemetations.

PR #2259, opened 2025-04-11
Add 'having' to known stable query types

whscullin:
~Part one of of adding this to the query builder.~ Just pulling that thread all the way...

### DIFF
--- a/packages/malloy-query-builder/src/query-ast.ts
+++ b/packages/malloy-query-builder/src/query-ast.ts
@@ -2713,6 +2713,44 @@ export class ASTSegmentViewDefinition
     return item;
   }
 
+  public addHaving(name: string, filter: ParsedFilter): ASTHavingViewOperation;
+  public addHaving(name: string, filterString: string): ASTHavingViewOperation;
+  public addHaving(
+    name: string,
+    path: string[],
+    filter: ParsedFilter
+  ): ASTHavingViewOperation;
+  public addHaving(
+    name: string,
+    path: string[],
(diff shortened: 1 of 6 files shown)


packages/malloy-query-builder/src/query-ast.ts
```diff
@@ -3777,8 +3819,11 @@ export class ASTTimeTruncationExpression extends ASTObjectNode<
   }
 }
 
-export class ASTWhere extends ASTObjectNode<Malloy.Where, {filter: ASTFilter}> {
-  constructor(node: Malloy.Where) {
+export class ASTWhere extends ASTObjectNode<
```
christopherswenson, line 3822: Should probably now be called ASTFilterOperation

packages/malloy-query-builder/src/query-ast.ts
```diff
@@ -3797,8 +3842,11 @@ export class ASTWhere extends ASTObjectNode<Malloy.Where, {filter: ASTFilter}> {
   }
 }
 
-export class ASTWhereList extends ASTListNode<Malloy.Where, ASTWhere> {
-  constructor(wheres: Malloy.Where[]) {
+export class ASTWhereList extends ASTListNode<
```
christopherswenson, line 3845: ASTFilterOperationList

christopherswenson (approved):
Approve with some small naming changes.

Another thing to consider is whether `FilterOperation` is the right name for a filterin the stable types. @mbooth4

PR #2285, opened 2025-04-23
Update duckdb to 1.2.1.

kylenesbit:
Apparently version 1.1.1. does not have prebuilt binaries in NPM -- thus resulting in log build times.


whscullin (approved):
LGTM

PR #2306, opened 2025-04-30
`grouped_by:`

christopherswenson:
```
##! experimantal.grouped_by
```

Adds a new `grouped_by` field property which indicates that a measure is _not_ additive over listed fields, and therefore must be grouped by those fields in order to be computed. E.g., in a rollup table, you might have `daily_active_users { grouped_by: device_type }`.

This allows us to support metrics where some dimensions are not additive for that metric.

```
source: users is duckdb.table('dau_rollup') extend {
  measure: daily_active_users is _daily_active_users.sum() { grouped_by: device_type }
}
```

## "Real" Example

```
source: users is duckdb.table('dau_rollup') extend {
  measure: daily_active_users is _daily_active_users.sum() { grouped_by: device_type }
}

// this query is illegal, because `daily_active_users.sum()` does not represent 
// "daily active users" unless `device_type` is grouped
// `device_type` is not grouped, so some users could be double-counted
run: users_rollup -> {
  aggregate: daily_active_users
}

// this query is good
run: users_rollup -> {
  group_by: device_type
  aggregate: daily_active_users
}

// this is also fine
run: users_rollup -> {
  group_by: device_type
  nest: metrics is {
    aggregate: daily_active_users
  }
}

// filtering by one value is fine, but at least for now still requires a mostly-no-op group by
run: users_rollup -> {
  where: device_type = 'iPhone'
  group_by: device_type
  aggregate: daily_active_users
}
```

## Interaction with Composite Sources

If the definition of a metric in one slice (input source) of a composite source has aggregate usage which requires certain `group_by:`s _which are not satisfied_ by the query during composite field resolution, that slice will not be used. This allows you to have a different definition further down the composite input list which _does_ allow that metric to be used without that grouping.

```
source: users_cube is compose(
  users_rollup,
  duckdb.table('engagement') extend {
    measure: daily_active_users is ...
  }
)

// This query cannot use `users_rollup`, because it doesn't group by `device_type`, 
// but it can use the raw table
run: users_cube -> {
  aggregate: daily_active_users
}
```

If the metric is defined either on a table shared by all slices of the cube, or on the `compose()`d cube (rather than on each slice individually), then slices which are missing fields listed in the `grouped_by:` (because those fields were `exclude:`d from the cube) will _not_ be required.


packages/malloy/src/model/malloy_types.ts
```diff
@@ -209,6 +210,7 @@ export interface SpreadExpr extends ExprE {
 export interface FieldnameNode extends ExprLeaf {
   node: 'field';
   path: string[];
+  at?: DocumentLocation;
```
mtoy-googly-moogly, line 213: I wonder if all expr nodes should possibly have `at:` ?

mtoy-googly-moogly: ... would make ir larger by a lot

test/src/util/db-matcher-support.ts
```diff
@@ -60,7 +60,7 @@ export async function runQuery(
       fail: {
         pass: false,
         message: () =>
-          `Could not prepare query to run: ${e.message}\n\nQUERY:\n${queryText}`,
+          `Could not prepare query to run: ${e.message}\n\nQUERY:\n${queryText}\n\nSTACK:\n${e.stack}`,
```
mtoy-googly-moogly, line 63: we are fighting over what this should do, i want to look at this one

mtoy-googly-moogly: or i guess yhou could even do ..

```TypeScript
if (e.stack) { failMsg += `STACK:\n${e.stack}\n`; }
if (e.message && (e.stack === undefined || !e.stack.includes(e.message)) {
  failMesg += ...
  ```

PR #2315, opened 2025-05-05
Util function to produce Malloy.Result from QueryResult

kylenesbit:


whscullin:
This exists as `API.util.wrapResult()`

kylenesbit:
Thanks for the pointer Will!

PR #2317, opened 2025-05-06
Write basic sqlserver connection class

vitorelourenco:
This is mainly so we can discuss the decisions I had to make and how do we want to go about it.
Some major ones are the wrapping of the main select block (not fully done atm but we can talk about the approach)
`groupByClause` instead of only `orderByClause`
Support for not needing regexp
The support for not having 'LIMIT' available in the dialect
Having booleans as numbers or expressions eg 1=1 

This PR is expected to fail all sorts of tests but I feel like this is the point to start ping-ponging ideas.


packages/malloy-db-sqlserver/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
```
mtoy-googly-moogly, line 22: We are having an internal discussion about the copyright block. I HOPE our policy about this is about to be
1) Documented
2) A much smaller wall of text.

**TODO** Check in again with me before this merges.

packages/malloy/src/dialect/dialect.ts
```diff
@@ -172,6 +177,18 @@ export abstract class Dialect {
   // Like characters are escaped with ESCAPE clause
   likeEscape = true;
 
+  // Added field supportsLimit. TSQL Doesn't support limit and limit is not part of the SQL Standard
+  // TODO (vitor): Discuss with malloy
+  supportsLimit = true;
```
mtoy-googly-moogly: Something like

```TypeScript
export type LimitingClause = 'limit' | 'top';


limitClause: LimitClause = 'top';
```

?

packages/malloy/src/dialect/dialect.ts
```diff
@@ -172,6 +177,18 @@ export abstract class Dialect {
...
+
+  // Added field supportsRegexp. TSQL Doesn't support regexp. Idk how to use this with = false yet (tsql case). Malloy seems to often go for regexp
+  // TODO (vitor): Discuss with malloy
+  supportsRegexp = true;
```
mtoy-googly-moogly: There are regular expressions in Malloy expressions, but no databases "support" regular expressions as a data type. A regular expression is just a string which has special quoting rules and which therefore acts differently with the `~` operator. I don't think you need to mark them "not supported". If you can implement `sqlRegexpMatch`, which I think you can, then that is all the "support" needed.

packages/malloy/src/dialect/dialect.ts
```diff
@@ -172,6 +177,18 @@ export abstract class Dialect {
...
+
+  // TSQL doesn't let you do complex operations on GROUP BY for aliased columns.
```
mtoy-googly-moogly, line 187: I think you get to name it. If a second dialect comes along and you picked a wrong name, we'll just change it then.

packages/malloy/src/dialect/tsql/function_overrides.ts
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import type {MalloyStandardFunctionImplementations as OverrideMap} from '../functions/malloy_standard_functions';
+
+// TODO (vitor): I don't know why NULLs are greatest and least but I've seen this in postgres and duckdb. Gotta figure it out.
```
mtoy-googly-moogly, line 10: I think this is a place where @christopherswenson should comment

packages/malloy/src/dialect/tsql/tsql.ts
```diff
@@ -0,0 +1,795 @@
...
+  supportsRegexp = false;
+  supportsLateGroupByEval = false;
+  orderByClause = 'expression' as const;
+  groupByClause = 'expression' as const;
+
+  // SQL Server does have ESCAPE but Synapse doesn't
+  likeEscape = false;
+
+  // TODO (vitor): idk about this cantPartitionWindowFunctionsOnExpressions...
+  // There's a problem with non top level CTE's with sqlserver
```
mtoy-googly-moogly, line 142: This is another place where I'd like @christopherswenson to weigh in

packages/malloy/src/model/filter_compilers.ts
```diff
@@ -63,10 +63,12 @@ function unlike(disLiked: string[], x: string) {
  * XXXXFilterCompiler.compile() will move to XXXFilterExpression.compile()
  */
 
+// TODO (vitor): this is where the where clause gets built. I need to change 'true' to 1=1 and check what to do with false
 export const FilterCompilers = {
   compile(t: string, c: FilterExpression | null, x: string, d: Dialect) {
     if (c === null) {
-      return 'true';
+      // TODO (vitor): not sure if this is the best place to put it but we need something like this for tsql.
+      return d.booleanAsNumbers ? '(1 = 1)' : 'true';
```
mtoy-googly-moogly, line 71: Yeah this seems file

packages/malloy/src/model/malloy_query.ts
```diff
@@ -1359,16 +1376,27 @@ class QueryField extends QueryNode {
...
+        return `NOT COALESCE(CASE WHEN (${expr.e.sql}) THEN 1 END, 0) = 1`;
       case 'unary-':
         return `-${expr.e.sql}`;
       case 'is-null':
         return `${expr.e.sql} IS NULL`;
       case 'is-not-null':
         return `${expr.e.sql} IS NOT NULL`;
       case 'true':
+        // TODO (vitor): check with the malloy team
+        return this.parent.dialect.booleanAsNumbers ? '1=1' : expr.node;
```
mtoy-googly-moogly, line 1393: so in other "booleanAsNumber" dialects, there are symbols "true" and "false" which evaluate to "1" and "0".

if that isn't the case for tsql then probably there should be a dialect function which returns the true and false values

```TypeScript
case 'true':
  return this.parent.dialect.booleanValue(true);
case 'false':
  return this.parent.dialect.booleanValue(false);
```

And everything except tsql uses

```
   booleanValue(v: boolean) { return v ? 'true' : 'false' }
```

test/src/util/index.ts
```diff
@@ -199,17 +199,21 @@ export async function runQuery(runtime: Runtime, querySrc: string) {
...
 export function booleanResult(value: boolean, dbName: string) {
-  if (dbName === 'mysql') {
+  if (dbName === 'mysql' || dbName === 'sqlserver') {
     return value ? 1 : 0;
   } else {
     return value;
   }
 }
 
+// TODO (vitor): Not sure... This is pretty hidden away and uses dbName instead of dialect for param
```
mtoy-googly-moogly, line 211: this should go away and be replaced by the dialect function i mentioned

packages/malloy/src/dialect/dialect.ts
```diff
@@ -480,7 +486,7 @@ export abstract class Dialect {
    */
   sqlBoolean(bv: boolean): string {
     if (this.booleanType === 'none') {
-      return bv ? '(1=1)' : '(1-0)';
```
vitorelourenco, line 483: Was `1-0` instead of `1=0` intentional?

PR #2327, opened 2025-05-15
Stabilize Malloy's Thrift interface

christopherswenson:


packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -63,22 +62,7 @@ struct AnonymousQueryInfo {
     // strings well?
   6: optional string code,
```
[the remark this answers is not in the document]
christopherswenson: Having `code` would result in a lot of repeated data, but might be more convenient than just `location`. Does Thrift compress (across fields?) for this to not be an issue?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -63,22 +62,7 @@ struct AnonymousQueryInfo {
...
-  2: required Position end,
-}
-
-struct Position {
-  1: required i32 line,
-  2: required i32 character,
+  7: optional DocumentLocation location,
 }
 
 struct Schema {
```
christopherswenson, line 68: Should `DimensionInfo` and `MeasureInfo` be merged into `AtomicFieldInfo` with a `field_type` field?

christopherswenson, line 68: Should `DimensionInfo` have specific fields for "result metadata" attributes?

Summary of result metadata returned by the compiler:

* `sourceField` — I believe this is not used
* `sourceExpression` — used by drill to drill on fields which were expressions
* `sourceClasses` — only used by legacy renderer to apply renderers based on suffixes of "inherited" field names
* `filterList` — used by drill to collect filters
* `fieldKind` — used by renderer to pick default axes
* `referenceId` — used by renderer to know whether two fields in the result are the same, for hover syncing and axis sharing etc.
* `limit` — not sure where this is used
* `orderBy` — used for formatting pivot tables

Currently, _none of these_ are used, because the new API attaches _annotations_ to fields instead. These are the annotations added today:
* `#(malloy) reference_id="a47ab428-91a2-4a6c-a76f-833116c6c931"` — replaces `referenceId`
* `#(malloy) calculation` — attached to fields where `fieldKind` is "measure"; indicates that a field was a calculation
* `#(malloy) drill_filters=["x > 10", "y = 'foo'"]` — an array of strings attached to nests representing code for the filters that were applied at that nest level, for drilling (replaces `filterList`)
* `#(malloy) drill_expression="distance * 10"` — replaces `sourceExpression`
* `#(malloy) limit=100` — replaces `limit`
* `#(malloy) ordered_by=[carrier=asc, flight_count]` — replaces `orderBy`
* `#(malloy) source_name` — attached to `Result.annotations`; used in drilling; should probably be changed to `#(malloy) source { name="flights" }` to also allow `#(malloy) source { code="duckdb.table('flights')" }` and other cases where the source is not a named model entity

The drilling PR (https://github.com/malloydata/malloy/pull/2271) adds the following:
* `#(malloy) drillable` — attached to nests which support drilling, e.g. _not_ multi-stage views or index views
* `#(malloy) drill_view="by_carrier"` — attached to _fields_ to indicate which view they came from
* `#(malloy) drill_filters = [{ code="x > 10" }, { code="y = 10" field_reference="y" drill_view="by_x"  kind="literal_equality" value { kind = "number_value" number_value=10 }}, { code="z=f'A, B'" field_reference="z" drill_view="by_z" kind="filter_expression" filter_expression="A, B" }]` — allows stable filter clauses to be generated

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -277,10 +246,12 @@ struct Field {
   1: required Expression expression,
   // TODO only two kinds of distinguishable annotations are before `aggregate:` and before `name is value`
   // between `name` and `is`, or between `is` and `value` are converted to before `name`.
```
christopherswenson, line 248: Are we okay with not distinguishing between the various positions annotations on a definition can go?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -277,10 +246,12 @@ struct Field {
...
+  // TODO consider having an annotation tree with inheritance
```
christopherswenson, line 249: Do we want to have the full annotation tree, rather than just the flattened list? This tree structure is currently not used by anyone, but might be useful for people.

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -277,10 +246,12 @@ struct Field {
...
   2: optional list<Annotation> annotations,
 }
 
 struct OrderBy {
+  // TODO consider making this a specialized reference like an "OrderByReference"
```
christopherswenson, line 254: Should the generic `Reference` (with name, path, and parameters) be used a) everywhere b) most places, with some exceptions like `OrderByReference` c) never; all should be specialized.

christopherswenson: Currently, everything uses `Reference` even if a) they only allow a name and not a path, or b) they don't support parameters.

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -277,10 +246,12 @@ struct Field {
...
   1: required Reference field_reference,
   2: optional OrderByDirection direction,
 }
```
christopherswenson, line 257: Re: https://github.com/malloydata/malloy/pull/2327/files#diff-9d0111becc433e2a89f480d46840047e845f6816c95d1b0ecd573ca173e5827eR263-R266

Currently, a couple trivial wrapper structs exist to wrap around a `union` so that it can use used in a different `union`. This is because the current TS types flatten unions. Is this okay, or should we find a way to make the TS types work with union branches being typed as unions somehow?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -348,28 +300,29 @@ struct QueryArrow {
...
 
 struct ViewArrow {
   1: required ViewDefinition source,
   2: required ViewDefinition view,
 }
 
-struct ViewSegment {
-  1: required list<ViewOperation> operations,
+// TODO: I'm not conviced about just calling this "block"
```
christopherswenson, line 323: Is there a better name to describe a "view block" / "view segment". Internally, we call this a "query operation".

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -429,7 +382,9 @@ union Expression {
 }
 
 struct TimeTruncationFieldReference {
-  1: required Reference field_reference, // TODO do I make this circular, more like actual grammar? e.g. TimeTruncation rather than TimeTruncationFieldReference
+  // TODO do I make this circular, referring back to `Expression`, more like actual grammar?
+  // e.g. TimeTruncation rather than TimeTruncationFieldReference
+  1: required Reference field_reference,
```
christopherswenson, line 387: Should `TimeTruncationFieldReference` be a `TimeTruncation` instead, having a `to_truncate: Expression` field, which is more similar to the actual grammar? Or is it okay to have this more specialized thing?

Pros:
* matches the grammar more closely
* if/when we add full (or just more) expressions to this interface, we would have to keep both `TimeTruncationFieldReference` and `TimeTruncation`

Cons: 
* makes it harder _now_, when we don't support anything other than a reference on the LHS of a time truncation, to typecheck

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -453,11 +408,13 @@ struct NumberCell {
 struct NullCell {}
 
 struct TimestampCell {
-  1: required string timestamp_value, // TODO another way to represent dates?
+  // TODO another way to represent dates?
+  1: required string timestamp_value,
```
christopherswenson, line 412: Is there a better way to represent dates in Thrift than ISO strings? Epochs?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -453,11 +408,13 @@ struct NumberCell {
...
 }
 
 struct DateCell {
-  1: required string date_value, // TODO another way to represent dates?
+  // TODO another way to represent dates?
+  1: required string date_value,
 }
 
 struct JSONCell {
```
christopherswenson, line 420: Re: https://github.com/malloydata/malloy/pull/2327/files#diff-9d0111becc433e2a89f480d46840047e845f6816c95d1b0ecd573ca173e5827eR441

Do `TimestampCell`s need to be different from `DateCell`s?

christopherswenson, line 420: Re: https://github.com/malloydata/malloy/pull/2327/files#diff-9d0111becc433e2a89f480d46840047e845f6816c95d1b0ecd573ca173e5827eR443

Is there a need for a `JSONCell`

christopherswenson, line 420: Currently there is a single `Result` struct that is used for both a "result with data" and a "result with just SQL". Should these be separated? (They are the same now because the _only_ difference between the two is the presence of the `data` field)

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -5,54 +5,53 @@
...
+// may have a `definition`, which indicates that it is a query that is representable
+// as a `Query`. Not all queries are representable as such, because this interface
+// has more limited capabilities than the full Malloy language (e.g. index queries
+// are not representable here).
+
 struct ModelInfo {
-  1: required list<ModelEntryValue> entries,
-  2: optional list<Annotation> annotations,
-  3: required list<AnonymousQueryInfo> anonymous_queries,
+  1: required list<ModelEntry> entries;
```
christopherswenson, line 19: `required` is deprecated, but is needed to generate typescript types which accurately reflect which fields are _supposed to be present_. @narreola Is there a way you can think of for us to annotate fields as required such that we can generate TS types properly, but without actually making them required in the thrift?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -610,88 +524,93 @@ struct LogMessage {
...
 struct CompileModelResponse {
-  1: optional ModelInfo model,
+  1: optional ModelInfo model;
 
-  8: optional list<LogMessage> logs,
-  9: optional CompilerNeeds compiler_needs,
+  8: optional list<LogMessage> logs;
+  9: optional CompilerNeeds compiler_needs;
   10: optional list<Translation> translations;
 }
```
christopherswenson, line 543: @narreola suggests that there should be different types for the different versions of the API (stateless, sessioned, and asynchronous). 

Should this look like

```thrift
struct CompileModelResponse {
  1: optional ModelInfo;
  2: optional list<LogMessage> logs;
}
struct StatelessCompileModelResponse {
  1: required CompileModelResponse response;
  2: optional CompilerNeeds compiler_needs;
}
struct SessionedCompileModelResponse {
  1: required CompileModelResponse response;
  2: optional CompilerNeeds compiler_needs;
  3: optional string session_id;
}
```

or

```thrift
struct StatelessCompileModelResponse {
  1: optional ModelInfo;
  2: optional list<LogMessage> logs;
  3: required CompileModelResponse response;
  4: optional CompilerNeeds compiler_needs;
}
struct SessionedCompileModelResponse {
  1: optional ModelInfo;
  2: optional list<LogMessage> logs;
  3: required CompileModelResponse response;
  4: optional CompilerNeeds compiler_needs;
  5: optional string session_id;
}
```

or something else?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -181,409 +162,342 @@ struct FilterExpressionType {
...
-  7: required FilterOperation having,
+union ViewStatement {
+  1: required GroupBy group_by;
+  2: required Aggregate aggregate;
+  3: required OrderBy order_by;
+  4: required Limit limit;
+  5: required FilterOperation where;
+  6: required Nest nest;
+  7: required FilterOperation having;
 }
```
christopherswenson, line 228: `DrillOperation` is theoretically going to be added, too. https://github.com/malloydata/malloy/pull/2271

christopherswenson: We should discuss whether we need a framework for "experimental" features in the "stable" interface?

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -181,409 +162,342 @@ struct FilterExpressionType {
...
+  1: required list<ViewStatement> operations;
 }
 
 struct Reference {
-  1: required string name,
-  2: optional list<string> path,
-  3: optional list<ParameterValue> parameters,
+  1: required string name;
+  2: optional list<string> path;
+  3: optional list<ParameterValue> parameters;
```
christopherswenson, line 331: This should probably be called `arguments`

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -181,409 +162,342 @@ struct FilterExpressionType {
...
 }
 
 struct ParameterValue {
-  1: required string name,
-  2: required LiteralValue value,
+  1: required string name;
+  2: required LiteralValue value;
```
christopherswenson, line 336: Parameters can technically be expressions in the language; someday this might become relevant in the interface, so we may want to enable that...

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: MToy votes that DimensionInfo and MeasureInfo is good, due to convenience and that the number of different types is small.

christopherswenson: MToy and I agree that it is nice that the tags are easily changeable without having to mess with the thrift. In the future, we might want to move these into the thrift interface, or might want to make different things accessible with different API settings.

The possible danger is that this allows Malloy developers to "secretly extend the thrift" interface... might be scary...

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: Michael would like to know whether the annotation was from the group or object (or inheritance), but doesn't care whether it was before the is or after the is.

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: Michael's vote is to keep the tree; it's easy to flatten it as a user.

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: Michael's vote (not super strong vote) is that there is a separate reference type.

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: MToy doesn't love any of these terms, but is okay with "Block". MToy would like us to pick a name and use it everywhere.

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: MToy thinks ISO makes sense, because everyone knows what an ISO means (particularly WRT timezones)

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: Mtoy: consider whether it is awkward to use if they're separate? maybe a utility function?

christopherswenson: MToy thinks there might be something to consider with `-> { select: * }` on a nested result.

christopherswenson: MToy likes something like this better (but doesn't feel too strongly):

```
struct CompileResult {
  2: required Schema schema,
  3: optional string sql,
  4: required string connection_name,
  5: optional list<Annotation> annotations,
  6: optional list<Annotation> model_annotations,
  7: optional string query_timezone,
  8: optional list<Annotation> source_annotations,
}

struct RunResult {
  1: required Data data,
  2: required CompileResult compile_result,
}
```

packages/malloy-interfaces/thrift/malloy.thrift
christopherswenson: For drill, MToy is happy to have it not be experimental. Discuss later about experimental features.

packages/malloy-interfaces/thrift/malloy.thrift
```diff
@@ -5,54 +5,53 @@
...
+  6: optional DocumentLocation location;
 }
 
 struct AnonymousQueryInfo {
-  2: required Schema schema,
-  4: optional list<Annotation> annotations,
-  // "openable query"
-  5: optional Query definition,
+  2: required Schema schema;
+  4: optional list<Annotation> annotations;
```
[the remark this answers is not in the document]
christopherswenson: Worth getting more info from Fernando about what he would use this for.

PR #2353, opened 2025-05-28
Treat null values as undefined in thrift unions

mbooth4:
I am trying to convert a thrift interface, but the Thrift object has 'null' as the value for all the non-active items in a union, like this:

```
array_type: null,
boolean_type: null,
date_type: null,
json_type: null,
number_type: {subtype: 1},
record_type: null,
sql_native_type: null,
string_type: null,
timestamp_type: null,
```

I have not exhaustively thought through the implications of this, so wanted to run it by the team before digging too deep.


mbooth4:
Since we don't seem to have a better way to accomplish it, I will go ahead and move forward with this change in the utility function.

PR #2374, opened 2025-06-03
Experimental implementation of an "exploreInfo" api

mtoy-googly-moogly:
```TypeScript

// from malloy.ts

export type SourceComponentInfo =
  | {type: 'table'; tableName: string; sourceID: string}
  | {type: 'sql'; selectStatement: string; sourceID: string};

// in use

import {SourceComponentInfo} from '@malloydata/malloy'

/*
 * Look for all sources in the model tagged with
 *   #(look_for_me) enable
 */
const theModel = runtime.loadModel('/model/url/name.malloy')
for (const src of theModel.explores) {
  const specialTags = src.tagParse({ prefix: /^#\(look_for_me\) /}).tag;
  if (specialTags.has("enable")) {
    const srcComponents= = src.getComponents();
    ... THIS 
  }
}
```


packages/malloy/src/exploreinfo.ts
```diff
@@ -0,0 +1,204 @@
...
+  } else if (sourceDef.type === 'sql_select') {
+    // Generate sourceID based on connection and SQL statement
+    const sourceID = `${sourceDef.connection}:${sourceDef.selectStr}`;
+
+    return {
+      type: 'sql',
+      selectStatement: sourceDef.selectStr,
+      sourceID: sourceID,
+    };
+  } else if (sourceDef.type === 'query_source') {
```
areding99, line 156: I thought we were going to handle a 'query_source' by walking to its underlying sql_ or table_source? That said, this could be better..

mtoy-googly-moogly: This might be better, we'll see

areding99:
lgtm - thanks Michael!

PR #2406, opened 2025-06-25
Throw an error if a select view is nested

mbooth4:
Disable the ability to nest a `select:` query. Lloyd explained that there are issues with allowing this which go beyond the script errors users hit in certain cases, such as issues with fan out.

Related to https://github.com/malloydata/malloy/pull/2368 which was an attempt to make this case work more consistently. Discussion during that PR's development led us to conclude that the best option is to prevent users from trying to nest a select.


whscullin:
Looks like this got caught in the jest refactor, merging in main should fix most of these errors.

PR #2430, opened 2025-07-07
Fix join (and time truncation) field references in drillable queries

christopherswenson:
This PR fixes a couple issues with drilling; namely that a query containing field references to joined fields, or time truncated fields could not drill properly (`run: flights -> { group_by: carriers.nickname, dep_time.month }`).

Original drilling feature just didn't take into account these cases, and the implementation to make these work is somewhat complex.

The problems:
* Compiler was not attaching enough information to the metadata tags to allow the renderer to know how to produce a drill on a joined field
* The stable API did not even have a way to represent a filter on a time truncated field

Solution:
* Move some of the logic for drill filter expression computation out of the renderer and into the compiler:
  * Compiler now adds `#(malloy) drill_expression { ... }` tags to indicate exactly what expression needs to be on the LHS of the filter
    * **Joins**: `#(malloy) drill_expression { kind = field_expression name = nickname path = [ carriers ] }`
    * **Time truncation**: `#(malloy) drill_expression { kind = time_trunctation field_reference.name = dep_month truncation = month }`
    * **"True drill" references**: `#(malloy) drill_expression { kind = field_expression name = carrier path = [ top_carriers ] }`
  * Query fields generated by the translator can now have a `drillExpression` which is the stable API `Expression` which represents the definition of that field; this is what enables the compiler to generate valid drill expressions for time truncations.
* Change the filter types in the stable API to have an expression (which might be a field reference expression) rather than a field reference.


christopherswenson:
Planning to move some of the logic from the compiler to the translator; will update soon.

PR #2432, opened 2025-07-09
Fix drilling context

whscullin:
make sure context is retrieved during rendering.


skokenes:
my bad I missed this, already merged in another PR

PR #2439, opened 2025-07-15
Draft: Add a Calculate View Operation and MovingAverageExpression

mbooth4:
In order to support smoothing in Malloy Explorer, we need a way to express the appropriate calculation in the Stable Query.

This is a draft because there are a few more tasks to complete:

**TO DO**
- The Calculate Expression will need a `partition_by` list added to it
- Probably need a function to convert a Calculate back into an Aggregate by removing the moving average

### DIFF
--- a/packages/malloy-interfaces/src/to_malloy.ts
+++ b/packages/malloy-interfaces/src/to_malloy.ts
@@ -318,6 +318,8 @@ function groupedOperationsToFragments(
       return havingToFragments(operations as Malloy.FilterOperation[]);
     case 'drill':
       return drillToFragments(operations as Malloy.DrillOperation[]);
+    case 'calculate':
+      return calculateToFragments(operations as Malloy.CalculateOperation[]);
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/malloy-interfaces/src/to_malloy.ts
```diff
@@ -505,6 +524,24 @@ function havingToFragments(having: Malloy.FilterOperation[]): Fragment[] {
...
+      fragments.push(NEWLINE);
+    }
+    if (operation.name) {
+      fragments.push(maybeQuoteIdentifier(operation.name));
+      fragments.push(' is ');
+    }
+    fragments.push(...fieldToFragments(operation.field));
+  }
+  return formatBlock('calculate', [fragments]);
+}
```
christopherswenson, line 543: I would try to make `aggregateToFragments`, `groupByToFragments`, and `calculateToFragments` share code at this point, because they're all basically the same (I see some minor differences but they may be able to be ignored, not sure).

packages/malloy-interfaces/src/to_malloy.ts
```diff
@@ -379,6 +381,23 @@ function expressionToFragments(expression: Malloy.Expression): Fragment[] {
       ];
     case 'literal_value':
       return literalToFragments(expression.literal_value);
+    case 'moving_average':
+      return [
+        'avg_moving',
+        ...wrap(
```
christopherswenson, line 387: Pass `{ spaces: false }` here so it comes out as `avg_moving(x, 1)` rather than `avg_moving( x, 1 )`.

christopherswenson (approved):
Looks great so far

packages/malloy-interfaces/src/types.ts
```diff
@@ -147,6 +147,22 @@ export const MALLOY_INTERFACE_TYPES: Record<string, MalloyInterfaceType> = {
     'name': 'BooleanType',
     'fields': {},
   },
+  'CalculateOperation': {
```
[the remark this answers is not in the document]
mbooth4: Good idea. Maybe let's do this in a different PR.

christopherswenson (approved):
LGTM

PR #2447, opened 2025-07-21
Resolve query output structs in the translator

christopherswenson:
The compiler is currently responsible for generating intermediate `StructDef`s for queries — that is, every time you write a segment `{ group_by: whatever }`, the translator calls into the compiler to decide what the schema of the output of that segment should be. This is pretty silly, since the translator has all the requisite information to do that itself. This PR *finally* makes the translator smart enough to do this itself.


packages/malloy/src/lang/ast/field-space/column-space-field.ts
```diff
@@ -34,6 +39,10 @@ export class ColumnSpaceField<T extends AtomicFieldDef> extends SpaceField {
     return this.sourceFieldDef;
   }
 
+  getQueryFieldDef(_fs: FieldSpace): QueryFieldDef | undefined {
+    return this.sourceFieldDef;
+  }
+
```
christopherswenson, line 45: TODO: figure out why I can't undo this change without breaking the world

PR #2455, opened 2025-07-23
fix: Optimize field creation performance in malloy-render

skokenes:
## Summary

This PR addresses a significant performance issue in the malloy-render package where excessive field creation and tag parsing were causing performance degradation, especially with large nested datasets.

## Problem

For a simple two-column Malloy query, we observed:
- **5 Field objects created** instead of the expected 2 (2.5x overhead)
- **7-8 tagFor calls** instead of the expected 2 (3.5-4x overhead)

This overhead scaled dramatically with nested data structures, potentially creating hundreds of thousands of unnecessary objects and expensive tag parsing operations.

## Root Cause

The RepeatedRecordField class was creating fields multiple times:
1. Once via its parent ArrayField's `elementField`
2. Again in its own constructor
3. Additional overhead from creating a synthetic `nestedRecordField`

## Solution

Implemented three key optimizations:

### 1. Lazy ElementField Creation
- Made `elementField` in ArrayField lazy-loaded
- Prevents unnecessary field creation for RepeatedRecordField instances

### 2. Field Sharing
- RepeatedRecordField now creates fields once and shares them between:
  - The main `fields` array
  - The synthetic `nestedRecordField`
  - The inherited `elementField` (via override)

### 3. Skip Redundant Processing
- Added `skipFieldCreation` parameter to prevent duplicate field creation
- Added `skipTagParsing` parameter to skip expensive tag parsing for synthetic fields

## Results

- **Field.from calls reduced from 5 to 2** (60% reduction)
- **tagFor calls reduced from 7-8 to 3-4** (50%+ reduction)

For complex nested structures, this optimization eliminates thousands of unnecessary operations, significantly improving render performance.

## Testing

- All existing tests pass
- Performance improvements verified with trace analysis
- No breaking changes to public APIs

## Documentation

Added comprehensive inline documentation explaining:
- Why each optimization exists
- When to use the new parameters
- How the field sharing mechanism works

This is a performance-critical fix that maintains backward compatibility while providing substantial performance improvements for data visualization.

### DIFF
--- a/packages/malloy-render/src/data_tree/fields/nest.ts
+++ b/packages/malloy-render/src/data_tree/fields/nest.ts
@@ -10,11 +10,32 @@ import type {
 
 export class ArrayField extends FieldBase {
   public readonly maxUniqueFieldValueCounts: Map<string, number> = new Map();
+  protected _elementField?: Field;
+
   constructor(
     public readonly field: ArrayFieldInfo,
-    parent: Field | undefined
+    parent: Field | undefined,
+    skipTagParsing = false
   ) {
-    super(field, parent);
+    super(field, parent, skipTagParsing);
+  }
+
+  /**
+   * Lazy getter for elementField to optimize field creation.
+   *
+   * Performance optimization: For RepeatedRecordField, this getter is overridden
+   * to return the nestedRecordField, preventing duplicate field creation.
+   * For regular ArrayField instances, the element field is created on-demand.
+   */
+  get elementField(): Field {
+    if (!this._elementField) {
+      const elementFieldInfo = {
+        name: 'element',
+        type: this.field.type.element_type,
+      };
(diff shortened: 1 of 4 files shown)


packages/malloy-render/src/data_tree/fields/nest.ts
```diff
@@ -41,6 +64,44 @@ export class RepeatedRecordField extends ArrayField {
...
+    const recordFieldInfo: RecordFieldInfo = {
+      name: 'record',
+      type: this.field.type.element_type,
+    };
+
+    const recordField = new RecordField(recordFieldInfo, this, true, true);
+
+    // Share our already-created fields to avoid duplication
+    recordField.fields = this.fields;
+    recordField.fieldsByName = this.fieldsByName;
```
whscullin, line 88: Why not pass these into the constructor and avoid the boolean trap?

PR #2471, opened 2025-08-08
Experimental partition composites

christopherswenson:
One of the primary uses for composite sources is to handle partition selection of cubes stored as partition tables. This results in a pattern like:

```malloy
##! experimental.composite_sources
source: partition_a is connection.table('cube_table') include { a } extend {
  where: partition = 'a'
}
source: partition_b is connection.table('cube_table') include { b } extend {
  where: partition = 'b'
}
source: partition_c is connection.table('cube_table') include { c } extend {
  where: partition = 'c'
}
source: cube is compose(partition_a, partition_b, partition_c)
```

For large cubes, this can result in significantly bloated models, since each partition gets its own source definition complete with its own field list; then based on the (not yet optimized) way composite sources are stored in model IR, each of those sources' IR is duplicated in the composite source's IR. 

Larger models lead to significant cache speed degradation, both in terms of fetching the pre-compiled model from the cache as well as deserializing the model to load into the translator. 

Since this is such a common pattern, we are experimenting with a more IR-space-efficient feature for handling this type of case. 

```malloy
#! experimental { partition_composite { partition_field=partition partitions={a = {a} b = {b} c = {c}} } }
source: cube is connection.table('cube_table')
```

**Important Note: because this first version is using annotations, the partition specification applies to the whole `source:` definition, including extensions. If you want to extend a cube table without making the new fields part of the composite selection, you must do so in a separate `source:` definition.**

If the experiment proves valuable, we should design and implement actual Malloy syntax for this feature rather than relying on compiler annotations. This may be as simple as:

```malloy
##! experimental.partition_composites
source: cube is connection.table('cube_table') extend {
  partition_composite: {
    partition_by: partition
    partition: a is [a]
    partition: b is [b]
    partition: b is [b]
  }
}
```

### DIFF
--- a/packages/malloy/src/lang/test/composite-field-usage.spec.ts
+++ b/packages/malloy/src/lang/test/composite-field-usage.spec.ts
@@ -541,13 +541,13 @@ describe('composite sources', () => {
   describe('partition composites', () => {
     test('basic okay', () => {
       expect(`
-        #! experimental { partition_composite { partition_field=astr partitions=[{id=ai fields=[ai]}] } }
+        #! experimental { partition_composite { partition_field=astr partitions={ai={ai}} } }
         source: a_partition is a
       `).toTranslate();
     });
     test('can use field from extension not mentioned in partition spec', () => {
       expect(`
-        #! experimental { partition_composite { partition_field=astr partitions=[{id=ai fields=[ai]}] } }
+        #! experimental { partition_composite { partition_field=astr partitions={ai={ai}} } }
         source: a_partition is a
 
         source: a_partition_ext is a_partition extend {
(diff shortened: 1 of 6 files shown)


packages/malloy/src/lang/test/composite-field-usage.spec.ts
```diff
@@ -538,4 +538,81 @@ describe('composite sources', () => {
       `).toTranslate();
     });
   });
+  describe('partition composites', () => {
+    test('basic okay', () => {
+      expect(`
+        #! experimental { partition_composite { partition_field=astr partitions=[{id=ai fields=[ai]}] } }
```
nachoarreola, line 544: I think it should be no problem since its using annotations, but can we add a couple tests where the field names have spaces/special characters pls?

nachoarreola (approved):

PR #2484, opened 2025-08-22
Fix "@malloydata/render" types path

franco-roura:
This PR fixes a bug where `@malloydata/render` types couldn't be resolved.

## Before
image

## After
image


whscullin:
Thank you for this fix!

PR #2508, opened 2025-09-14
fix: Top level run statement Annotation parsing

aszenz:
Fixes the function malloyToQuery, originally created in https://github.com/malloydata/malloy/pull/2181

### DIFF
--- a/packages/malloy/src/lang/malloy-to-stable-query.ts
+++ b/packages/malloy/src/lang/malloy-to-stable-query.ts
@@ -105,7 +105,12 @@ export class MalloyToQuery
     cx: HasAnnotations
   ): Malloy.Annotation[] | undefined {
     const annotations = cx.ANNOTATION().map(a => {
-      return {value: a.text};
+      // Strip trailing newline/EOF from annotation text but keep the '#' prefix
+      let value = a.text;
+      if (value.endsWith('\n')) {
+        value = value.slice(0, -1);
+      }
+      return {value};
     });
     return annotations.length > 0 ? annotations : undefined;
   }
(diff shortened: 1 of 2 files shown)


mtoy-googly-moogly (changes requested):
Annotations, by definition, contain the full text, including the trailing new line, fix that and the test and I'll merge it, thanks for a good fix and test!

aszenz:
> Annotations, by definition, contain the full text, including the trailing new line, fix that and the test and I'll merge it, thanks for a good fix and test!
> 
> ... or i have a branch with that change ... and your test which i made just to check out the change, it's trivial for me to hit the merge button on that, also happy to let you update this pr and merge
> 
> #2509

Thanks updated the PR now, I was wondering how this bug hasn't affected other uses

mtoy-googly-moogly (approved):
Thanks, LGTM. I can't merge this right now, because it's failing tests. We are still working out details of how to get changes into the tree.

I'm checking with @whscullin on process, we will get this landed.

aszenz:
> you might be the first to want annotations on a run statement

Well this query comes from malloy explorer, it's used in my app to translate malloy explorer's `Query` to URL Search params  and then back to `Query` from it.

> If you are in a hurry to get this change in, the fastest path is to merge my copy of this branch.
> 
> If you are not in a hurry, I am very happy (since I need to figure this all out eventually anyway) to move this through the CI process.

No rush i already fixed it in my app manually using regex parsing.

mtoy-googly-moogly:
1) the Linux foundation requires us to have contributors pass a DCO check. i am not 100% crystal clear on all the mechanisms, but for myself, I had to set up my git configuration so that my commits were signed. you can also make a DCO remediation commit to this PR without setting up signing, but then the next commit you make will have the same problem.

2) There is some problem where the CI is failing possibly because it is running against a very old branch. THe mechanism for handling the state of the CI checks in main vs the state of the CI checks in a PR is hard to untangle and currently the ci scripts run by the merging of your changes and the main branch are not working.

We have been trying to make submissions easier, but this case, where the branch and the main from the fork are older than than the current version of the ci checks is one we didn't test, and it doesn't work.

I don't have more time today to spend on this, it isn't anything you've done wrong. Tomorrow I will start staring at things to see if I can untangle all of this.

PR #2525, opened 2025-09-26
write nested joins as 1=1 with where

mtoy-googly-moogly:
ok there are four pieces to this

1) mark field usage from an on expression as coming from an on expression
2) when expanding field usage, propogate the on expression to all fields needed to satisfy the on expression
3) when building the active join list, look for joins with on expressions which go through another join and mark that
4) when generating joins, for marked joins, move the on expression to the filters and make the on expression be 1=1

the approach was suggested by lloyd
the hard parts of this were written by and and claude


mtoy-googly-moogly:
BAdd approach, abandoning

PR #2535, opened 2025-10-13
Fix coalesce across sources

cata02:
**Problem**
Coalesce expressions (??) didn’t merge field usage from both operands when referencing joined sources. This led to missing joins and SQL errors.

**Root Cause**
In ExprCoalesce.getExpression(), the fieldUsage property wasn’t merged from both operands. The query planner didn’t include all required joins.

Example query:

```
source: flights is duckdb.table('malloytest.flights') extend {
  join_one: airports on airports.key = origin
  join_one: carriers on carriers.code = carrier
}

run: flights -> {
  select:
    result1 is flight_id ?? airports.city ?? carriers.name
}
```
For this malloy query, the generated SQL was missing the LEFT joins to airports and to carriers leading to broken SQL syntax; the COALESCE syntax itself was correct.

With the fix on this PR, the following SQL code is now generated:

```
SELECT COALESCE((COALESCE((base."id2"),airports_0."city")),carriers_0."name") as result1
FROM malloytest.flights base
LEFT JOIN malloytest.airports airports_0 ON airports_0."code" = base."origin"
LEFT JOIN malloytest.carriers carriers_0 ON carriers_0."code" = base."carrier"
```
Added new tests to cover this scenario and param ?? constant and param ?? field

Signed-off-by: Catalin Adler <catalin.adler@bird.com>

### DIFF
--- a/packages/malloy/src/lang/test/composite-field-usage.spec.ts
+++ b/packages/malloy/src/lang/test/composite-field-usage.spec.ts
@@ -1295,4 +1295,27 @@ describe('field usage with compiler extensions', () => {
       }
     }
   });
+
+  test('coalesce across joined sources', () => {
+    const query = model`
+      source: c is a
+      source: abc is a extend {
+        join_one: c with astr
+        join_one: b with astr
+      }
+      run: abc -> {
+        select:
+          result1 is null ?? af ?? b.af ?? c.af
+      }
+    `;
+    expect(query).toTranslate();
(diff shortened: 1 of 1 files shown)


mtoy-googly-moogly (changes requested):
Great catch. The tests for this should be in packages/malloy/src/lang/test/composite-field-usage.spec.ts not in the api package though.

PR #2582, opened 2025-12-01
Add GizmoSQL support to @malloydata/db-duckdb

sjruhland:
## Summary

Adds GizmoSQL connection support to the `@malloydata/db-duckdb` package, following the established pattern for DuckDB variants (similar to MotherDuck and WASM).

**Key changes:**
- `GizmoSQLConnection` class extending `DuckDBCommon`
- Python ADBC bridge for Arrow Flight SQL protocol
- SQL injection protection for catalog identifiers
- Efficient data transfer via Apache Arrow IPC format

## Architecture

GizmoSQL is DuckDB served over Arrow Flight SQL with the same SQL dialect. Following the project's convention, this implementation belongs in `db-duckdb` alongside other DuckDB transport mechanisms:

```
@malloydata/db-duckdb/
├── DuckDBConnection       (native bindings)
├── DuckDBWasmConnection  (WASM)
└── GizmoSQLConnection    (Flight SQL)
```

## Implementation Details

**Connection Layer:**
- Extends `DuckDBCommon` for shared dialect and schema introspection
- Python subprocess bridge handles ADBC Flight SQL communication
- Apache Arrow for efficient columnar data transfer

**Security:**
- Catalog identifier validation prevents SQL injection
- Credentials managed via secure keystore in VSCode extension

**Dependencies:**
- Python: `pyarrow`, `adbc-driver-flightsql`
- Node: `apache-arrow` (already a dependency)

## Testing

Connection tested successfully with:
- Basic connectivity
- Query execution
- Schema introspection
- Error handling

### DIFF
--- a/packages/malloy-db-duckdb/src/gizmosql_connection.spec.ts
+++ b/packages/malloy-db-duckdb/src/gizmosql_connection.spec.ts
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files
+ * (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge,
+ * publish, distribute, sublicense, and/or sell copies of the Software,
+ * and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import {describeIfDatabaseAvailable} from '@malloydata/malloy/test';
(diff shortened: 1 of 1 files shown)


packages/malloy-db-duckdb/src/gizmosql_connection.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  public async *runSQLStream(
+    sql: string,
+    {rowLimit, abortSignal}: RunSQLOptions = {}
+  ): AsyncIterableIterator<QueryDataRow> {
+    const defaultOptions = this.readQueryOptions();
+    rowLimit ??= defaultOptions.rowLimit;
+    await this.setup();
+
+    const statements = sql.split('-- hack: split on this');
```
sagarswamirao: Can you remove the `-- hack: split on this` delimiter. I believe this isn’t a valid or stable way to split SQL statements and GizmoSQL doesn’t define or rely on this marker anywhere

Comment based splitting is risky

packages/malloy-db-duckdb/src/gizmosql_connection.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+const [describe] = describeIfDatabaseAvailable(['gizmosql']);
+
+describe('GizmoSQLConnection', () => {
+  let connection: GizmoSQLConnection;
+
+  beforeAll(async () => {
+    const uri = process.env['GIZMOSQL_URI'];
+    const username = process.env['GIZMOSQL_USERNAME'];
+    const password = process.env['GIZMOSQL_PASSWORD'];
+    const catalog = process.env['GIZMOSQL_CATALOG'] || 'main';
```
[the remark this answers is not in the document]
sagarswamirao: I created an account and was able to run the follwoing query: `SELECT * FROM customer LIMIT 10;`

Here's what I found in the Network Tab:
{
    "action": "query",
    "sql": "SELECT * FROM customer LIMIT 10;",
    "parameters": []
}

I didn't have any catalog specified. I believe catalog is optional

Here's another query:
{
    "action": "query",
    "sql": "\n                SELECT column_name, data_type, is_nullable\n                FROM information_schema.columns\n                WHERE table_schema = ?\n                AND table_name = ?\n                ORDER BY ordinal_position\n            ",
    "parameters": [
        "main",
        "nation"
    ]
}

packages/malloy-db-duckdb/src/gizmosql_connection.ts
sjruhland: Can do but fyi this was borrowed from `duckdb_*.ts` implementations

sagarswamirao: Ah I see. Then it's fine :)

Issue #2601, opened 2025-12-23
`maybeQuoteIdentifier` incorrectly splits identifiers containing dots


AnatolyRugalev:
### What happens?

The `maybeQuoteIdentifier` function in `malloy-interfaces/src/util.ts` incorrectly handles identifiers that contain dots. When an identifier like `foo-bar.baz` needs quoting, the function splits it by `.` and quotes each segment individually, producing `` `foo-bar`.baz `` instead of the correct `` `foo-bar.baz` ``.

This causes parsing errors when converting a query AST back to Malloy text via `queryToMalloy()`, as the parser interprets `` `foo-bar`.baz `` as a path expression rather than a single quoted identifier.

### To Reproduce

1. Create a source with a name containing special characters and dots (e.g., `foo-bar.baz`)
2. Build a query AST referencing this source
3. Convert the AST back to Malloy text using `queryToMalloy()`
4. Attempt to parse the resulting Malloy text

The parsing fails with:

```
no viable alternative at input '`foo-bar`.baz'
```

**Example:**
```typescript
const query: Query = {
  definition: {
    kind: 'arrow',
    source: {kind: 'source_reference', name: 'foo-bar.baz'},
    view: {...}
  }
};
const malloyText = queryToMalloy(query);
// Produces: `foo-bar`.baz -> { ... }
// Should produce: `foo-bar.baz` -> { ... }
```

### OS:

macOS

### Malloy Client:

TypeScript/Node.js

### Malloy Client Version:

v0.0.309 (also verified bug exists in main branch)

### Database Connection:

N/A (issue is in query serialization, not database interaction)

PR #2602, opened 2025-12-23
fix: maybeQuoteIdentifier incorrectly splits identifiers containing dots

AnatolyRugalev:
The function was splitting identifiers by '.' and quoting each segment individually, producing `foo-bar`.baz instead of `foo-bar.baz`.

This caused parsing errors when converting a query AST back to Malloy text via queryToMalloy(), as the parser interprets the split version as a path expression rather than a single quoted identifier.

Fixes https://github.com/malloydata/malloy/issues/2601


mtoy-googly-moogly (approved):
Good catch
Good Fix
Good Test
Thank you, running CI and if it clears I will merge.

PR #2604, opened 2025-12-23
Feature/duckdb node neo

jswir:
## Summary

Migrate `@malloydata/db-duckdb` from the legacy `duckdb` package to `@duckdb/node-api` (Node Neo).

### Why

The legacy `duckdb` package has issues on Windows with MotherDuck connections (crashes on query execution). The new `@duckdb/node-api` package:
- Has better cross-platform support
- Uses a modern Promise-based API
- Is the officially recommended Node.js API going forward

### Changes

**Dependencies:**
- Remove `duckdb: ^1.3.4`
- Add `@duckdb/node-api: ^1.4.3-r.1`

**API migration in `duckdb_connection.ts`:**
- `new Database()` → `DuckDBInstance.create()`
- `database.connect()` → `instance.connect()`
- `connection.all()` → `connection.run()` + `getRowObjectsJson()`
- `connection.stream()` → `connection.stream()` + `yieldRowObjectJson()`

## Note

VS Code extension build system updates coming in a separate PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4054,9 +4054,9 @@
       }
     },
     "node_modules/@duckdb/duckdb-wasm": {
-      "version": "1.29.1-dev132.0",
-      "resolved": "https://registry.npmjs.org/@duckdb/duckdb-wasm/-/duckdb-wasm-1.29.1-dev132.0.tgz",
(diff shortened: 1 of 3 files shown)


mtoy-googly-moogly (changes requested):
It turns out the changes we made to pass the compound atomic tests require that we update duckdb-wasm to a more recent version which parses the same SQL as the duckdb node neo package

[duckdb-wasm-upgrade.patch](https://github.com/user-attachments/files/24399853/duckdb-wasm-upgrade.patch)

PR #2720, opened 2026-03-12
feat(render): Dashboard UX improvements — 12-column grid, smart defaults, card styling

jswir:
## Summary

- **Dual layout modes**: Dashboards without explicit `# span`, `# dashboard { columns }`, or `# dashboard { gap }` tags use a **flex layout** where items size naturally based on content. When any of those tags are present, the dashboard switches to a **12-column CSS grid** with smart default spans (measures → span 3, nests → span 4-12 based on weighted child content).
- **Card styling**: All dashboard items (KPIs, charts, tables) render in cards with shadow, border-radius, and hover effects. KPI measure labels match `big_value` styling (uppercase, letter-spacing, 32px value). KPIs self-size height instead of stretching to match adjacent items.
- **`# span = N` tag**: Explicit 12-column grid span (1-12) on any dashboard item, overrides defaults. Triggers grid mode.
- **`# break` tag**: Starts a new row before the annotated field.
- **`# subtitle` tag**: Adds subtitle text below the item title.
- **`# borderless` tag**: Opt out of card styling on any item type.
- **`# dashboard { columns = N }` tag**: Force N equal columns instead of 12-col grid.
- **`# dashboard { gap = N }` tag**: Custom gap spacing in pixels.
- **`# label` tag**: Custom display label; default is snake_case → Title Case conversion.
- **Responsive breakpoints**: `@container` queries collapse grid to single column at ≤600px, halve to 6 columns at ≤900px. Flex mode wraps naturally.
- **Tables fill width** via `shouldFillWidth: true`, no double scrollbars (overflow-x clipped only on fill-width tables, preserving scroll on wide tables with many columns).
- **big_value integration**: Flattened inner card when inside dashboard, hidden redundant header, centered layout.
- **Tag validation**: Semantic validation for `span` (1-12), `columns` (>0), `gap` (>=0). Dashboard tags resolved at setup time for headless validation support.
- **KPI card fix**: `container-type: inline-size` and `cqi` font scaling scoped to grid mode only. In flex mode, KPI cards size dynamically based on title/value content instead of collapsing due to the circular dependency between `fit-content` width and inline-size containment.
- **Storybook fix**: Pre-bundle heavy dependencies (`@duckdb/duckdb-wasm`, `apache-arrow`, etc.) via `optimizeDeps.include` to fix first-load failures from Vite runtime dep discovery.
- **20+ storybook stories** covering all features plus legacy backwards-compat testing.

https://www.loom.com/share/581a55d7cc1c4cd487fec943555d2825


mtoy-googly-moogly:
My AI is pretty unhappy with the storybook deps loading fix. It feels they are expedient, poorly explained and possibly maintenance problem. Again this is not my area of expertise. But until both our AIs are happy that one is a problem.

mtoy-googly-moogly:
In collaboration with Claude -- I have read or written all of this ... a little more verbose than I would reply myself but I am letting Claude decide how much to write:

First off — this is looking great. The setup-time resolution, the defensive fallbacks, and the error-message wording are all exactly what I was hoping for, and the responsiveness to the earlier rounds shows. Thank you.

A couple of the things I flagged earlier I communicated badly, and you (reasonably) did extra work to satisfy contradictory asks. Let me clear those up. And there's one new thing that recent renderer work has me thinking about that I'd like to get ahead of in this PR.

1. Ownership vs. setup-time resolution — my fault, two asks that pulled opposite ways

Back in March I said "resolve dashboard item metadata at setup time, stop reading tags at render time." Then in April I said "consume span/subtitle/borderless through the #2750 ownership mechanism rather than resolving globally up front." Those two notes point in opposite directions, and you ended up doing both: resolving the child tags at setup time in resolveBuiltInTags and declaring them in childOwnedPaths. That's on me for not reconciling them.

The setup-time resolution is the one I actually want — it's what docs/validation.md calls the default. Given that, the childOwnedPaths entries are now redundant: the setup-time reads already mark those tags as read, so the unread-tag detector never fires on them. I confirmed this — emptying childOwnedPaths entirely leaves the "no warnings for # span / # subtitle / # borderless / # break" tests green.

So: keep the setup-time resolution, and drop the span/subtitle/borderless/break entries from renderer-validation-specs.ts (the dashboard spec can go back to empty). One mechanism, not two.

Smaller related thought, not a blocker: the resolved values currently live as four new fields on FieldBase (_resolvedSpan, _resolvedBorderless, …). Those are dashboard-only concepts sitting on the universal field base. If it's cheap, I'd lean toward stashing them in a per-child config object rather than growing the base class — but I don't want to over-rotate on this if the base-class route is genuinely simpler, so your call.

2. The big-value CSS coupling — also imprecise on my part

When I said "move the dashboard overrides into dashboard-owned styling," that got actioned as relocating the .dashboard-item… selectors out of big-value.css and into dashboard.css — which is good, but dashboard.css now reaches the other way into big-value's internals (.malloy-big-value-card, -value-row, -comparison, …). The leak moved sides rather than closing.

What I was actually after is reducing how much either component knows about the other's internal class names, not just which file the selectors live in. The clean shape is big-value exposing an "embedded/flattened" mode that dashboard opts into, instead of dashboard overriding big-value's guts. That's more than a CSS shuffle, so I don't want to block the PR on it — but let's name it as known debt so it doesn't quietly become a regression source when the two evolve apart.

3. New issue: grid/flex is an inferred mode you can't cleanly override

This is the one from recent experience. I was copying and re-theming dashboards by extending views, and hit a wall that this PR should get ahead of.

Right now the grid-vs-flex decision is inferred: grid if columns set, OR gap set, OR any item has span, OR any item has break. That makes the mode hard to override when you copy a dashboard:

flex → grid is fine — add # dashboard { columns = 3 } to the copy and you're in grid.
grid → flex only works if grid came from columns/gap (negate with # dashboard { -columns }). If grid was triggered by per-item span/break, you can't turn it off from an extension — those triggers live on inherited child fields, and a refinement can't reach back to negate them. I confirmed both: # -span in a refinement only affects fields you redeclare there, and negating columns while a child still has a span leaves you in grid.

I think the real fix isn't a new mode tag — it's that span and break shouldn't be mode triggers at all:

break is meaningful in flex. The row-grouping already runs in both modes (nonDimensionsGrouped splits on break regardless of grid/flex), so in flex a break just forces a new wrapping row. It doesn't need grid and shouldn't switch you into it.
span is meaningless outside grid. It's "N of 12 columns" — flex has no column track, and nothing in the flex CSS path reads it. So span should be a grid-scoped hint: ignored with a warning when there's no grid, not a silent mode switch.

So my proposal: narrow the grid signal to columns/gap only. Make span a grid-scoped sizing hint (warn if used without a grid). Let break work in both modes. The payoff is that the mode is then controlled entirely at the view level, which is cleanly negatable — so grid↔flex round-trips on a copy with a single annotation, and stray spans become harmless no-ops in flex instead of forcing a layout you can't escape.

(Happy to share the little probe I ran if it'd help — it walks the resolved tags for each override case.)

Minor / logistics

The new validator tests are good and cover the right branches. One gap: they assert that an error/warning fires, not what it says. Since I asked for the Invalid <path> on '<field>': expected … got … Fix: … shape, could you add one assertion that pins Fix: (or the "expected … got") so a regression back to a terse message gets caught? And a gap = 0 "no error" assertion would close the last row of that verification matrix.
The branch is a good way behind main and won't build on the current TypeScript toolchain (moduleResolution=node10 is now a hard error). It merges cleanly, so a main merge/rebase should sort it. Looks like #2746 is properly separated out now — thanks for that.

mlennie:
@mtoy-googly-moogly round 2 is merged into `feat/dashboard-ux` (via jswir/malloy#2), so it's in this PR now. Walking your points:

### Ownership vs. setup-time resolution

Kept the setup-time resolution and dropped the span/subtitle/break/borderless entries from the dashboard spec in `renderer-validation-specs.ts`, so `childOwnedPaths` is back to `[]`. The "no warnings for `# span` / `# subtitle` / `# borderless` / `# break`" tests stay green. One mechanism now.

I also took the per-child config route you suggested: the four resolved values no longer live on `FieldBase`. `resolveBuiltInTags` builds a single `DashboardChildConfig` per child and stashes it via `setDashboardChildConfig`, and the dashboard reads it back with `getDashboardChildConfig<DashboardChildConfig>()`. `FieldBase` keeps only `_resolvedLabel`.

### Big-value coupling

Closed it rather than just flagging it as debt. big-value now has an embedded/flattened mode (`.malloy-big-value--embedded` in `big-value.css`), and the dashboard opts in via `customProps` (`{ big_value: { embedded: true } }`) instead of overriding big-value's internals. `dashboard.css` no longer references `.malloy-big-value-card` / `-value-row` / `-comparison`; the only big-value mentions left are `:has(.malloy-big-value)` presence-conditions on the dashboard's own elements. Standalone big-value is unchanged (embedded defaults off).

### Grid/flex mode

Narrowed the grid signal to columns/gap only: `useGrid` is now just `columns !== undefined || gap !== undefined`. `span` and `break` are no longer mode triggers. `break` works in both modes (the row grouping was already mode-agnostic). `span` is a grid-scoped hint that warns when there's no grid instead of silently switching into one. Net effect is the one you described: the mode is view-level and cleanly negatable, so a copied dashboard round-trips grid/flex with a single annotation and a stray span is a harmless warned no-op. Thanks for the probe offer; I worked through the override cases (negating columns with a child span still present, `# -span` in a refinement) and believe they're covered.

### Validator tests

Pinned the message text: the span-range test asserts the `expected ... got ... Fix: # span=6` shape, so a regression to a terse message fails. Added the `gap=0` no-error case, plus span-without-grid-warns and span-with-grid-stays-clean.

### Build / main

Resolved. `feat/dashboard-ux` is synced with `main` now, so the branch builds on the current toolchain and the round-2 work is rebased cleanly on top.

One thing I noticed and left alone: a span out of range under columns mode (e.g. `# span=15` with `columns=3`) emits both the range error and the "ignored in columns mode" warning. It's pre-existing, both messages are true, and suppressing one turns it into a two-step fix, so I left it. Easy to tidy if you'd rather see only one.

mtoy-googly-moogly:
First comment .... (more to come)

So here is how grids work in HTML and Bootstrap

>A Bootstrap/CSS grid is a fixed track of 12 columns that flows: each item claims N columns, and when a row runs
> past 12 the next item automatically wraps to a new row. This isn't that — here every # break group is forced
>  into a single non-wrapping row, and # span just sets each tile's proportion of that one row, so spans never
> wrap and you only get a new row by writing # break yourself.
 
 This will make it possible for a human to write layouts, it will make it unlikely for an AI, intuitively guessing what a "grid" means to generate correct layouts and an AI is particularly bad at correcting itself in these scenarios.
 
 I don't know what you are actually trying to accomplish. I just want to make sure that what you are proposing merging is what you want to accomplish.
 
 "Fix it in documentation" seems to be what you suggested when I pointed it out last time, I think using the word "grid" to describe this feature is going to be a problem. Claude suggested some re-wording of how this feature is presented to the outside world which avoids the word grid, I am not sure if that solves this problem or not. If you called it "skebzig" layout an AI would for sure have to look up the documentation for what that was. I don't know what happens if you just erase the word "grid"
 
 Another solution would be to make this look like an actual grid layout engine.

mtoy-googly-moogly:
Second comment ... 

On #2836 I said the current theme.* mechanism is a backhanded way to override CSS properties, that it should land experimental and not be framed as the real thing, and that mapColor was the last one-off knob of that kind — anything more had to come through a real theme MVP.

This PR adds nine more of exactly that kind (dashboardCardRadius, dashboardCardPadding, dashboardTitleSize/Weight/Color, dashboardValueSize, dashboardGap, …). getThemeValue reads the raw text of a tag and pastes it straight into a <style> block — there's no interpretation, so it isn't a theme, it's the author writing CSS through a tag keyhole.
  
 It also expands it in the wrong direction: the prior art says a theme's core is the color palette and typography-by-role, and our set is already lopsided toward table/card chrome — these nine add still more chrome. And every interim key we ship now is one more thing the real theme design has to reconcile or walk back — the transition cost I flagged on #2836.

Keep the dashboard styling on CSS-variable defaults internally — that's fine. But I don't want to add new public theme tag keys here. New theme surface should come  through the themes design on #2836, not be invented per-component in a layout PR.

mtoy-googly-moogly:
I apologize again for this. These things are only being revealed to me in conversation as I tease comprehension out of the AI. Here's the history

1. The AI pointed out and I agreed that the proposed new layout should not be called "grid" mode because there is a standard meaning for "grid" layout and this wasn't it.
2. I asked you to implement grids or stop calling this grid, because it was your feature, I didn't know what best met your goals.
3. Then you agreed to call it not-grid but didn't really decide what it was
4. So I put on my user hat and asked the AI to think about that with me, by asking "what IS this thing", we hit a snag. I don't think the "almost grids" is a good design.

So I have to push back again. I feel like we need to pick one
- Really implement grids 
- .. or do something else that makes sense, and yes I have a suggestion (below)
- ... or I am open to the suggestion that I just don't understand how cool and useful this feature is as it is.

## Maybe it is "table-like" not "grid-like"

The non-default dashboard layout could behave like a HTML **table**, where you place tiles by how many **columns** they occupy. `columns=N` triggers "columns" mode layout. There is no need for a magic number, you can have a 5 column layout if you want.

- There are N columns; a tile occupies one of them by default. 
- `colspan=n` makes a tile occupy n columns
- `break` ends the current row
- Broken tables with more than N columns of data flow into the next row
  - It's an error, it will be ugly, they will fix it.
  - It would be awesome if we could WARN in validation, but only if that falls out naturally; if it is a big lift, skip it.

## Two modes — use these exact names in every user-facing moment

- **flex mode** — the original/default layout. Tiles flow into a row until it's
  full, then wrap to a new row.
- **columns mode** — you place tiles in columns. Entered by `columns=N`.

Every error message and every doc sentence should speak in those two terms.

## Naming cleanup that falls out of this

- `# span` → `# colspan` 
- Error messages: rewrite the `Ignored # span ...` / `no grid` / `columns mode`
  strings to speak only in flex-mode / columns-mode terms. No "grid."
- Docs (overview + cheatsheet): same vocabulary; drop "12-column grid" and
  "per-row grid" entirely.

## Theme keys

Already handled on your side — thank you. Pulling the nine `dashboard*` theme keys
is exactly right; that surface routes through the themes design on #2836, not
through this PR.

mlennie:
@mtoy-googly-moogly Thanks, this lands it for me. I want to confirm I have the model right before I start, since it changes some current behavior.

Two modes, named exactly that way everywhere (errors, docs, stories): **flex mode** is the default, tiles flow into a row and wrap when it fills; **columns mode** is entered by `columns=N`, where there are N columns, a tile takes one by default, `# colspan=n` makes it take n, `# break` ends the row, and content past N columns flows to the next row. No magic number, any N works. I'll drop the word "grid" everywhere and rename `# span` to `# colspan`.

The one thing I want to confirm explicitly, because it removes shipped behavior: today `gap` (without `columns`) turns on a separate proportional layout where `# span` sets each tile's fraction of a single non-wrapping row. Your model has no proportional sizing, only integer colspan in an N-column layout, and `gap` goes back to being plain spacing rather than a mode trigger. So I'd remove the proportional span behavior entirely (a couple of existing stories use it). **Good to assume that's intended?**

If yes, here's what I'll do:

- rename `# span` to `# colspan`
- make columns mode honor colspan, with overflow wrapping to the next row
- drop the magic 12
- rewrite the validator messages and the docs (overview plus cheatsheet) in flex-mode and columns-mode terms, no "grid"
- add an overflow warning only if it falls out naturally

The theme keys are already out, as you saw.

PR #2760, opened 2026-04-10
fix: improve query execution handling in SnowflakeExecutor

sagarswamirao:
- Align Snowflake variant schema sampling docs with behavior by documenting the default `schemaSampleTimeoutMs` as 15 seconds in `SnowflakeConnectionOptions`.  
- Make `SnowflakeConnection.runSQL` honor connection-level `queryOptions` (including default `rowLimit` / future flags) by merging them with per-call options and passing the merged options to `SnowflakeExecutor.batch`.  
- Harden `SnowflakeExecutor._execute` abort handling to avoid races: fail fast if the `AbortSignal` is already aborted, register the abort listener with `{ once: true }`, and ensure both the listener and timeout are always cleaned up (including when `conn.execute` throws synchronously).  
- Propagate `RunSQLOptions` and `timeoutMs` into session initialization (`_setSessionParams` / `ensureSessionInitialized`) so `ALTER SESSION` statements and user-supplied `setupSQL` respect the same abort/timeout semantics as the queries that triggered them.  
- Ensure `SnowflakeExecutor.stream` also uses the new abort-aware `ensureSessionInitialized`, so aborting a streaming query while the pool is contended or session is initializing will prevent the underlying Snowflake statement from ever running.  
- Preserve existing behavior for query results (including `rowLimit` trimming in `runSQL`) while tightening lifecycle guarantees around timeouts, aborts, and session setup.  
- Covered by existing Snowflake tests, including the abort-under-contention integration test (`snowflake_executor_abort.spec.ts`) and the setup SQL tests; no public API surface changes, only internal behavior and documentation improvements.


mtoy-googly-moogly:
Codex says:

P1

Streaming queries no longer cancel while `conn.execute()` is in flight

This refactor removes the abort listener before conn.execute() and only installs the replacement later inside streamSnowflake(), after complete fires. Any abort that happens while the Snowflake statement is queued or actively running before complete now does nothing, so long-running streaming queries can no longer be cancelled once execution has started.


packages/malloy-db-snowflake/src/snowflake_executor.ts:352-356

P2
`rowLimit` stops the client stream without cancelling the Snowflake statement

When rowLimit is reached we now call handleEnd() directly, which only destroys the Node stream and releases the pooled connection. The previous code called stmt.cancel() in this path. Without that cancel, runSQLStream(..., {rowLimit}) can keep running server-side after the caller has stopped reading, wasting warehouse work and potentially racing with reuse of the same connection.


packages/malloy-db-snowflake/src/snowflake_executor.ts:407-415

P2
Synchronous `conn.execute()` failures still leak streaming connections

_execute() now has a try/catch to clean up listeners and timers when conn.execute() throws synchronously, but stream() still calls conn.execute() without similar cleanup. If the SDK throws before invoking complete, the promise rejects and the already-acquired pool connection is never released, permanently shrinking the pool.


packages/malloy-db-snowflake/src/snowflake_executor.ts:352-353

mtoy-googly-moogly:
Our robot overlords are mostly pleased:

> "Re-reviewed the current head and the three issues raised earlier look addressed"

No idea if this is over cautiousness or not, 

> "One small follow-up: can we add or point me to a targeted regression test for the original P1 timing window, where abort fires after the pool is acquired and `conn.execute()` has started, but before the Snowflake `complete` callback fires? The new contention test is useful, but it mostly covers abort while waiting for a pool connection."

If you want to merge as is, I am OK, or you can answer the unending asking for more. I've kicked off a CI run. I am also now looking at it with my eyes ...

My AI buddy and I would like to see: https://github.com/malloydata/malloy/issues/2771 but think that this PR can go without that.

sagarswamirao:
> Our robot overlords are mostly pleased:
> 
> > "Re-reviewed the current head and the three issues raised earlier look addressed"
> 
> No idea if this is over cautiousness or not,
> 
> > "One small follow-up: can we add or point me to a targeted regression test for the original P1 timing window, where abort fires after the pool is acquired and `conn.execute()` has started, but before the Snowflake `complete` callback fires? The new contention test is useful, but it mostly covers abort while waiting for a pool connection."
> 
> If you want to merge as is, I am OK, or you can answer the unending asking for more. I've kicked off a CI run. I am also now looking at it with my eyes ...
> 
> My AI buddy and I would like to see: #2771 but think that this PR can go without that.

Added a test for abort on mid-execution:

image

We can merge this PR. I'd like to work on https://github.com/malloydata/malloy/issues/2771 in a separate PR.

PR #2762, opened 2026-04-13
refactor(snowflake): redesign VARIANT schema inference and sampling

mtoy-googly-moogly:
## Summary
- **Inference algorithm**: replace the class-based SnowField/SnowObject/SnowArray walker in `snowflake_connection.ts` with a state-based shape-merging algorithm in a new `snowflake_variant_schema` module. Rules: top-level ARRAY/OBJECT from DESCRIBE are authoritative; descendant paths imply ancestor shape; conflicting shapes degrade only the conflicting prefix to variant.
- **Honest type conflicts**: Record fields which are both a scalar and a nested object now require an explicit type cast to access the record variant (e.g. `col::{intParam :: number}.intParam` if sometimes the value of `col` is a scalar like `23`)
- **Tiered sampling**: add an INFORMATION_SCHEMA probe before the variant sample query and pick one of three strategies (`pickSampleStrategy`, pure, unit-tested): full-scan small base tables (catches rare fields), TABLESAMPLE-only on known-large base tables (no unsafe plain-LIMIT fallback), best-effort TABLESAMPLE→LIMIT for views and other unknown-size sources.
- Raise default sample row limit from 100 to 1000. Expose `schemaSampleRowLimit` and `schemaSampleFullScanMaxBytes` as connection options

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "useNx": false,
-  "version": "0.0.375",
+  "version": "0.0.376",
   "command": {
     "version": {
       "allowBranch": "main"
--- a/package-lock.json
+++ b/package-lock.json
@@ -28527,12 +28527,12 @@
     },
     "packages/malloy": {
       "name": "@malloydata/malloy",
-      "version": "0.0.375",
+      "version": "0.0.376",
       "license": "MIT",
       "dependencies": {
-        "@malloydata/malloy-filter": "0.0.375",
-        "@malloydata/malloy-interfaces": "0.0.375",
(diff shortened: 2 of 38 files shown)


sagarswamirao:
P2:
observedTypeToShape('variant') poisoning - theoretical, Snowflake TYPEOF() doesn't return 'variant' in practice. Nice-to-have guard, not blocking.

mtoy-googly-moogly:
@sagarswamirao i am ok with that P2 -- i also looked at the comment and decided to leave it,and the code itself is commented that this is defensive. a future ai which is less smart about TYPEOF() might come along and demand a check there.

> P2: observedTypeToShape('variant') poisoning - theoretical, Snowflake TYPEOF() doesn't return 'variant' in practice. Nice-to-have guard, not blocking.

PR #2770, opened 2026-04-20
fix(render): allow # y tag on numeric dimensions in line charts

jswir:
## Summary

Match bar chart behavior so a numeric dimension (e.g. from a pipeline or `select`) can be assigned to the y-axis with the `# y` tag in line charts.

Previously, the line chart had a separate `measures.length === 0` guard that rejected any query without a true measure (calculation field), even if a numeric dimension had been explicitly tagged with `# y`. The bar chart already supported this pattern; this change brings the line chart in line.

## Changes

- Validate `# y` tagged fields are numeric or calculations (viz-level and embedded), matching bar chart behavior.
- Replace `measures.length === 0` guard with `yChannel.fields.length === 0` so a y-tagged numeric dimension satisfies the y-axis requirement.
- Subtract dimensions used as y from the dimension-count check to avoid false "too many dimensions" errors.
- Exclude dimensions already in `yChannel` from the auto-series fallback so a y-tagged dimension is not also re-assigned as the series field (which previously caused each value to render as its own series).
- Add `numeric_dimensions_as_y` story mirroring the existing bar chart equivalent.

## Test plan

- [x] Existing line chart unit tests pass (`npx jest --testPathPattern='line-chart-plugin'`)
- [x] Verify storybook story `Line Charts > numeric dimensions as y` renders a single line (not one series per data point)
- [x] Verify `misconfigured` case (non-numeric dimension tagged `# y`) shows a clear validation error
- [x] Verify `not_tagged` case (no `# y`, no measures) shows the "no measures" error

### DIFF
--- a/packages/malloy-render/src/plugins/line-chart/get-line_chart-settings.ts
+++ b/packages/malloy-render/src/plugins/line-chart/get-line_chart-settings.ts
@@ -201,11 +201,17 @@ export function getLineChartSettings(
   if (vizTag.text('x')) {
     xChannel.fields.push(getField(vizTag.text('x')!));
   }
+  const isValidYField = (path: string) => {
+    const f = explore.fieldAt(path);
+    return f.isNumber() || f.wasCalculation();
+  };
   if (vizTag.text('y')) {
-    yChannel.fields.push(getField(vizTag.text('y')!));
+    const path = getField(vizTag.text('y')!);
+    if (isValidYField(path)) yChannel.fields.push(path);
   } else if (vizTag.textArray('y')) {
     vizTag.textArray('y')!.forEach(ref => {
-      yChannel.fields.push(getField(ref));
+      const path = getField(ref);
+      if (isValidYField(path)) yChannel.fields.push(path);
     });
   }
   if (vizTag.text('series')) {
@@ -225,7 +231,7 @@ export function getLineChartSettings(
       if (tag.has('x')) {
         embeddedX.push(pathTo);
       }
-      if (tag.has('y')) {
+      if (tag.has('y') && (field.isNumber() || field.wasCalculation())) {
         embeddedY.push(pathTo);
       }
       if (tag.has('series')) {
--- a/packages/malloy-render/src/plugins/line-chart/get-line_chart-settings.ts
+++ b/packages/malloy-render/src/plugins/line-chart/get-line_chart-settings.ts
@@ -202,9 +202,29 @@ export function getLineChartSettings(
     xChannel.fields.push(getField(vizTag.text('x')!));
   }
   if (vizTag.text('y')) {
-    yChannel.fields.push(getField(vizTag.text('y')!));
+    const yFieldRef = vizTag.text('y')!;
+    const yFieldPath = getField(yFieldRef);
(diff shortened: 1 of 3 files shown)


mtoy-googly-moogly:
Two changes that would help, both in `packages/malloy-render/src/plugins/line-chart/get-line_chart-settings.ts`.

### Auto-x should respect explicit channel tags

`get-line_chart-settings.ts:280` — the implicit x fallback picks `dimensions[0]` even when that dimension is tagged `# y`. So:

```malloy
group_by:
  # y
  total_sales
  brand
```

silently renders with `x=total_sales`, `y=total_sales`, `series=brand`. Could the auto-x candidate list exclude fields already assigned to `yChannel.fields` (and ideally fields tagged `# series`)? A Storybook case with the y-tagged numeric dimension listed first would also keep this from regressing.

### Y-type check would be more useful as a logged error

`get-line_chart-settings.ts:209` — the "y must be numeric" check throws from settings construction. That follows the existing chart pattern (bar chart does the same thing at `get-bar_chart-settings.ts:113`), so this isn't a rule you broke — it's actually a rule we're trying to introduce, and we hadn't written it down anywhere you could have seen. That's on us, not on you.

The reason it'd be better routed differently: throws from `create()` are caught by `RenderFieldMetadata.instantiatePluginsForField` and re-emitted as `"Plugin <name> failed for field '<key>': <msg>"` with no `Tag` reference. So VS Code Problems and the headless `@malloydata/render-validator` can't point the user at their `# y` tag in source — they only see a generic "line chart failed" message attached to the field.

If the same check moves to `RenderFieldMetadata.validateFieldTags()` and uses `log.error(msg, tagRef)`, the source location survives. The check still fires before render, the user still gets a clear error before anything tries to draw, headless validation still catches it — but now with a clickable location.

The bar-chart precedent will get migrated to this pattern as part of separate cleanup work; your PR doesn't need to touch anything outside its own diff.

jswir:
Thanks for the context on the validateFieldTags/log.error direction — that distinction (source location survives through re-emission) wasn't obvious from the bar chart precedent, so the pointer was useful.

Addressed in the latest commits:

**Auto-x filter** — auto-x candidate list now excludes fields in `yChannel.fields` and `seriesChannel.fields`, so `# y` or `# series` on a leading dimension no longer gets picked as x. Added a `y_tagged_first` story case with the y-tagged numeric dimension listed first to pin the behavior.

**Y-type check** — moved to `RenderFieldMetadata.validateFieldTags()` using `log.error(msg, tagRef)`. Covers both the top-level `# viz=line { y=field }` ref and the embedded `# y` tag on child fields (scoped to `viz=line` parents so it doesn't run in other contexts).

One follow-up call I made: I also skip non-numeric y fields when building the settings, so the existing "No measures found" throw fires and the chart area renders as a clean error box instead of a degenerate plot with a string y-axis. The specific error still surfaces via `log.error` (and hence VS Code / render-validator) — the skip is just to avoid the misleading render. Happy to back that out if you'd rather have the misconfigured case render-through and wait for the separate plugin-error-state work.

mtoy-googly-moogly:
#2774 is in flight which would supercede this pr ... only one of these two should land, talk to me if you don't want to accept #2774

PR #2789, opened 2026-04-29
feat(db-snowflake): normalize Snowflake privateKey PEMs for SDK compatibility

sagarswamirao:
- Added normalizeSnowflakePrivateKey for Snowflake privateKey inputs to reconstruct flattened single-line PEMs, convert PKCS#1 RSA keys to PKCS#8, decrypt encrypted PKCS#8 with privateKeyPass, and reject legacy OpenSSL-encrypted PKCS#1 (Proc-Type: 4,ENCRYPTED) with a clear error.
- Updated SnowflakeExecutor to normalize in-memory privateKey values once in the constructor so all Snowflake code paths using ConnectionOptions.privateKey get the same behavior.
- Added focused tests covering supported input shapes and failure modes: multi-line and single-line PKCS#8, multi-line and single-line PKCS#1, encrypted PKCS#8 with correct/missing/incorrect passphrase, and legacy OpenSSL-encrypted PKCS#1 rejection.

Note: This PR is limited to normalizing Snowflake RSA privateKey inputs into SDK-compatible PKCS#8. It does not add support for other key algorithms or unrelated PEM formats, and the helper can be extended later if the community needs that support.

Applies it once in SnowflakeExecutor, so all code paths benefit. It currently handles:
- single-line to multi-line PEM reconstruction
- PKCS#1 RSA to PKCS#8 conversion
- encrypted PKCS#8 decryption when privateKeyPass is provided
- clear rejection of legacy OpenSSL-encrypted PKCS#1 (Proc-Type: 4,ENCRYPTED)


the-michael-toy:
I dug into this and came out in a different place than the PR. Two findings drove it:

**Snowflake's in-memory pickiness is intentional.** The SDK's file path (`privateKeyPath` + `privateKeyPass`) goes through `loadPrivateKey()`, which uses `crypto.createPrivateKey()` and is permissive about format and encryption. The in-memory path (`privateKey`) deliberately routes around that and validates with a strict `startsWith('-----BEGIN PRIVATE KEY-----')` check. The comment in `util.js` says it explicitly: *"The private key is expected to be decrypted when set in the connection string."* That's a design choice, not an oversight, and I feel obligated to respect it. So I don't want to change malloy's interpretation of `privateKey` — it should remain a strict pass-through to the SDK.

**The decryption support also feels wrong on principle.** Encryption-at-rest only adds real security when the encrypted key and the passphrase travel through *different* channels with different access controls. In any path that reaches malloy with both halves at once — config file, runtime-constructed config object, env vars to the same process — the encryption has already stopped doing security work. The legitimate "key and passphrase from separate channels" pattern is exactly what `privateKeyPath` + `privateKeyPass` is for, and the SDK already handles it. If a caller (Publisher or anyone else) has both halves in hand at runtime, they're in a position to decrypt themselves before calling malloy and hand us a clean PKCS#8 — that's the right layer for it, not malloy.

What I think belongs in Malloy is one of:

1. **Nothing beyond what the SDK already offers.** Customers with PKCS#1 or encrypted keys use `privateKeyPath` (+ `privateKeyPass` if encrypted). Documentation, not code.

2. **A narrowly-scoped helper property — `privateKeyPEM`** — that accepts a few common unencrypted shapes and normalizes them into a legal `privateKey` for the SDK. Specifically:
   - Newline reconstruction (for PEMs flattened by env-var pipelines)
   - PKCS#1 → PKCS#8 conversion
   - **No decryption.** Encrypted inputs get rejected with an error pointing at `privateKeyPath`.

Option 2 gives the customers you mentioned a working in-memory path for the unencrypted PKCS#1 case without compromising on decryption or silently overloading the SDK's `privateKey` contract. The name signals that we normalize the input, so the contract stays honest.

Either way, I'd want to drop the silent transformation of `privateKey` and the decryption logic from this PR. Happy to look at a follow-up that takes the option-2 shape if you think it's worth doing.

sagarswamirao:
That makes sense. Thanks for digging into the SDK behavior and highlighting the intent behind the strict `privateKey` handling.

My initial goal here was to support the range of key formats that the Snowflake SDK accepts via `privateKeyPath` (through `crypto.createPrivateKey()`), since users transitioning to in-memory `privateKey` can run into friction. In practice, we’ve seen PKCS#1 keys and flattened PEMs fail in this path, leading to confusion and extra back-and-forth.

That said, I agree that silently transforming `privateKey` (especially beyond minimal normalization) isn’t the right approach, and we should respect the SDK’s contract.

I also agree that we shouldn’t try to support every format or include decryption logic at this layer, keeping behavior simple and predictable is more important.

A reasonable middle ground would be:
- Keep `privateKey` as a strict pass-through to the SDK
- Optionally support newline reconstruction only (to handle env-var–flattened PEMs), as a minimal, non-invasive normalization
- Let the SDK surface errors for unsupported formats

I did consider introducing something like a `privateKeyPEM` field for broader normalization, but wasn’t sure if expanding the Malloy config surface in that direction is desirable.

I’ll avoid introducing broader normalization (e.g., PKCS#1 → PKCS#8) or new config surface like privateKeyPEM for now, but happy to revisit that separately if needed.

Happy to revise the PR along these lines, let me know if this aligns with your thinking.

mtoy-googly-moogly:
Well, the question is, is publisher the only one who wants this, or does everyone want it, or does everyone want something a little different.

If it is something publisher wants, which everyone else is going to want EXACTLY like publisher wants it, then we should add to core.  Otherwise it's probably best still in publisher.

I do not know the answer to these questions, Do you have any insight?

sagarswamirao:
Anyone using malloy-db-snowflake connection will need it. But I agree with you. We should just have what snowflake-sdk offers as a part of the Malloy core. For now we have added it at the Publisher level.
Closing this PR for now. I am happy to create a small PR if you think there's something from this PR that'll help the Malloy community. 

Thanks for looking into this. Really appreciate it :)

PR #2836, opened 2026-05-25
feat(render): add explicit theme prop on MalloyRenderer

mlennie:
## Summary

Walkthrough video: https://www.loom.com/share/e31e852ca1664526a1e7a355a07bddd6

  Adds an optional `theme?: MalloyExplicitTheme` prop on `MalloyRendererOptions` (and the underlying `MalloyRender` SolidJS component) so embedders can inject colors, fonts, and palette values directly into the
  renderer instead of relying on the `# theme.*` annotation chain or the `--malloy-theme--*` CSS variable fallback alone.
  
  Motivation: a downstream consumer (Malloy Publisher) is adding an in-app Theme Editor where operators pick palette + font values that need to reach the chart canvas, table chrome, and dashboard tile colors. The
  annotation channel works for model-level styling but doesn't give the embedder a runtime hook; the CSS variable fallback covers HTML but doesn't reach inside Vega's chart canvas. The new prop fills that gap.
  
  ## What it covers
  
  The `MalloyExplicitTheme` interface exposes (all optional):
  
  - Table chrome: `tableRowHeight`, `tableBodyColor`, `tableFontSize`, `tableHeaderColor`, `tableHeaderWeight`, `tableBodyWeight`, `tableBorder`, `tableBackground`, `tableGutterSize`, `tablePinnedBackground`,
  `tablePinnedBorder`
  - Renderer chrome: `background`, `fontFamily`
  - Choropleth maps: `mapColor` (saturated end of the gradient used by `# shape_map` / `# segment_map`)
  
  ## Resolution order (highest wins)
  
  1. `theme.<key>` from `MalloyRendererOptions`
  2. `# theme.<key>` on the result
  3. `## theme.<key>` on the model
  4. `var(--malloy-theme--<kebab>)` CSS fallback
  
  Empty string is treated the same as `undefined` so a cleared form input drops back to the next layer rather than emitting a malformed CSS declaration.
  
  ## Backwards compatibility
  
  The prop is purely additive. When `theme` is `undefined` (every existing consumer), behavior is bit-identical to `main`:
  
  - `getColorScale` gained a 4th positional parameter that defaults to `undefined`. With no `explicitTheme`, the internal `sequentialGradient` is `null` and every branch falls back to its prior literal range. The
  ordinal branch's new `isRectMark` guard returns the same `['#C2D5EE', '#1A73E8']` on both sides of the guard when `sequentialGradient` is null.
  - Map plugins resolve to `background: 'transparent'` (matches the long-standing default).
  - The 8 in-tree call sites that pass only 2 args to `getColorScale` (scatter-chart, html/cartesian_chart, html/sparkline, html/area_sparkline, html/shape_map, html/segment_map, html/point_map, 
  html/bar_sparkline) are untouched.
  - Storybook stories without a `theme` prop render unchanged.
  
  ## Notable details
  
  - **Map plugins** now set `background: themeForRender?.background || 'transparent'`, snapshotted into a local at the top of `renderToDOM` so an async theme swap mid-render doesn't race the in-flight Vega
  pipeline.
  - **`mapColor` gradient low-end** is derived from the WCAG relative luminance of the high-end. A light high pairs with a dark neutral; otherwise pairs with `#f5f5f5`. Bad-hex inputs fall back to the light
  default.
  - **Ordinal heatmaps** keep their tuned `['#1A73E8', '#E8710A']` palette even when `mapColor` is set: the new `isRectMark` guard on the ordinal branch matches the existing guard on the quantitative branch.
  - **`MalloyViz.updateOptions({ theme: ... })`** deep-merges the `theme` key so partial updates preserve previously-set fields. Passing `theme: undefined` clears the whole theme.
  - **Plugin style-override precedence**: explicit `theme` keys suppress matching plugin-emitted CSS-var overrides. Previously bar-chart and line-chart plugins set `--malloy-render--background` from
  `vegaConfig.background` in `getStyleOverrides()`, which silently beat an embedder-supplied `theme.background` in the same inline style block.
  - **Package exports**: `package.json` adds `"import"` (ESM `.mjs`) and `"require"` (CJS `.umd.js`) conditions. The ESM bundle was already shipping in `dist/` but unreachable through the exports map.
  `@malloydata/malloy-explorer` is an ESM consumer of `@malloydata/render` that imports named symbols, so this change is required for its build to resolve.
  
  ## Test plan
  
  - [x] `npm run -w packages/malloy-render build` — clean exit, both `.mjs` and `.umd.js` artifacts produced.
  - [x] `MALLOY_DATABASE=duckdb npx jest --selectProjects malloy-render` — 101/101 pass (83 pre-existing + 18 new in `html/utils.spec.ts` covering `getColorScale` with and without `explicitTheme`, the `isRectMark`
   guards, luminance swap on a light `mapColor`, and the malformed-hex fallback).
  - [x] Manual end-to-end against a downstream consumer (Malloy Publisher): map background follows `theme.background`; `mapColor` recolors `# shape_map` / `# segment_map`; clearing a theme field falls back through
   the annotation / CSS-var chain; existing samples without a `theme` prop render bit-identically.
  
  ## Paired consumer
  
  A downstream consumer is being added in [malloydata/publisher](https://github.com/malloydata/publisher) (in-app Theme Editor + viewer light/dark toggle). That PR is gated on this one merging and the next
  renderer publish.
  
  ## Notes
  
  - All 12 commits include DCO sign-off.
  - No version bump or changelog entry: the Malloy CI bot handles those on merge to `main`.
  - No README in `packages/malloy-render/`; documentation lives as JSDoc on `MalloyExplicitTheme` and `MalloyRendererOptions.theme` and is surfaced via `docs.malloydata.dev`.


mtoy-googly-moogly:
The following is a computer generated synthesis of my first look at this feature. Sometimes I am the person who knows a piece of code the most deeply, even more than the AI, and I am very confident that the synthesis is the correct stance.

This is not one of those times. So if this seems short sighted or wrong headed, let's let that be part of the conversation, I am OK with that, I will learn along with you.

------

First, thank you for this — and the walkthrough. It surfaces something real I'd been able to ignore until now: CSS variables theme the DOM chrome but can't reach inside Vega's canvas, so there genuinely is no runtime hook for an embedder today. That's a true gap and you've shown exactly where it bites.

Here's the reframe I'd ask you to take on, though. What you've built isn't really "a prop on the renderer" — it's **the themes feature of the renderer**, and the moment it ships it's a public, supported surface: something we document, hold stable, and have to live with. I'd like us to treat it that way from the start rather than arrive at it through a feature PR.

So before this lands as code, I'd like you to step back and design it as that public feature — spelled out, written down — answering these three questions, in this order:

1. **What *is* a theme — the data?** The complete set of theme properties, the type/format of each, and what each one controls. Right now the keys live in three places that have to stay in sync (the prop interface, the `# theme.*` annotation names, and the `--malloy-theme--*` CSS variables), and they don't all behave the same way — `mapColor` is baked into a Vega scale while the rest become CSS vars. The feature needs one definition of what a theme contains.

2. **How is a theme *specified* — the input?** There are several channels now: `## theme.*` on the model, `# theme.*` on the result, this new embedder prop, and the CSS-var fallback. What's the precedence and the merge story across them, and is that the model we actually want? (For instance: should the embedder always win, or should a model be able to pin a value the embedder can't override?)

3. **How does the renderer *subscribe* — the output?** Given a fully resolved theme, how does each value reach the pixels — DOM chrome via CSS vars, the chart canvas via the Vega spec, maps, and so on? Ideally each property would declare where it goes, so applying a theme is uniform rather than hand-wired per key.

My instinct is that (1) is the keystone — if each property is defined once with its type, its meaning, and where it gets applied, then (2) and (3) largely fall out as a read-in-precedence pass and an apply pass. But I'd like to see the feature designed and written down rather than inferred from the diff.

None of this is a knock on the code — it's good, and the Vega-canvas finding is exactly the kind of thing that makes me want to get the public shape right. If Publisher needs unblocking in the meantime, we can talk about landing a narrow, explicitly-provisional version while the real feature design proceeds.

mlennie:
@mtoy-googly-moogly 

Thanks, this reframe is right and I'm glad you pushed on it. Agreed it's the themes feature, not a prop, and a public surface the moment it ships, so I wrote the design down rather than leave it implied by the diff. Full design doc here: https://docs.google.com/document/d/1Jk0lNHYY5kafmsJyJ-piPhpWMSB290hVKRZsAZfujx0/edit?usp=sharing

**On timing**: Publisher's theme-customization feature can't deploy until this prop merges, so I'd like to take you up on your offer to land a narrow, explicitly-provisional version, the current prop marked `@experimental` (shape may change), while we settle the design. 

The surface Publisher consumes is the prop's key set, which the registry work (mostly internal) should keep stable, so the exposure is low-risk. If you'd rather settle the design first, I'm happy to wait, just flagging Publisher is what's blocked.

The short version, and where I'm unsure:

  **1. The data.** I think you're right that this is the keystone, so I went in that direction: one property registry, each property defined once with its type, default, and where it applies, so the interface, the CSS defaults, and the docs all derive from that one definition instead of the four places they drift across today (the fourth being the `mapColor` Vega path). If a single registry feels like over-structure for ~14 properties, that's the first place I'd want your read.
  
  **2. The input.** Resolving becomes one precedence pass (embedder prop > `# theme` > `## theme` > CSS fallback > default), first defined wins, empty means unset. On your embedder-vs-model-pin question: I lean embedder-wins for v1 since the embedder owns the page, but I don't hold it strongly, the one thing I'd flag is that pinning is only additive later if we build the v1 resolve signature to carry a marker now, so it's worth deciding even if we don't implement it. One honest finding from writing this up: today `mapColor` and the map background only honor the embedder prop and silently ignore the annotations, so "make the chain uniform" is a real behavior change, not just a refactor.

  **3. The output.** Your "each property declares where it goes" instinct mostly holds: each property names a sink (CSS var for chrome, Vega-spec write for the canvas), and applying is one dispatch pass. The honest caveat is that it doesn't fully fall out, `mapColor` has real conditional logic (it branches on scale type / rect-mark and derives its low end by luminance), so the registry co-locates that logic with the property rather than deleting it, and there's a third background path (charts emit `--malloy-render--background` and `render.tsx` suppresses it) the clean two-sink picture has to absorb. I'd rather name those than pretend they disappear.

  The parts I'd most like your read on: whether the registry is the right level of structure, the embedder-wins-vs-pin contract, and the two-tier `--malloy-theme--` / `--malloy-render--` namespace (I lean keep-and-document). And if the overall shape is wrong just let me know. Thank you

mtoy-googly-moogly:
My summary:

- The ESM entry point change feels like it doesn't belong bundled in here and is a distraction for both me and the AI, is there some reason it needs to be in here?
- In the long run, I think renderer needs to support an MVP notion of "theme" which looks very different than what `theme.*` looks like today. A quick AI survey of what Vega and pretty much everyone else does shows more of a "class / property" taxonomy of styling.
- I am open to landing a version of this very close to what the PR looks like, with the understanding that anything else that people might want to theme, like say the color palette of a chart, or the fonts used in labels vs. the fonts used for values, has to come through a "theme MVP" and not what exists now, which is sort of a backhanded way to override some CSS properties.
  - The problem with this is there are likely to be rough edges when transitioning between "interim CSS property themes" and "grownup themes."

After much conversation, here is the AI written summary:

**To land now (this PR), three small things:**

1. **Pull the `package.json` `import`/`require` change out into its own PR.** It's a separate packaging decision — it affects every consumer and has its own history, independent of theming — so it's better judged on its own, keeping this PR about the theme prop.
2. **Mark the new surface (`theme` prop + `MalloyExplicitTheme`) as experimental, and let's not frame it as the finished theme** — in docs or in how it's named. Whether that's a JSDoc note or something more visible at the call site is open; preferably something a consumer can't miss, since the shape may change as the fuller set lands.
3. **`mapColor` lands, marked experimental — and it's the last one-off knob of this kind.** It's the single property that crosses from chrome into real chart-data color, let through to keep Publisher unblocked. Beyond it, new theme/color surface should arrive as the designed theme below, not another bespoke property.

**Follow-up (separate — the real design):** a defined set of theme properties — palette, typography-by-role, continuous schemes, chrome — the set that makes it a theme people can actually use, driven by what theme consumers need. The existing design doc is a starting point; the property set should be built from the prior art plus actual consumer needs rather than what's cheapest to wire.

**Net:** ship the runtime channel now (minus the packaging change, marked experimental), and treat "what a complete theme contains" as the next piece — unblocking Publisher without publicly committing to a property set that would have to be walked back.

mlennie:
Agreed on all of it, this is the right split. I'll land this PR as the runtime channel only and treat the full property set as the follow-up.

For this PR:
1. Pulling the package.json import/require change into its own PR. It's a packaging decision for every consumer and shouldn't ride along with theming.
2. Marking the theme prop and MalloyExplicitTheme @experimental, with a call-site signal a consumer can't miss, and docs that won't frame it as the finished theme.
3. mapColor stays, marked experimental, as the last one-off of its kind. Any new color or theme surface after it comes through the designed theme, not another bespoke prop.

The follow-up is the real piece: a defined property set (palette, typography-by-role, continuous schemes, chrome) built from the prior art plus what consumers actually need, with the design doc as the starting point rather than the spec.

One thing on your side: the check-permission job is red because it's a fork PR, so the rest of CI needs a maintainer to approve the run. Anything you need from me there?

mlennie:
Pushed an update that clears both points.

I merged current upstream/main into the branch, so it is no longer behind and carries the latest CI process changes. The one conflict was in render.tsx, from the dashboard layout PR (#2943): this branch had moved generateThemeStyle and getThemeValue into a new theme.ts module, while #2943 added the dashboard CSS variables to the old in-render.tsx copy. I kept the relocation and ported #2943's --malloy-render--dashboard-* variables into theme.ts's generateThemeStyle, so the dashboard theming is preserved. The render-validator suite, including the dashboard tag tests, passes.

On the agreed changes: the @experimental markings (the theme prop, MalloyExplicitTheme, and mapColor) and the package.json export split were committed locally but had not actually been pushed, so the remote was stale. They are on the branch now. package.json no longer carries the export conditions here, since those live in the merged #2945, so its diff against main is empty.

On the call-site signal: you left it open between a JSDoc note and something more visible. I went with the @experimental JSDoc on the prop, MalloyExplicitTheme, and mapColor (each noting the shape may change) and left it there rather than adding something heavier you hadn't asked for. If you'd prefer something a consumer really can't miss, a rename or a runtime notice, tell me which and I'll add exactly that.

I also did a correctness and security pass over the diff. One small cleanup landed: I dropped a redundant theme alias in the two map plugins. One security note worth recording: theme values (props.theme and the # theme / ## theme annotations) are written unescaped into the container's inline style attribute, so a value can add an extra declaration on that element, e.g. background-image: url(...). It is bounded by that inline-style sink, with no selector or HTML breakout. I did not add input filtering here, because background and tableBackground legitimately carry url(...) images and a blocklist would break them; per-key format validation belongs with the designed property set in the follow-up.

Green locally: typecheck, the theme and render-validator jest suites, eslint, and prettier. Ready for you to kick off CI and take another look.

mtoy-googly-moogly:
LGTM!

PR #2842, opened 2026-05-27
[#788] fix(render): 3-tier x-axis label rotation, stop truncating short labels

### LINKED ISSUE #2777: renderer hides x-axis too aggressively
### What happens?

Image

vs. 

Image

### To Reproduce

adjust width of a line chart and see how the x-axis (years) collapses as you narrow the window. 

### OS:

macos

### Malloy Client:

publisher > renderer?

### Malloy Client Version:

credible - 380?

### Database Connection:

duckdb

mlennie:
Fixes #2777 and https://github.com/malloydata/publisher/issues/788 

  ## The bug

  `getXAxisSettings` in `chart-layout-settings.ts` truncated x-axis labels like `1979` to `1...` on narrow charts even when the natural label would have fit. The labelLimit was capped at 35% of plot height regardless of the label's actual width, so a 4-character year (≈25 px) got clipped to a few pixels on tightly-packed dashboards.

  Rotation was also binary (0° or -90°), so any label that didn't fit horizontally went straight to vertical. Medium-length labels (month names, short categorical values) lost a sensible middle option.

  ## The fix
  
  Replace the two-state flip with a 3-tier algorithm:

  | Tier | Angle | Picked when |
  |---|---|---|
  | 1 | 0° (horizontal) | `xSpacePerLabel >= maxStringSize + labelSeparation` |
  | 2 | -45° (diagonal) | horizontal fails but the rotated bounding box `(width + lineHeight) * cos(45°)` fits both horizontally and vertically |
  | 3 | -90° (vertical) | fallback. Truncate only when the natural label exceeds the post-title height budget |

  Tier 3 truncation reserves `labelLimit + ellipsesSize` for the rendered band so Vega's appended ellipsis stays inside the reserved area (matching the prior code's intent that got lost when the threshold-based math was refactored). `labelLimit` is clamped to at least 1 in the truncation path so Vega does not interpret 0 as "no limit".

  The diagonal tier reserves `(width + lineHeight) * cos(45°)` rather than `width * cos(45°)`, because the bounding box of a rotated label includes both the width and the line-height projection. Reserving only `width * cos(45°)` underestimates by 8 px or so, causing labels to clip into the plot area at the boundary.

  The horizontal-tier `xAxisHeight` formula now matches the rotated tiers' formula (both include `xLabelPadding` via a shared `titleBlock` term). Existing horizontal-axis charts pick up about 6 px more bottom padding as a result. That alignment is intentional, the asymmetry in the old code was a latent inconsistency rather than a deliberate choice.

  ## Tests

  Added `chart-layout-settings.spec.ts` covering:
  
  - Tier 1 selection (short label + wide chart, no truncation)
  - Tier 2 selection (medium label + medium chart, diagonal -45°)
  - Tier 2 band geometry: `(width + lineHeight) * cos(45°)`
  - Tier 3 selection without truncation (short label + narrow chart)
  - Tier 3 selection with truncation (long label + dense axis): asserts `labelLimit = labelHeightBudget` and `reservedLabelBand = labelLimit + ellipsesSize`
  - Degenerate case (`chartHeight <= titleBlock`) regression test: pins the falsy-zero idiom fix so `xAxisHeight` does not balloon past the chart bounds
  - `isSpark` hides the axis

  `getXAxisSettings` was previously uncovered; the spec mocks the DOM text-measure helpers (`getTextWidthDOM`, `getTextHeightDOM`) so it runs hermetically in the existing node-mode Jest project.

  ## Manual verification
  
  Verified visually in Storybook against the `line_charts.stories.malloy` stories (`Lots Of Lines`, `Lots Of Lines Nested`, `Dates`) by resizing the canvas across the full width range. Short year labels remain readable (no `1...` / `2...`) at every width; long date strings rotate cleanly without clipping into the plot.

Screenshot 2026-05-27 at 1 21 09 PM
Screenshot 2026-05-27 at 1 24 10 PM
Screenshot 2026-05-27 at 1 24 19 PM


mtoy-googly-moogly:
Nice fix — traced it through and the short-label case (#2777) is genuinely solved, and the band/extent math checks out against how the spec generators spread `xAxis` into Vega.

One thing I want your call on before this lands. Removing the `0.35 * plotHeight` cap means a long-but-fitting label now reserves its full length in tier 3, and since the spec uses `autosize: {contains: "padding"}`, that band eats straight into the plot. A ~160px label on a md chart (252px) drops the plot from ~95px (old, truncated) to ~14px (new, full label). lg charts are less dramatic (~149 → ~98). The Storybook stories you checked land in tier 2 or tier-3-with-truncation, so I don't think this medium-tall-chart + long-vertical-label case got exercised.

That `0.35` cap was there precisely to stop labels from starving the plot, so this is a real tradeoff, not a bug per se. Two ways to go:

1. **Accept it** — the short-label win is the common case, and genuinely-long categorical labels crushing the plot is rare/already-ugly. Fine by me if you'd rather not add complexity.
2. **Re-cap tier 3** — clamp `reservedLabelBand` (the no-truncate branch) to some fraction of `chartHeight`, which keeps the short-label fix and restores plot protection.

Your call — happy either way, just don't want to ship the 14px-plot case by accident. Everything else (tests, lint, the license-header cleanup) looks good.

mlennie:
@mtoy-googly-moogly Great catch. I went with option 2: re-capped tier 3.
  
  Tier-3 vertical labels can now reserve at most 35% of chart height, so a long categorical label can't starve the plot like your md example (the ~160px label that dropped the plot to ~14px now keeps ~75px). Two deliberate choices worth flagging:

  1. I kept the 0.35 fraction but apply it to chartHeight rather than the old plotHeight = chartHeight - 2*titleOffset - titleSize - ellipsesSize expression. That base subtracted a label-width term (ellipsesSize) from a height and collapsed to single digits on short charts (~9px at the smaller presets), which was part of what made short labels over-truncate in the first place. 0.35 * chartHeight gives the same plot protection without that degenerate collapse. Happy to switch to a principled post-title plotHeight if you'd prefer, just not the old expression.
  2. On the truncate path I reserve labelLimit + ellipsesSize for the band (matching the pre-rewrite labelHeight += ellipsesSize), so Vega's appended ellipsis stays inside the reserved area. It errs ~10px toward protecting the plot, never toward starving it.

  Short labels still skip truncation entirely (labelLimit = 0), so #2777 stays fixed, and I added a unit test pinning your exact md + long-label case (chartHeight=252, ~160px label, asserts the band caps at ~98px instead of reserving the full 160px).

  Verified in Storybook across all six sizes (xs through 2xl) on both bar and line charts (screenshots below):
  - Long categorical labels: truncate with an ellipsis and keep a readable plot at md, and the reserved band scales with chart height (less truncation as the chart grows), which is the behavior you described.
  - 4-digit years: never clipped, and you can see the full tier progression (vertical on xs/sm/md, diagonal on lg, horizontal on xl/2xl).

Screenshot 2026-06-01 at 11 40 00 AM
Screenshot 2026-06-01 at 11 46 09 AM
Screenshot 2026-06-01 at 11 46 30 AM
Screenshot 2026-06-01 at 11 47 16 AM

PR #2858, opened 2026-06-02
fix(dialect): escape control chars in backslash-style literals and identifiers

jswir:
## Problem

Backslash-style dialects (BigQuery, Snowflake, MySQL, Databricks) render SQL string literals and quoted identifiers with C-style escapes, and such tokens **cannot contain a raw newline** — an embedded LF/CR terminates the token early and the engine reports `Syntax error: Unclosed string literal`.

The base `Dialect` escaped only the backslash and the closing delimiter:

```ts
const escaped = literal.replace(/\\/g, '\\\\').replace(/'/g, "\\'");
```

So any value containing a newline, carriage return, or tab was emitted **raw** between the quotes, producing invalid SQL. A value with an embedded CRLF reaching a generated string literal hit exactly this against BigQuery (`Unclosed string literal`).

## Fix

Introduce a single helper, `escapeBackslashStyle(body, delim)`, that escapes the backslash, the closing delimiter, and the control characters `\n` / `\r` / `\t`. Both `sqlLiteralString` and `sqlQuoteIdentifier` route through it — the identifier path had the **identical** gap, since a raw newline breaks a backtick-quoted identifier the same way it breaks a string literal.

### Scope decisions

- **By character set, not by dialect.** Escaping is applied uniformly across all four backslash dialects, because all four decode `\n`/`\r`/`\t` to the same bytes. This is correct backslash-token encoding, not a BigQuery-only workaround: values stay byte-exact (ordered `<` / `>` comparisons remain correct) and generated SQL stays single-line. Carving out BigQuery would fragment the shared `EscapeStyle.Backslash` abstraction to emit *messier* raw-newline SQL elsewhere for no correctness gain.
- **Only the three confirmed control characters.** `\0` and the Unicode line/paragraph separators (U+2028 / U+2029) are passed through. There is no evidence they terminate these dialects' lexers, and escaping bytes we haven't verified risks corrupting values; if a concrete failure surfaces for one of them, it gets added with a test.
- **The `doubled`/ANSI branch is unchanged** (Postgres, DuckDB, Trino): those literals may span raw newlines and treat backslash as a literal character, so escaping there would corrupt data.

## Why the existing test didn't catch it

`escape.spec.ts` already had a `newline` case, but its simulated `parseStringLiteral` consumed raw newlines as ordinary characters (`value += c`), so backslash dialects round-tripped "successfully" while emitting SQL a real engine rejects. This PR:

- makes the simulated **literal and identifier** parsers reject a raw newline (matching real lexers);
- adds `carriage_return` / `crlf` / `multiline_blob` cases;
- adds an **exact-encoding** block that pins the literal and identifier output byte-for-byte, so a decodable-but-different form (e.g. `\x0a`) can't slip past round-trip alone.

A simulated parser is only ever as faithful as we make it, so the fix is also proven end-to-end: `test/src/databases/all/escape.spec.ts` adds `newline` / `crlf` / `multiline_blob` to the live-engine `STRING_CORPUS` (and teaches `toMalloyString` to escape control chars), validating the generated SQL against the real databases.

## Testing

- `jest --selectProjects malloy-core` — the `escape.spec.ts` round-trip + exact-encoding corpus across every registered dialect (729 tests passing).
- `test/src/databases/all/escape.spec.ts` runs against the live database matrix in CI. A lone `\r` is intentionally left to the unit corpus, since some engines normalize a bare CR, which would make a live assertion engine-dependent rather than a clean escape check.

### DIFF
--- a/packages/malloy/src/dialect/CONTEXT.md
+++ b/packages/malloy/src/dialect/CONTEXT.md
@@ -262,7 +262,9 @@ Every dialect must declare three fields that drive identifier and string-literal
 
 The base class provides safe implementations of `sqlQuoteIdentifier`, `sqlLiteralString`, and `sqlLiteralRegexp` driven by these flags. Dialects normally do not override them. If a subclass forgets to set a flag, the base method throws at first use with a message naming the dialect and the missing flag — `escape.spec.ts` exercises this fail-fast path on every registered dialect, so a forgotten flag fails CI rather than producing wrong SQL silently.
 
-The contract is verified by `packages/malloy/src/dialect/escape.spec.ts`, which iterates `getDialects()` and asserts that an adversarial input corpus round-trips through each escape function. A new dialect is covered automatically the moment it is registered in `dialect_map.ts`; you do not need to edit the spec file.
+For `EscapeStyle.Backslash`, both the literal and identifier paths share `escapeBackslashStyle`, which escapes the backslash, the closing delimiter, and the control characters `\n` / `\r` / `\t`. A raw newline terminates a backslash-style token early in BigQuery ("Unclosed string literal"); the other backslash dialects tolerate it but decode the named escapes to the same bytes, so escaping uniformly keeps values byte-exact. The escape set is deliberately those three control characters — `\0` and the Unicode line/paragraph separators are passed through, since there is no evidence they break these lexers.
+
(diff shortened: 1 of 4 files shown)


mtoy-googly-moogly:
Fixed in #2859

jswir:
Thanks @mtoy-googly-moogly — agreed, #2859 is the right fix. This PR only handled the backslash string-literal case and wrongly assumed the doubled/ANSI dialects + the identifier path were safe. Following #2859.

PR #2871, opened 2026-06-09
  fix(render): honor # label tags on chart axis, legend, and tooltip titles

mlennie:
## Summary

Bar, line, and scatter charts rendered their axis, legend, and tooltip titles from the raw field name instead of the field's `# label`, so a dimension or measure tagged with `# label="..."` showed e.g. `audience_name` on the axis instead of `Audience`. The table and dashboard renderers already honor the label; this brings the charts in line.

Before:
chart-labels-before-bug

After:
chart-labels-after-fixed

## Root cause

`getLabel()` (which resolves the `# label` tag at setup time) was added to the field abstraction and adopted by the table and dashboard renderers, but the chart plugins still read `.name` for their display text.

## What changed

- Bar, line, and scatter charts use `field.getLabel()` for axis titles, legend titles, and (bar/line) tooltip labels, covering both the custom `# tooltip` entries and the default measure row.
- Every `.name` used as a data reference (vega `field:` paths, `row.column(...)` lookups) is left unchanged, as are dimensional-series tooltip entries, whose value is the series value rather than a field name.
- The bar and line synthetic series fields gain a `getLabel()` so the multi-series legend path keeps working.
- For a multi-measure (measure-series) chart, the color-legend entries are the raw measure field names from the fold transform; they are relabeled to their `# label` through an ordinal scale, so the legend matches the axis and tooltip and an explicit `# label=""` renders blank.

## Before / after

Before: x-axis `audience_name`, y-axis `reach`. After: x-axis `Audience`, y-axis `Reach (HH)`.

## Testing

- Added bar and line spec tests asserting the axis, legend, and measure-series legend titles come from the labels, and that an explicit empty `# label=""` is preserved.
- Verified in Storybook across bar, line, and scatter; single and multi-measure; dimensional and measure series; empty labels; and custom and default tooltips.
- Reproduced the original behavior on the pre-fix baseline (raw field names) and confirmed it is resolved.

### DIFF
--- a/packages/malloy-render/.npmignore
+++ b/packages/malloy-render/.npmignore
@@ -4,3 +4,4 @@ tsconfig*
 *.spec.*
 storybook-static
 CONTEXT.md
+dist/module/plugins/spec-test-support
--- a/packages/malloy-render/src/component/bar-chart/get-custom-tooltips-entries.ts
+++ b/packages/malloy-render/src/component/bar-chart/get-custom-tooltips-entries.ts
@@ -21,7 +21,7 @@ export function getCustomTooltipEntries({
   customTooltipFields.forEach(f => {
     records.forEach(rec => {
       customEntries.push({
-        label: f.name,
+        label: f.getLabel(),
         value: () =>
           applyRenderer({
             dataColumn: rec.__row.column(f.name),
--- a/packages/malloy-render/src/component/vega/measure-series-label-scale.ts
+++ b/packages/malloy-render/src/component/vega/measure-series-label-scale.ts
@@ -9,15 +9,12 @@ import type {NestField} from '@/data_tree';
 export const MEASURE_SERIES_LABEL_SCALE = 'measureSeriesLabel';
 
 /**
- * For a measure-series legend the color-domain values are the raw measure
- * field names; this ordinal scale maps them to their `# label` so the legend
- * entries match the axis and tooltip. Keeping the labels in scale data (not
- * in a parsed expression) renders any label verbatim. An explicit empty
- * `# label=""` intentionally renders a blank legend entry, the same way it
- * hides an axis title. Domain and range are deduped in lockstep via a Map
- * keyed by field name: d3 dedupes an ordinal scale's domain but keeps its
- * range verbatim, so a measure repeated in the y channel would otherwise
- * shift the indices and map a later measure to the wrong label.
+ * Maps the measure-series legend's domain (raw measure field names) to each
+ * field's `# label`. Labels live in scale data so they render verbatim: an
+ * explicit `# label=""` deliberately shows a blank legend entry, the same
+ * way it hides an axis title. Domain and range dedupe in lockstep because
+ * d3 dedupes an ordinal scale's domain but not its range; a repeated y
+ * measure would otherwise shift later labels onto the wrong measure.
  */
 export function getMeasureSeriesLabelScale(
   explore: NestField,
--- a/packages/malloy-render/src/component/vega/measure-series-label-scale.ts
+++ b/packages/malloy-render/src/component/vega/measure-series-label-scale.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright Contributors to the Malloy project
+ * SPDX-License-Identifier: MIT
+ */
+
+import type {Scale} from 'vega';
+import type {NestField} from '@/data_tree';
+
+export const MEASURE_SERIES_LABEL_SCALE = 'measureSeriesLabel';
+
+/**
+ * For a measure-series legend the color-domain values are the raw measure
+ * field names; this ordinal scale maps them to their `# label` so the legend
+ * entries match the axis and tooltip. Keeping the labels in scale data (not
+ * in a parsed expression) renders any label verbatim. An explicit empty
+ * `# label=""` intentionally renders a blank legend entry, the same way it
+ * hides an axis title. Domain and range are deduped in lockstep via a Map
+ * keyed by field name: d3 dedupes an ordinal scale's domain but keeps its
+ * range verbatim, so a measure repeated in the y channel would otherwise
+ * shift the indices and map a later measure to the wrong label.
+ */
+export function getMeasureSeriesLabelScale(
+  explore: NestField,
+  yFieldPaths: string[]
+): Scale {
+  const labelByName = new Map<string, string>();
+  for (const fieldPath of yFieldPaths) {
+    const field = explore.fieldAt(fieldPath);
+    labelByName.set(field.name, field.getLabel());
+  }
+  return {
+    name: MEASURE_SERIES_LABEL_SCALE,
+    type: 'ordinal',
+    domain: [...labelByName.keys()],
+    range: [...labelByName.values()],
+  };
+}
--- a/packages/malloy-render/src/plugins/bar-chart/bar-chart-plugin.tsx
+++ b/packages/malloy-render/src/plugins/bar-chart/bar-chart-plugin.tsx
@@ -38,12 +38,13 @@ import {
   barChartSettingsSchema,
 } from './bar-chart-settings';
 import {barChartSettingsToTag} from './settings-to-tag';
+import type {SyntheticSeriesField} from '@/plugins/synthetic-series-field';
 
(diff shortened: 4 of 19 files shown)


mtoy-googly-moogly:
A few requests before this lands — the fix itself is right (routing display titles through `getLabel()` while leaving `.name` data refs alone is exactly the correct split, and it's backward-compatible). I'd like the test code up to the bar we hold the rest of the package to first.

**1. Drop the `as unknown as …PluginInstance` casts.** Both spec files build a fake plugin and assert it into the instance type:
```ts
} as unknown as BarChartPluginInstance;
```
A double-cast through `unknown` means the test exercises a hand-built lookalike rather than a real instance, so it can silently drift from the actual type. I realize it's faked because `BarChartPluginFactory.create(field)` drags in solid-js and the layout math — which is also why the two `jest.mock(...)` stubs exist — so this is really a testability question. Could we either (a) drive the real factory, or (b) narrow `generateBarChartVegaSpecV2` to take just the inputs it needs (metadata + settings) instead of the whole plugin, so the test constructs a real, typed value with no cast? Happy to talk through which is less work.

**2. Trim the comments.** A few have grown into paragraphs that restate the code below them — e.g. the 7-line mock-justification header and the `buildBarSpec` preamble ("Compile + run the query, build field metadata…, build the data tree, then drive…"). One line is plenty; the d3-dedup comment is the right kind and can stay.

**3. Factor the duplicated test harness.** The two spec files are near-identical (same mocks, same `buildXSpec`, same three cases with field names swapped). Pulling the shared bits into one small helper keeps them from drifting apart.

One observation, your call: the ~22-line measure-series scale block is verbatim in both `generate-bar_chart-vega-spec.ts` and `generate-line_chart-vega-spec.ts`. The two generators are already parallel twins so I won't block on it, but if a shared helper is easy it'd be the better home.

mtoy-googly-moogly:
Experimentally asked the Fable model to look and it would like to add one question: with `# label=""`, the legend now renders a color swatch with no text next to it. On an axis an empty title reads naturally as "hide the title," but in a legend a naked swatch is ambiguous. Is the blank legend entry the intended UX, or just fallout from "honor the label verbatim"? Fine either way — just want it to be a decision rather than an accident.

mlennie:
@mtoy-googly-moogly Thanks for the careful read. All four points are addressed in the follow-up commit.

**1. Casts dropped, via your option (b).** The generators now take Pick-based input types (`BarChartSpecInputs` / `LineChartSpecInputs`) covering exactly the plugin properties each one reads: `getMetadata`, `field`, `chartDisplay`, `getTopNSeries`, `syntheticSeriesField`, and (bar only) `hasMultipleSeriesFields`. It turned out the generators need a bit more than metadata + settings (the synthetic series field and chart display config among them), so a `Pick` of the instance type was the honest narrowing: the production plugins pass their full instances unchanged, the tests construct a real value the compiler checks with no cast, and if a generator starts reading a new plugin property the Pick has to grow, which breaks the test compile until the test provides it. That drift-proofing felt like the point of your comment, and it was less work than driving the real factory, which would have needed two more module stubs (chart-v2's solid-js and vega-interpreter).

**2. Comments trimmed.** The 7-line mock header is a one-liner pointing at the stubs (which carry the why), the `buildBarSpec` preamble is gone, and the d3-dedup comment survives, now in the shared helper's doc comment.

**3. Harness factored.** The shared bits live in `src/plugins/spec-test-support/`: the two mock stub modules, the query runner, the data, and the three cases parametrized by chart tag. Each spec file is now ~40 lines: the two `jest.mock` calls (which jest's hoisting requires to stay per-file), a typed `buildSpec`, and one `describeChartLabelTests` call.

**4. Scale block shared.** Took your suggestion: the ~22-line block now lives once in `src/component/vega/measure-series-label-scale.ts`, called by both generators, with the scale name exported as a constant so the legend encodes and the tests reference the same string.

On the `# label=""` question: deliberate, now documented as such in the helper's doc comment. The reasoning: an explicit empty label is the only way to ask for a blank entry (omitting the tag already gets you the field name), and we wanted one rule everywhere rather than the same tag meaning "blank" on an axis but "fall back to the name" in a legend. If real-world usage shows the naked swatch confuses more than it helps, a legend-only fallback is a two-line change in the helper, but we'd rather start consistent.

Re-verified after the changes: full malloy-render jest run, tsc, lint, and Storybook screenshots across bar/line/scatter axes, legends (dimensional and measure series, including the blank entry), and tooltips.

mtoy-googly-moogly:
Very close. Codex reports two remaining requests before this lands:

1. Give synthetic series an honest type.
   *  The follow-up removes double casts from the tests, but the production plugins still construct partial Field lookalikes using as unknown as Field. This PR adds getLabel() to both objects because the generators started reading another Field member, demonstrating exactly how these casts can silently drift.
   * Please define a narrow synthetic-series type containing the properties the generators actually consume, and change syntheticSeriesField?: Field accordingly.

2. Exercise the newly changed paths in automated tests.
    * The new tests cover bar/line axes and legends, but explicitly stub custom tooltips and do not exercise default measure tooltips, scatter labels, or either synthetic-series label path.
    *  Since those behaviors are part of this PR, please add focused coverage demonstrating them.

  The implementation otherwise looks consistent, and the fresh main / main CI job has passed.

mlennie:
Both requests are addressed, split into a production commit and a test commit so the delta is easy to see.

**1. Synthetic series have an honest type.** New `plugins/synthetic-series-field.ts` exports `SyntheticSeriesField` with exactly the members the generators read off the series variable: `name`, `getLabel`, `valueSet`, `referenceId`, and an optional `maxString` (never set by the synthetics, but the bar legend-sizing read needs it typed; there's a one-line comment on the member). `syntheticSeriesField?: Field` became `?: SyntheticSeriesField` on both plugin instances, the generators' series locals are `Field | SyntheticSeriesField`, and both literals construct the interface directly with no cast. A nice side effect of "exactly what the generators consume": the six `is*()` predicates the old literals faked turned out to be read nowhere, so they're gone and each literal is now four honest properties.

**2. The changed paths are now exercised.** The suite went from 98 to 106 tests, and I mutation-tested each new assertion (reverting the corresponding production line fails exactly the intended test, and nothing else).

- Default measure tooltips (bar and line): tests call the returned `getTooltipData` directly and assert measure-series entries use the `# label` tags while dimensional-series entries stay series values. One honesty note: the scenegraph items are fabricated by a small harness helper with a documented cast, because real items only exist inside a running vega View and vega 6 is ESM-only, which the node jest project can't load. To keep both tooltip branches per generator covered by the same tests, the duplicated entry-label expression is hoisted into one `entryLabel` closure per generator.
- Custom tooltips: the stub you called out is deleted. The real `get-custom-tooltips-entries` now runs in the tests (only its solid-js renderer dependency is stubbed, and the entry value is a lazy thunk so it never executes), and new bar/line cases assert a `# tooltip`-tagged field's entry label comes from its `# label` tag.
- Scatter: the spec construction moved verbatim out of `renderToDOM` into `generate-scatter_chart-spec.ts` so it's testable without the View; the render path is unchanged. The new test compiles the builder's output with real vega-lite and asserts the compiled x/y axis titles and color/size/shape legend titles, plus negatives for the raw field names. One deliberate three-word change while moving it: the color/size/shape titles now use the documented `legend:` prop rather than the `axis:` prop vega-lite only honored on those channels by accident; the rendered result is the same and the test pins it.
- Synthetic series, both paths: the bar test (two `# series` dimensions) and the line YoY test drive the real plugin factories through `processData`, so the de-cast literals themselves execute; they assert the legend title is the label join ("Quarter - Region") and the synthetic "Year" respectively.

I also ran a security pass over the diff since the labels flow author-controlled strings into the specs: everything lands in escaped text contexts (tooltip labels as JSX text nodes, scatter titles through vega-scenegraph's SVG serializer, verified by rendering a `</text><script>` label and watching it come out escaped), the synthetic name join never reaches a vega field path, and the round-1 data-driven label scale is untouched. It turned up two packaging nits, both fixed: the test-support `.d.ts` files were landing in the npm tarball (one `.npmignore` line), and the scatter test was type-importing from the undeclared transitive `vega-util` (now sourced from `vega`).

Re-verified after all the changes: the malloy-render and malloy-render-test jest projects, tsc, lint, the vite library and Storybook builds, and Storybook screenshots across the label stories plus multi-series and YoY views.

PR #2893, opened 2026-06-16
fix(lang): reject `limit:` in a nested `select:` at compile time

lloydtabb:
## Problem

A nested `select:` (project) that has a `limit:` **compiled successfully but failed at the database at runtime**:

```malloy
view: user_detail is {
  select: user_email is email
  limit: 200
}
run: cohorts -> {
  group_by: cohort
  nest: users_in_cohort is user_detail   -- nested select + limit
}
```

```
Binder Error: Referenced column "user_email__0" not found in FROM clause!
```

### Why

A nested `select:` is compiled by collapsing its rows into a `LIST(...)` aggregate in an early stage. When the nest also has a `limit:`, Malloy emits a `ROW_NUMBER() OVER (... ORDER BY <projected field> ...)` in a later stage to enforce the limit — but the projected field was already swallowed into the `LIST`, so it no longer exists as a column.

Note the asymmetry that made this slippery:
- `nest:` inside a `select:` is **already** caught by the compiler.
- `select:` inside a `nest:` is legal — and works fine **without** a limit.
- A nested *reduce* with a limit works (it row-numbers by real grouping columns).
- A *top-level* select with a limit works (plain SQL `LIMIT`).

Only the nested-project-**with-limit** combination was broken, and whether it errored was shape/dialect dependent — so a runtime/SQL test is a poor guard. This should be a **compile-time** error.

## Fix

Catch it in `NestFieldDeclaration.getFieldDef` — the single chokepoint every nest flows through, whether written inline or as a view reference. If a nest's terminal stage is a `project` segment with a `limit`, log a new `limit-in-nested-select` error:

> Cannot use `limit:` in a nested `select:` query

## Tests

The two previous `refine in nest` tests asserted `.toTranslate()` on exactly this broken construct (they encoded the bug). Updated to expect the error, and added coverage for:
- inline / refined / view-referenced nested select + limit → error
- nested select without a limit → legal
- nested reduce + limit → legal
- top-level select + limit → legal

`parse.spec.ts` passes 90/90. No new regressions in the lang suite. A repo-wide search found no existing test or sample that nests a `select:` with a `limit:`.

> [!NOTE]
> DuckDB db-test verification was not run in the authoring session (suspected wrong Node version producing empty jest output). Worth a quick db-test pass before merge.

### DIFF
--- a/packages/malloy/src/lang/ast/query-properties/nest.ts
+++ b/packages/malloy/src/lang/ast/query-properties/nest.ts
@@ -40,6 +40,22 @@ export class NestFieldDeclaration
           ? pipeline[0].refSummary
           : undefined;
       const checkedPipeline = detectAndRemovePartialStages(pipeline, this);
+      // A nested `select:` (project) is compiled by collapsing its rows into a
+      // `LIST(...)` aggregate. A `limit:` on that terminal stage is implemented
+      // with a `ROW_NUMBER()` ordered by the projected fields, but those fields
+      // no longer exist as columns once they have been listed -- producing
+      // invalid SQL at run time. Reject it at compile time instead.
+      const lastStage = checkedPipeline[checkedPipeline.length - 1];
+      if (
+        lastStage &&
+        model.isProjectSegment(lastStage) &&
+        lastStage.limit !== undefined
+      ) {
+        this.logError(
+          'limit-in-nested-select',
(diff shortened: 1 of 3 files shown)


mtoy-googly-moogly:
Closing, this now "works" #2897

PR #2896, opened 2026-06-16
Add Amazon Redshift dialect and connection

ttarinov:
Adds a native **Amazon Redshift** dialect (`@malloydata/db-redshift` + a `RedshiftDialect` in `@malloydata/malloy`), so Malloy runs against Redshift directly instead of through Postgres federation or a DuckDB pushdown. `RedshiftDialect` extends `PostgresDialect` and overrides only where Redshift actually differs.

  ## Motivation
  Dot ([getdot.ai](https://getdot.ai)) uses Malloy as one of its semantic layers and needed native Redshift support rather than routing through a federation layer, so we built this and wanted it upstream.

  ## Postgres -> Redshift divergences
  - group-set fan-out emits a `UNION ALL` value list instead of `GENERATE_SERIES`, which is leader-node-only on Redshift and fails once compute-node data is involved.
  - `SUM(DISTINCT)` casts to `DECIMAL(38,9)`, not bare `NUMERIC` - bare `NUMERIC` is `NUMERIC(38,0)` here and was silently truncating the fraction to 0. Fixed in the shared Postgres-family path via `defaultDecimalType` so it can't drift Redshift-only.
  - `GREATEST`/`LEAST` ignore `NULL`s where Malloy/BigQuery propagate them - kept as a pinned test rather than worked around.

Nesting, arrays, and records are off for now - Redshift has SUPER/PartiQL, but Malloy's JSONB+UDF nesting pipeline doesn't map onto it, so they're gated with explicit errors until someone builds a SUPER-native path.

  ## Testing
1.  I ran the full `db-all` suite against a real Redshift + S3 cluster: **all green**, with the divergent features skipping through the existing `when()`  gates.

2.  For the experimental → supported question, the guard is an offline SQL-gen suite that runs in normal CI with no AWS - it pins the generated SQL byte-for-byte plus a forbidden-token check, so a base-class change that brings back `GENERATE_SERIES` / `make_interval` / etc fails there without a cluster.

No local Redshift exists (the Docker images are Postgres underneath), so a live CI pass needs a throwaway cluster + S3, same as Databricks.


mtoy-googly-moogly:
Wow ... OK, I am looking at this now.

mtoy-googly-moogly:
OK I am just going to start spewing things one at a time.

redshift absolutely does have TIMESTAMP WITH TIME ZONE

mtoy-googly-moogly:
as an experimental dialect, there is no guarantee that i will not break this from release to release. just last week i changed how dialects work to generate certain kinds of queries. i might want to push on whether we can can move this from experimental to supported, what the work is i do not know yet. do you have a clean run of ci-redshift?

ttarinov:
@mtoy-googly-moogly Now it's green :) I've attached the logs here.

Okay, about the timestamptz you were right, that actually got me to go back through the whole dialect, and I found a few more bits it had inherited from Postgres that Redshift is not supporting (GENERATE_SERIES, decimals truncating, make_interval), fixed as well.

About not breaking it release to release - I built the SQL gen suite that runs in CI without AWS and pins the generated SQL, so it supposed to help.

[redshift-ci-db-all-run.log](https://github.com/user-attachments/files/29254367/redshift-ci-db-all-run.log)

Sha-Bang:
Awesome, really nice work. This handles all the Redshift landmines I'd check for (leader-only `GENERATE_SERIES`, `NUMERIC(38,0)` truncation, no `FILTER`/`STRING_AGG`/`make_interval`, `svv_columns` for Spectrum, the UTC OID parsers). The offline sqlgen suite with the forbidden-token check is a great way to keep an experimental dialect stable.

A few things we may want to consider changing before merging, all in the pooled connection:

1. **`PooledRedshiftConnection` looks like it drops `setupSQL`.** The Postgres pool runs it in its `acquire` handler, but the Redshift `getPool()` override only fires the two SETs, so `setupSQL` (search path, WLM, etc.) gets silently ignored on pooled queries.

2. **The `acquire` SETs are fire-and-forget.** `client.query("SET TIME ZONE 'UTC'")` isn't awaited, so a query dispatched right after acquire can race it. Inherited from the Postgres pool, but it bites harder here since timestamp parsing and `enable_case_sensitive_identifier` both depend on those SETs winning.

3. **`stringLiteralStyle` still says `Doubled`** while `sqlLiteralString` uses backslash escaping. Works today because `sqlLiteralRegexp` routes through `sqlLiteralString`, but anything that keys off the flag directly would emit the wrong escaping. Should be a one-liner.

PR #2906, opened 2026-06-18
Emit single-group-set nested queries without the group_set fan-out

mtoy-googly-moogly:
The nested `select:` work in #2897 left a top-level `group_by` whose only nests are projections on the full group_set fan-out — a cross-join against a one-row group_set table, the `FILTER`/`CASE WHEN group_set=N` demux, and a second combine stage — even though the query has a single grouping grain and needs none of it. Lloyd flagged the generated SQL as more complex than it should be. These queries now emit the single `GROUP BY` a person would write by hand.

The behavior was already correct; this changes only the shape of the SQL. The general fan-out path is untouched — only a query with `maxGroupSet === 0` (no `reduce` nest, ungrouping, or total) takes the new route, where the no-fan-out emitter grew to handle a projection nest as an array-agg.

### DIFF
--- a/packages/malloy/src/dialect/databricks/databricks.ts
+++ b/packages/malloy/src/dialect/databricks/databricks.ts
@@ -33,7 +33,7 @@ import type {
   OrderByRequest,
   QueryInfo,
 } from '../dialect';
-import {Dialect, EscapeStyle, qtz} from '../dialect';
+import {Dialect, EscapeStyle, qtz, turtleGroupSetCondition} from '../dialect';
 import type {DialectFunctionOverloadDef} from '../functions';
 import {expandBlueprintMap, expandOverrideMap} from '../functions';
 import {DATABRICKS_DIALECT_FUNCTIONS} from './dialect_functions';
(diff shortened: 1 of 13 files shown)


mtoy-googly-moogly:
For a concrete sense of the change, this query:

```malloy
run: duckdb.table('state_facts.parquet') -> {
  group_by: f is substr(popular_name,1,1)
  nest: pn is {
    select: popular_name, state
    limit: 3
  }
}
```

**Before** — full group_set fan-out for a single-grain query:

```sql
WITH __stage0 AS (
  SELECT
    group_set,
    SUBSTR(base."popular_name",1,1) as "f__0",
    COALESCE((LIST({
      "popular_name": base."popular_name",
      "state": base."state"}  ORDER BY  base."popular_name" ASC NULLS LAST) FILTER (WHERE group_set=0))[1:3],[]) as "pn__0"
  FROM 'state_facts.parquet' as base
  CROSS JOIN (SELECT UNNEST(GENERATE_SERIES(0,0,1)) as group_set  ) as group_set
  GROUP BY 1,2
)
SELECT
  "f__0" as "f",
  MAX(CASE WHEN group_set=0 THEN "pn__0" END) as "pn"
FROM __stage0
GROUP BY 1
ORDER BY 1 asc NULLS LAST
```

**After**:

```sql
SELECT
   SUBSTR(base."popular_name",1,1) as "f",
   COALESCE((LIST({
    "popular_name": base."popular_name",
    "state": base."state"}  ORDER BY  base."popular_name" ASC NULLS LAST))[1:3],[]) as "pn"
FROM 'state_facts.parquet' as base
GROUP BY 1
ORDER BY 1 asc NULLS LAST
```

One `SELECT`, one `GROUP BY` — no CTE, no `CROSS JOIN`, no `group_set` column, no `FILTER (WHERE group_set=…)`, no combine stage. `limit: 3` is the `[1:3]` array slice; the nest's order rides inside `LIST(… ORDER BY …)`.

PR #2931, opened 2026-06-22
DIAGNOSTIC (do not merge): bigquery 8 + log the unwrapped transport error

mtoy-googly-moogly:
**Diagnostic, not for merge.** The connectors-group bump to `@google-cloud/bigquery@8` fails CI with `failed-to-fetch-table-schema` / "Unexpected Gaxios Error" across the bigquery tests. That string is gaxios-7's wrapper for a **non-Error thrown by the transport**, and it does **not** reproduce locally — every credential state (valid user creds → clean `ApiError`; no creds → clean "could not load credentials"; bogus service-account key → clean `GaxiosError` invalid_grant) produces a clean Error, never the wrapper. So the non-Error only manifests in the CI runner's environment (gaxios 7 → native fetch/undici on the ubuntu runner).

This PR bumps the bigquery stack to 8 and adds a temporary `[bq8-diag]` log in `bigquery_connection.ts`'s error path that dumps the full thrown value, including the wrapped `cause`. Re-run the `db-bigquery` job and grep the log for `[bq8-diag]` to see what's actually being thrown — then we replace this with the real fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4627,55 +4627,305 @@
       }
     },
     "node_modules/@google-cloud/bigquery": {
-      "version": "7.3.0",
-      "resolved": "https://registry.npmjs.org/@google-cloud/bigquery/-/bigquery-7.3.0.tgz",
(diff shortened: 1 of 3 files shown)


mtoy-googly-moogly:
Diagnostic complete — it surfaced the gaxios-7 ESM-under-jest cause. Resolution: defer bigquery 8 (issue #2932, hold PR #2933).

PR #2957, opened 2026-06-29
Fix query-class conflict diagnostics

### LINKED ISSUE #2901: Query-class conflict errors should name the operation, and not double-fire
## Problem

When an operation appears in a query segment of the wrong class, the error names the abstract query **class** instead of the **operation the user actually wrote**, which reads as a non-sequitur.

```malloy
run: a -> { select: astr nest: anest1 is { select: one_n is 1 } }
```
produces:
```
Use of grouping is not allowed in a select query
```
The user wrote `nest:`, not a grouping clause, so "Use of grouping" is confusing. The message comes from `qop-desc.ts` `computeType()` (code `illegal-project-operation`), which only knows the offending element's `forceQueryClass` (`Grouping`) — not that it is specifically a `nest:`.

**Every operation that can trigger this should produce an explicit, parseable message that names the operation and the conflicting context**, e.g.:
```
nest: statements are not legal in a query with select:
```

Operations that route through this generic message:
- `nest:`, `group_by:`, `aggregate:` (`forceQueryClass = Grouping`) inside a `select:` segment
- `select:` (`forceQueryClass = Project`) inside a grouping segment

Model of correct behavior: `having:` carries no `forceQueryClass`, so it is reported once, cleanly, by the builder (`ProjectBuilder.execute` → "Illegal statement in a select query operation").

## Double-error on `group_by:` in a select

```malloy
run: a -> { select: astr group_by: ai }
```
produces **two** errors:
```
Use of grouping is not allowed in a select query     (qop-desc.ts, code: illegal-project-operation)
Illegal statement in a select query operation        (project-builder.ts, code: illegal-operation-in-select-segment)
```
qop-desc's class-conflict check **and** `ProjectBuilder.execute`'s reject list both fire. These should be deduped to a single, clear, operation-named error. (`nest:` does not double-fire only because it is not in the builder's reject list — so it gets just the confusing qop-desc message.)

## Code pointers
- `packages/malloy/src/lang/ast/query-properties/qop-desc.ts` — `computeType()` logs the generic `illegal-${guessType}-operation` conflict.
- `packages/malloy/src/lang/ast/query-builders/project-builder.ts` — `execute()` rejects `having`/`GroupBy` with `illegal-operation-in-select-segment`; `nest` is absent from that list.

## Acceptance
- Each such conflict produces exactly **one** error.
- The message names the offending operation (`nest:`, `group_by:`, `aggregate:`, `select:`) and the conflicting context, in a consistent/parseable form.

## Note
`packages/malloy/src/lang/test/query.spec.ts` has a test ("nest illegal in a select segment") asserting the current `Use of grouping is not allowed in a select query` message; it will need updating when the wording changes.

MianliWang:
Fixes #2901

This PR updates query-class conflict diagnostics so illegal operations in the wrong query segment produce a single operation-specific error.

The change keeps the ProjectBuilder defensive check intact and handles duplicate diagnostic suppression in QOpDesc. I also checked the related forced query-class operations index: and sample:.

This is my first contribution to this repo, so feedback is very welcome.

Tests run:

npx jest --config jest.config.ts --selectProjects malloy-core --runTestsByPath packages/malloy/src/lang/test/query.spec.ts --setupFilesAfterEnv jest-expect-message --runInBand

npx jest --config jest.config.ts --selectProjects malloy-core --runTestsByPath packages/malloy/src/lang/test/parse.spec.ts packages/malloy/src/lang/test/locations.spec.ts  packages/malloy/src/lang/test/syntax-errors.spec.ts --setupFilesAfterEnv jest-expect-message --runInBand

npx eslint packages/malloy/src/lang/ast/query-builders/project-builder.ts packages/malloy/src/lang/ast/query-properties/qop-desc.ts packages/malloy/src/lang/test/locations.spec.ts packages/malloy/src/lang/test/parse.spec.ts packages/malloy/src/lang/test/query.spec.ts packages/malloy/src/lang/test/syntax-errors.spec.ts

npx tsc --build packages/malloy/tsconfig.json --pretty false

git diff --check

### DIFF
--- a/packages/malloy/src/lang/ast/query-builders/project-builder.ts
+++ b/packages/malloy/src/lang/ast/query-builders/project-builder.ts
@@ -12,7 +12,6 @@ import {isPartialSegment, isProjectSegment} from '../../../model/malloy_types';
 
 import {ErrorFactory} from '../error-factory';
 import type {SourceFieldSpace} from '../types/field-space';
-import {GroupBy} from '../query-properties/group-by';
 import {ProjectFieldSpace} from '../field-space/project-field-space';
 import type {QueryProperty} from '../types/query-property';
 import {QuerySegmentBuilder} from './reduce-builder';
@@ -41,7 +40,7 @@ export class ProjectBuilder
   }
 
   execute(qp: QueryProperty): void {
-    if (qp.elementType === 'having' || qp instanceof GroupBy) {
+    if (qp.elementType === 'having') {
       qp.logError(
         'illegal-operation-in-select-segment',
         'Illegal statement in a select query operation'
--- a/packages/malloy/src/lang/ast/query-properties/aggregate.ts
+++ b/packages/malloy/src/lang/ast/query-properties/aggregate.ts
@@ -18,4 +18,5 @@ export class Aggregate
   elementType = 'aggregateList';
   readonly queryRefinementStage = LegalRefinementStage.Single;
   readonly forceQueryClass = QueryClass.Grouping;
+  readonly statement = 'aggregate:';
 }
--- a/packages/malloy/src/lang/ast/query-properties/group-by.ts
+++ b/packages/malloy/src/lang/ast/query-properties/group-by.ts
@@ -18,4 +18,5 @@ export class GroupBy
   elementType = 'groupBy';
   queryRefinementStage = LegalRefinementStage.Single;
   forceQueryClass = QueryClass.Grouping;
+  statement = 'group_by:';
 }
--- a/packages/malloy/src/lang/ast/query-properties/indexing.ts
+++ b/packages/malloy/src/lang/ast/query-properties/indexing.ts
@@ -13,6 +13,7 @@ export class Index extends MalloyElement implements QueryPropertyInterface {
   elementType = 'index';
   weightBy?: FieldName;
   forceQueryClass = QueryClass.Index;
+  statement = 'index:';
   queryRefinementStage = undefined;
 
   constructor(readonly fields: FieldReferences) {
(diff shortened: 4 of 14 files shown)


MianliWang:
It looks like the failing check is the external contributor permission gate. I ran the relevant tests locally and included the commands in the PR description. Happy to address any review feedback.

mtoy-googly-moogly:
Wow thanks, this was more of a "note for future me", thank you for cleaning this up. Two changes please:

1. Rather than the `queryOperationName()` switch on elementType, put the statement name on the operation itself: add a statement field (string | undefined) to QueryPropertyInterface, set it on the ops that force a query class (group_by:, aggregate:, nest:, select:, index:, sample:), and have computeType read el.statement directly. That removes the switch and its fallback.
2. InProjectBuilder.execute, delete the `|| qp instanceof GroupBy` clause and the now-unused GroupBy import — getOp already skips class-conflicting ops before they reach the builder, so that branch can't be hit. 

to test ....

```
(if you haven't done it yet) ...
npm run build-duckdb-db

npm run lint
npm run precheck
```

that runs pretty much all the tests that don't involve external databases, maybe 3 minutes on a fast machine

MianliWang:
Done, thank you for clarifying!

I applied both requested changes:
- moved the statement names onto the query-property operations and now read `el.statement` directly in `computeType`
- removed the unreachable `GroupBy` check/import from `ProjectBuilder.execute`

I also ran the requested checks in a clean Linux Docker checkout:

- `npm run build-duckdb-db`
- `npm run lint`
- `npm run precheck`

All passed.

PR #2963, opened 2026-07-04
Possible fix for #2905: preserve Trino order_by after multi-stage nests

### LINKED ISSUE #2905: Trino: top-level order_by is dropped when the query has a multi-stage nest
## Summary

On a dialect without `SELECT * REPLACE` (Trino), a top-level `order_by` is **silently dropped** whenever the query also contains a **multi-stage nest**. The ordering is emitted onto an inner CTE instead of the outermost SELECT, and Trino (per SQL semantics) does not preserve inner-CTE ordering. Adding an explicit `order_by` does **not** work around it — Malloy isn't putting the `ORDER BY` where it would take effect.

DuckDB is unaffected (it inlines the nest pipeline, so there is no extra outer stage, and single-node execution preserves the order anyway).

## Repro

```malloy
run: trino.table('malloytest.state_facts') -> {
  group_by: f1 is substr(popular_name, 1, 1)
  order_by: f1
  nest: m is { group_by: popular_name } -> { select: popular_name }
}
```

Expected `f1` order: `A E I M O S`. Actual on Trino: `A E O I M S` (arbitrary).

## Verified contrast (Trino)

| query | `f1` sequence | final stage has `ORDER BY`? |
|---|---|---|
| `order_by: f1`, **no** multi-stage nest | `AEIMOS` (sorted) | yes |
| `order_by: f1`, **with** multi-stage nest | `AEOIMS` (unsorted) | no |

## Root cause

The `order_by` is applied in the **combine** stage. For an ordinary query the combine stage *is* the outermost SELECT, so Trino honors it. But a multi-stage nest causes `generatePipelinedStages` (in `packages/malloy/src/model/query_query.ts`) to add **another** outermost SELECT on top of the combine — the carry-forward stage that re-pipes the nest's later stages on non-`supportsSelectReplace` dialects — and that stage does **not** re-emit the `ORDER BY`:

```sql
, __stage1 AS (
    SELECT "f1__0" as "f1", ARRAY_AGG(...) FILTER (WHERE group_set=1) as "m"
    FROM __stage0
    GROUP BY 1
    ORDER BY "f1" ASC          -- order_by lands on this inner CTE
)
SELECT "f1", (… re-pipe of m …) as "m"
FROM __stage1                  -- outermost SELECT, no ORDER BY -> Trino reorders
```

## Scope

- Dialects that reach the carry-forward stage: `supportsSelectReplace = false` (today, **only Trino**) **and** a query with a multi-stage nest **and** a top-level `order_by`.
- The usual Trino mitigation (add an explicit `order_by`) does not help, because the `order_by` is dropped from the final stage.

## Provenance

Pre-existing in `generatePipelinedStages` — that function never re-applied `ORDER BY`. It only became observable once #2899/#2900 unblocked multi-stage nests on Trino at all (before that, the path errored before you could see the ordering). It is **not** introduced by that fix.

## Suggested fix

Have `generatePipelinedStages` re-emit the `ORDER BY` on its output stage, using the carried-forward column names (which include the order key) — the mirror of what the combine stage already does via `genereateSQLOrderBy`. Localized to the non-`REPLACE` path; verify on Trino.

## Notes

Low urgency (no Trino users at present) but a real correctness bug, so recording it. `test/src/databases/all/nested-select.spec.ts` currently works around it: its multi-stage reduce-first test is filtered to a single deterministic row, with a comment pointing here.

MianliWang:
This is a possible fix for #2905.

For Trino/non-`SELECT * REPLACE` dialects, a multi-stage nest adds a final carry-forward stage after the combine stage. The existing top-level `order_by` was applied to the combine stage, but not re-emitted on that final stage, so Trino could return the top-level rows in arbitrary order.

This change passes the generated combine-stage `ORDER BY` through to the final carry-forward stage.

I’m still pretty new to working on issues in this codebase, so feedback is very welcome.

Testing:
- `npx prettier --check packages/malloy/src/model/query_query.ts test/src/databases/all/nested-select.spec.ts`
- `npx tsc --build packages/malloy/tsconfig.json`
- `MALLOY_DATABASE=duckdb npx jest test/src/databases/all/nested-select.spec.ts test/src/databases/all/orderby.spec.ts test/src/databases/all/havinglimit.spec.ts --selectProjects db-all --runInBand`
- Compile-only Trino SQL check for the issue repro and a few nearby cases

Fixes #2905.


test/src/databases/all/nested-select.spec.ts
```diff
@@ -172,12 +171,19 @@ runtimes.runtimeMap.forEach((runtime, databaseName) => {
...
-        // Group 'A' is all "Ava", so the reduce first stage distincts it to one
-        // row and the projection second stage passes that through.
-        await expect(multiStageQuery).toMatchResult(tm, {
-          f1: 'A',
-          m: [{popular_name: 'Ava'}],
-        });
+        if (!runtime.dialect.supportsSelectReplace) {
+          const sql = await runtime.loadQuery(multiStageQuery).getSQL();
+          expect(sql.trim()).toMatch(/ORDER BY\s+"f1"\s+ASC$/);
+        }
```
mtoy-googly-moogly: in general I don't like pattern matching the SQL of a generated query because we might change how we generate SQL. There is a test which actually checks the order and that is suifficent

mtoy-googly-moogly:
Thanks! One change requested ^^^

test/src/databases/all/nested-select.spec.ts
MianliWang: Thanks, that makes sense. I removed the generated-SQL pattern check and kept the result-order regression test.

mtoy-googly-moogly:
New snowflake CI setup, had to merge main into the branch. Will merge this when it CI clears.

PR #2964, opened 2026-07-06
fix(persist): compile PersistSource.getSQL with finalize=false (bare source SELECT)

kylenesbit:
## Summary

`PersistSource.getSQL()` compiled the source's inner query with `finalize=true`, which appends the dialect's result-serialization stage (`Dialect.sqlFinalStage`) whenever `Dialect.hasFinalStage` is true. **Postgres is the only dialect with `hasFinalStage = true`**, and it wraps the last stage in `SELECT row_to_json(finalStage) as row FROM ...`.

For a `#@ persist` source, `getSQL()` feeds two things: the `CREATE TABLE AS (...)` used to build the materialized table, and the `BuildID` (`mkBuildID` over this SQL) used as the manifest key. The serve path resolves a persist source from the **bare (unfinalized) source SELECT**. So on Postgres the `row_to_json` wrapper caused two failures:

1. **Wrong physical table** — `CREATE TABLE AS (getSQL())` materialized a single JSON `row` column instead of real columns. (BigQuery, `hasFinalStage=false`, got real columns and worked.)
2. **Manifest key mismatch** — the build-time BuildID (hashed over the `row_to_json` SQL) didn't equal the serve-time lookup key (hashed over the bare SELECT), so queries fell through to a base-table scan instead of routing to the materialized table.

Compiling with `finalize=false` returns the bare source SELECT, which fixes both: the materialized table has real columns, and the BuildID matches the serve-time key. It is a **no-op for dialects without a final stage** (BigQuery, duckdb, etc.), since `finalize` only adds a stage when `hasFinalStage` is true.

## Changes

- `packages/malloy/src/api/foundation/core.ts` — `PersistSource.getSQL()` now passes `finalize=false` to `compileQuery` (both the `sql_select` and `query_source` branches).
- `packages/malloy/src/lang/test/test-translator.ts` — add a Postgres table (`pgTableDef` / `pg_a`) and `_pg_` connection to the mock harness (Postgres is the only dialect that exercises the final-stage path).
- `packages/malloy/src/api/foundation/persist-source-sql.spec.ts` — new spec asserting `getSQL()` on a Postgres persist source omits the `row_to_json` final stage (and is unaffected for duckdb).

## Test plan

- [ ] `jest --selectProjects malloy-core --testPathPattern persist-source-sql`
- [ ] Confirm no other persist / build-plan snapshots regress
- [ ] End-to-end: a `#@ persist` source on Postgres builds a table with real columns and a query routes to it (verified on a downstream deployment)

> Note: I could not run the full suite in my local checkout due to a pre-existing, unrelated dependency/build state (a parser-generator migration left the `malloy-filter` `dist` requiring `moo`/`nearley`, and the global `test/jest.setup.ts` imports `@malloydata/db-duckdb`). The new spec compiles cleanly and mirrors existing `test-translator`-based specs; CI should exercise it.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/malloy/src/api/foundation/core.ts
+++ b/packages/malloy/src/api/foundation/core.ts
@@ -1757,17 +1757,11 @@ export class PersistSource implements Taggable {
     const sd = this.persistableDef;
     const queryModel = this.model.queryModel;
 
-    // Compile with finalize=false: a persist source's SQL defines the data the
-    // materialized table should physically contain (a bare projection), not a
-    // runnable result set. Finalizing appends the dialect's result-serialization
-    // stage (Dialect.sqlFinalStage) whenever Dialect.hasFinalStage is true —
-    // e.g. Postgres wraps the last stage in `row_to_json(...) as row`. If that
-    // wrapper were included, CREATE TABLE AS would store a single JSON column
-    // instead of real columns, and the BuildID (mkBuildID over this SQL) would
-    // not match the serve-time lookup, which resolves persist sources from the
-    // bare (unfinalized) source SELECT. Both effects break query routing to the
-    // materialized table on dialects with a final stage. No-op where
-    // hasFinalStage is false (e.g. BigQuery/duckdb).
+    // Compile with finalize=false so this SQL is the bare source SELECT.
+    // The build-time key (makeBuildId over this SQL) must equal the serve-time
+    // manifest lookup key, which query_query.ts recomputes from the same
+    // unfinalized SELECT; finalizing would diverge the two on dialects with a
+    // final stage (Postgres) and mis-materialize the table.
     if (sd.type === 'sql_select') {
       return getCompiledSQL(
         sd,
--- a/packages/malloy/src/api/foundation/core.ts
+++ b/packages/malloy/src/api/foundation/core.ts
@@ -1757,14 +1757,25 @@ export class PersistSource implements Taggable {
     const sd = this.persistableDef;
     const queryModel = this.model.queryModel;
 
+    // Compile with finalize=false: a persist source's SQL defines the data the
+    // materialized table should physically contain (a bare projection), not a
+    // runnable result set. Finalizing appends the dialect's result-serialization
+    // stage (Dialect.sqlFinalStage) whenever Dialect.hasFinalStage is true —
+    // e.g. Postgres wraps the last stage in `row_to_json(...) as row`. If that
+    // wrapper were included, CREATE TABLE AS would store a single JSON column
+    // instead of real columns, and the BuildID (mkBuildID over this SQL) would
+    // not match the serve-time lookup, which resolves persist sources from the
+    // bare (unfinalized) source SELECT. Both effects break query routing to the
+    // materialized table on dialects with a final stage. No-op where
+    // hasFinalStage is false (e.g. BigQuery/duckdb).
     if (sd.type === 'sql_select') {
       return getCompiledSQL(
         sd,
         options ?? {},
-        (query, opts) => queryModel.compileQuery(query, opts).sql
+        (query, opts) => queryModel.compileQuery(query, opts, false).sql
       );
     } else {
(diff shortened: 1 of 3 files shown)


mtoy-googly-moogly:
The test as exists is a string comparison of generated SQL.  If we change how we wrap
SQL for postgres, it will silently *pass* even if the bug comes back. Swap
`row_to_json` for `to_jsonb` and the same bug returns, but `not.toMatch(/row_to_json/)`
succeeds: green while broken.

The true invariant is that the id of a persisted source when it is built
matches the id of a persisted source when it is used in a query.

Claude offers this ...

## The test I'd want instead

Assert key-equality through its only observable consequence: **a query that references
the persist source, compiled with a manifest built from `getSQL()`, routes to the
materialized table.** All in-memory, no live DB — same machinery `persist.spec.ts`
already uses:

1. Compile a Postgres persistent source (the pg mock).
2. `buildId = source.makeBuildId(connDigest, source.getSQL())`.
3. Manifest: `{ entries: { [buildId]: { tableName: 'MATERIALIZED_TBL' } } }`.
4. Compile a query using the source with that manifest + `connectionDigests` (same `connDigest`).
5. Assert the generated SQL contains `MATERIALIZED_TBL` (it routed).

On the buggy code the builder hashes finalized SQL, the serve path hashes bare SQL, the
keys differ, the lookup misses, the query inlines the source instead of `SELECT * FROM
MATERIALIZED_TBL` — fails. On the fixed code they match — passes. It never mentions
`row_to_json`, so changing Postgres codegen can't false-green it, and *any* future
divergence between the two paths trips it. Still Postgres-only for the right reason
(build==serve trivially on non-final-stage dialects), still a `malloy-core` unit test.

## Minor

- **Trim the comment.** The 11-line block duplicates the PR body / commit message almost
  verbatim into `core.ts`. Cut it to 2-3 lines that name the invariant (build-key must
  equal serve-key) and point at where serve-time computes it; the mechanism narrative
  belongs in the PR, not the file.
- **`pg_a` is unused.** It's added to `TestTranslator` contents mirroring `bq_a`, but
  nothing references it (the new spec uses `_pg_.table('aTable')`). Drop it. The rest of
  the test-translator change — `pgTableDef` in `mockSchema`, the `_pg_` connection +
  dialect — is needed and follows the bq precedent; keep that.
- The e2e ("verified on a downstream deployment") isn't checkable from the repo; the
  in-repo test should stand on its own, which the routing test does.

Everything else (the two `, false`s, the connection wiring, the naming symmetry) is good.

kylenesbit:
Thanks @mtoy-googly-moogly — this is exactly right, and all three are addressed in 75c0b1e.

**Routing test instead of the SQL string match.** Replaced the `not.toMatch(/row_to_json/)` assertions with the invariant test you described: build a manifest keyed by the source's build-time `BuildID` (`makeBuildId(connDigest, getSQL())`), compile a query that references the source with that manifest + matching `connectionDigests`, and assert the SQL routes to `MATERIALIZED_TBL`. The serve path recomputes its own key from the bare SELECT in `query_query.ts`, so the two keys only agree when build == serve. It never mentions `row_to_json`, so changing the Postgres final-stage codegen can't false-green it, and any future divergence between the paths trips it.

I sanity-checked it the way you'd want: flipping `getSQL()` back to `finalize=true` makes the Postgres test **fail** (the query inlines the `row_to_json(finalStage)` wrapper instead of routing) while the duckdb no-op guard still passes; with the fix in place both pass. Still Postgres-only for the right reason, still a malloy-core unit test.

**Trimmed the comment.** Cut the 11-line block down to 4 lines that name the invariant (build-key must equal serve-key) and point at `query_query.ts` where the serve path recomputes it; the mechanism narrative stays in the PR/commit.

**Dropped `pg_a`.** Removed — nothing referenced it. Kept `pgTableDef` in `mockSchema`, the `_pg_` connection, and its `postgres` dialect registration, which the routing test relies on.

PR #2968, opened 2026-07-07
feat(db-postgres): forward TLS ssl options to pg Client/Pool

### LINKED ISSUE #2960: db-postgres: expose TLS ssl/servername passthrough on PostgresConnectionConfiguration (verified TLS through a tunnel)
### Summary

`@malloydata/db-postgres`'s `PostgresConnectionConfiguration` exposes no way to pass TLS options (`ssl` / `servername`) through to the underlying `pg` `Client`/`Pool`. This makes it impossible to do **verified TLS to a Postgres server reached through a local port-forward / tunnel**, because `pg` verifies the certificate against the host it actually connected to.

### Details

`PostgresConnectionConfiguration` currently accepts only:

```ts
{ host?, port?, username?, password?, databaseName?, connectionString?, setupSQL? }
```

and `getClient()` builds `new Client({ user, password, database, port, host, connectionString })` — no `ssl` field. `pg` then derives its TLS options from the host/env: with a verified `PGSSLMODE` (`require`/`verify-*`), `rejectUnauthorized` is true and `servername` is set to `host` **only when `host` is a DNS name, not an IP** (`pg/lib/connection.js`, guarded by `net.isIP(host) === 0`).

When a caller tunnels Postgres — e.g. an SSH bastion / local forward where the driver connects to `127.0.0.1:<localport>` — the server's certificate is for the *real* database hostname, so verification fails on hostname mismatch (or `servername` is never sent at all, since the host is an IP). The caller has no way to tell `pg` "connect to `127.0.0.1` but verify the cert against `db.example.com`", which `pg` supports natively via `ssl: { servername, ca, ... }`.

### Request

Expose a TLS passthrough on `PostgresConnectionConfiguration` / `PostgresConnectionOptions` — minimally `servername`, ideally the full `ssl` object (`ca`, `rejectUnauthorized`, `servername`, `checkServerIdentity`) — and forward it into the `pg` `Client`/`Pool` config. `pg` already merges a provided `ssl` object straight into its TLS options (`connection.js`: `if (self.ssl !== true) Object.assign(options, self.ssl)`), so this is a small change on the malloy side.

### Use case

Malloy Publisher is adding SSH-bastion / proxied Postgres connections (server opens an in-process local forward; the driver connects to `127.0.0.1`). With a trusted CA in place, full verification through the tunnel needs `ssl.servername = <real DB host>`. Without this passthrough, tunneled connections can only use `sslmode=no-verify` (encrypt, don't verify) rather than proper certificate verification.

Sha-Bang:
## What

`@malloydata/db-postgres` exposed no way to pass TLS options to the underlying `pg` `Client`/`Pool` — `getClient()`/`getPool()` built the config with no `ssl` field, so `pg` derived TLS only from host/env. That made **verified TLS through a tunnel impossible**: when you port-forward Postgres and connect to `127.0.0.1`, the server cert is for the real DB host, so verification fails on hostname mismatch and the only option left is `sslmode=no-verify`.

This adds an `ssl` passthrough, mirroring how `@malloydata/db-trino` already exposes `ssl`. **Backwards compatible: when `ssl` is unset, behavior is byte-for-byte identical to today.**

Closes #2960.

## Design

- **Two-tier typing.** `PostgresConnectionConfiguration.ssl` accepts pg's full `ssl` (`boolean | tls.ConnectionOptions`) for programmatic callers — so `checkServerIdentity`-based verify-ca against a cert with no matching SAN/CN (e.g. Cloud SQL legacy per-instance certs) is possible via the config-reader constructor form. `PostgresConnectionOptions` (saved/registry `json` config) is narrowed to a serializable `PostgresSSLConfig` subset (PEM strings + booleans; `ca`/`crl` accept `string | string[]`), registered as a `type: 'json'` manifest property like Trino's.
- **Shared config builder.** Both `getClient()` and `getPool()` go through `buildClientConfig()`.
- **Reacts to pg's error instead of predicting its host resolution.** pg verifies the cert against the host it connects to, not `ssl.servername` (which it drops for DNS-name hosts). Rather than reproduce pg's host-resolution logic up front, the connector catches pg's terse `ERR_TLS_CERT_ALTNAME_INVALID` at connect time and annotates it with actionable guidance (connect via the DB's IP and set `servername`, or omit `servername`). [Per review feedback — earlier revisions guessed the host ahead of pg, which is fragile.]

## Semantics (pg / Node `tls`, not libpq — documented in the package README)

- `ssl: true` → full verification against the trust store, including hostname (verify-full).
- `ssl: { rejectUnauthorized: false }` → encrypt but do not verify.
- `ssl: { ca }` → trust a specific CA (PEM, or an array for rotation).

## Tests

11 unit tests (pg mocked, no live DB): `ssl` forwarding for `Client` and `Pool`, `ssl: true`, the unset regression guard, the non-`ssl` field mapping, and the certificate-mismatch annotation on both the pooled and non-pooled paths (including that unrelated errors pass through untouched).

### DIFF
--- a/packages/malloy-db-postgres/README.md
+++ b/packages/malloy-db-postgres/README.md
@@ -5,3 +5,32 @@ Malloy is a modern open source language for describing data relationships and tr
 ## This package
 
 This package facilitates using the `malloydata/malloy` library with Postgres - see [here](https://github.com/malloydata/malloy/blob/main/packages/malloy/README.md) for additional information.
+
+## Verified TLS through a tunnel
+
+`PostgresConnectionConfiguration.ssl` is forwarded verbatim to `pg`'s TLS options. To get servername-based certificate verification through a local tunnel (e.g. an SSH bastion forwarding to a remote Postgres), connect via the loopback IP rather than `localhost` — `pg` overwrites `servername` with `host` whenever `host` is a DNS name. If the certificate doesn't match the host `pg` connects to, `pg` throws a certificate error and the connector annotates it with this servername/host guidance.
+
+```ts
+import {PostgresConnection} from '@malloydata/db-postgres';
+
+const connection = new PostgresConnection({
+  name: 'tunneled_postgres',
+  host: '127.0.0.1', // must be an IP, not 'localhost', for servername to take effect
+  port: 5432,
+  ssl: {
+    servername: 'db.example.com',
+    ca: '<PEM-encoded CA certificate>',
+  },
+});
+```
+
+### Semantics (pg/Node `tls`, not libpq)
+
+- `ssl: true` — full verification against the system trust store, hostname included (verify-full). A self-signed or custom-CA server fails unless you supply `ca` (and/or set `servername` for a tunnel).
+- `ssl: { rejectUnauthorized: false }` — encrypt but do **not** verify (libpq's `sslmode=require`/`no-verify`). Use only when verification is impossible; it does not stop MITM.
+- `ssl: { ca }` — trust a specific CA (a PEM string, or an array of PEMs for rotation). This is the path for AWS RDS (or use `NODE_EXTRA_CA_CERTS`) and Cloud SQL.
+- **verify-ca** against a cert whose SAN/CN doesn't match any reachable name (e.g. Cloud SQL *legacy* per-instance certs) needs a `checkServerIdentity` override. That is a function, so it is **not** expressible in a saved `json` config — construct the connection programmatically and pass the full `pg` `ssl` (`tls.ConnectionOptions`) directly.
+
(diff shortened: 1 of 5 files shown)


mtoy-googly-moogly:
At some point we should do a systematic review, per connection type, of what things are legal. There was a lot of pragmatic, feature-cutting going on back in the day.

**1) The `ssl:` block** — Yes. This is where we are headed with other dialects. Eventually I want the connection editor schema to know what is legal in here, but for now this is in line with what other dialects do.

**2) The servername guard — maybe flip it into an error handler.** Right now it tries to predict what pg will do with the host *before* pg runs — which means reproducing pg's own internal host-resolution logic on our side. That logic lives inside the driver and can change, so we don't want to be in the business of guessing it ahead of time.

But once pg actually throws, we could safely infer the cause and say something helpful. Catch the connection error, and turn pg's cryptic cert-mismatch into a clear message that points at the servername/host mismatch and how to fix it.

We can land a smaller PR with just the `ssl:` block, or we can wait for this one to remove the servername guard, or you can talk me into the servername guard.

PR #2976, opened 2026-07-10
Fix date calendar-unit interval measurement

### LINKED ISSUE #2973: Interval measurement in units larger than a day errors: `Cannot measure interval using 'week'/'month'/'quarter'/'year'`
### What happens

Measuring the duration between two dates works for `seconds`, `minutes`, `hours`, and `days`, but the coarser calendar units — `weeks()`, `months()`, `quarters()`, `years()` — all fail to compile with `Cannot measure interval using '<unit>'`.

### To reproduce

Against a DuckDB connection (any source will do; the interval args are literals):

```malloy
// works
run: my_source -> { select: d is days(@2020-01-01 to @2021-06-01) }     // => 517
run: my_source -> { select: h is hours(@2020-01-01 00:00:00 to @2020-01-02 06:00:00) } // => 30

// all error
run: my_source -> { select: w is weeks(@2020-01-01 to @2021-06-01) }    // Cannot measure interval using 'week'
run: my_source -> { select: mo is months(@2020-01-01 to @2021-06-01) }  // Cannot measure interval using 'month'
run: my_source -> { select: q is quarters(@2020-01-01 to @2021-06-01) } // Cannot measure interval using 'quarter'
run: my_source -> { select: y is years(@2020-01-01 to @2021-06-01) }    // Cannot measure interval using 'year'
```

### Expected

Either these units measure the interval (as `days`/`hours` do), or — if measuring in variable-length calendar units is intentionally unsupported — the error message and the functions documentation make that explicit and point users to the workaround (measure in `days` and divide, e.g. `days(a to b) / 7`), since the units are otherwise accepted as truncations (`.week`, `.month`, `.quarter`, `.year`) which makes their absence here surprising.

### Environment

- Observed via **Credible's deployed Malloy runtime**, DuckDB connection, 2026-07. I have not reproduced against a standalone open-source `malloy` build — happy to if that helps narrow it, and can provide the exact compiler version if you can tell me where to read it. The error text matches the Malloy compiler layer, so filing here.
- `seconds` / `minutes` / `hours` / `days` are unaffected.

MianliWang:
### Summary

Fixes #2973.

This enables `week`, `month`, `quarter`, and `year` interval measurement for date-to-date ranges across the supported dialects.

Calendar-unit measurement for timestamps remains unsupported, but the error message is now more explicit.

### Testing

- Targeted runtime tests passed on DuckDB, MySQL, PostgreSQL, and Trino.
- SQL generation was checked for all affected dialects.
- Prettier, ESLint, and `git diff --check` passed.

I was not able to run live tests against BigQuery, Databricks, or Snowflake locally because I do not have credentials for those services.

### DIFF
--- a/packages/malloy/src/dialect/databricks/databricks.ts
+++ b/packages/malloy/src/dialect/databricks/databricks.ts
@@ -528,8 +528,17 @@ export class DatabricksDialect extends Dialect {
   }
 
   sqlMeasureTimeExpr(df: MeasureTimeExpr): string {
-    let lVal = df.kids.left.sql;
-    let rVal = df.kids.right.sql;
+    const from = df.kids.left;
+    const to = df.kids.right;
+    let lVal = from.sql;
+    let rVal = to.sql;
+    if (
+      TD.isDate(from.typeDef) &&
+      TD.isDate(to.typeDef) &&
+      ['week', 'month', 'quarter', 'year'].includes(df.units)
+    ) {
+      return `TIMESTAMPDIFF(${df.units.toUpperCase()}, ${lVal}, ${rVal})`;
+    }
     if (inSeconds[df.units]) {
       lVal = `UNIX_MICROS(CAST(${lVal} AS TIMESTAMP))`;
       rVal = `UNIX_MICROS(CAST(${rVal} AS TIMESTAMP))`;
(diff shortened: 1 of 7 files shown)


mtoy-googly-moogly:
* I think, generally, I approve of the plan to allow more measurement units when looking at date ranges. Timestamps are problematic but restricting this to dates is a great idea.
* I'm looking at the implementation and thinking

PR #3004, opened 2026-07-19
fix(persist): quote manifest table paths per-dialect on the serve path

jswir:
## The bug

A query on a `#@ persist` source over **Snowflake** fails with
`SQL compilation error: Object 'DB.SCHEMA.TABLE__G000__AB12CD34' does not exist or not authorized`
even though the materialized table exists and holds correct data.

The build side CREATEs the physical table with **quoted, case-preserved** segments, so a lowercase-assigned name (`scratch.orders_mz__g000__ab12cd34`) is stored case-sensitively lowercase. The serve side pasted the build-manifest `tableName` into `FROM` **verbatim and unquoted** (`query_query.ts` `query_source` case, and `sql_compiled.ts` `expandPersistableSource`). Snowflake folds unquoted identifiers to UPPERCASE, so the read can never resolve the table. Postgres folds down (matches a lowercase name) and BigQuery doesn't fold, which is why the write/read quoting asymmetry went unnoticed — the two warehouses the existing tests exercise both mask it.

## The fix

- New `Dialect.sqlQuoteTablePath(tablePath)` on the base dialect, composed from the existing primitives: `decodeDottedTablePath` (strips any existing quote layer per segment) + `sqlQuoteIdentifier` per segment, joined with `.`.
  - **Idempotent**: bare (`a.b`), quoted (`\"a\".\"b\"`), or mixed input all come out with exactly one properly-escaped quote layer.
  - **Non-breaking fallback**: input that doesn't parse as dotted identifier segments for the dialect (e.g. a DuckDB file path, which is canonical as a string literal) is returned verbatim — prior behavior preserved.
- Both manifest render sites now call it. Manifest entries are already validated at ingestion (`requireCanonicalTablePathAnyDialect`), so a validated entry always takes the quoting path.
- User-authored `.table()` paths are untouched: they render via the separate, translator-validated `'table'` case, so nothing gets double-quoted.

## Tests

- `dialect/table-path-quoting.spec.ts`: per-dialect quoting (snowflake double quotes, standardsql backticks, postgres doubled-quote escaping), idempotence, mixed-form normalization, hyphenated BigQuery project ids, DuckDB file-path passthrough.
- `api/foundation/persist-source-sql.spec.ts`: adds a snowflake test connection (`_sf_`) and routing assertions pinning the quoted `FROM` on snowflake (the case-folding engine, i.e. the test that would have caught this) and standardsql (backtick byte-compatibility with the build side).
- `test/src/core/persist.spec.ts`: existing routing assertions updated to pin the quoted form.

Verified: `malloy-core` and `db-all`/`db-duckdb`/`db-duckdb-core` jest projects green. Not verified here: a live Snowflake round-trip (the failure was reproduced and root-caused against a production Snowflake deployment; a follow-up will confirm end-to-end once this is consumable downstream).

### DIFF
--- a/packages/malloy/src/api/foundation/persist-source-sql.spec.ts
+++ b/packages/malloy/src/api/foundation/persist-source-sql.spec.ts
@@ -30,33 +30,37 @@ function persistSourceNamed(model: Model, name: string): PersistSource {
   return source;
 }
 
+const CONN_DIGEST = 'test-conn-digest';
+
+// Compile a query on `sourceName` with a manifest whose single entry is keyed
+// by the source's build-time BuildID (makeBuildId over getSQL()) and names
+// `tableName` as the persisted table. The serve path recomputes its own
+// lookup key from the bare source SELECT (query_query.ts); on a key match the
+// FROM becomes the materialized table.
+function routedSQL(
+  model: Model,
+  sourceName: string,
+  connectionName: string,
+  tableName: string
+): string {
+  const source = persistSourceNamed(model, sourceName);
+  const buildId = source.makeBuildId(CONN_DIGEST, source.getSQL());
+  const manifest: BuildManifest = {
+    entries: {[buildId]: {tableName}},
+  };
+  return model.getPreparedQuery().getPreparedResult({
+    buildManifest: manifest,
+    connectionDigests: {[connectionName]: CONN_DIGEST},
+  }).sql;
+}
+
 describe('PersistSource build key matches serve-time manifest lookup', () => {
-  const CONN_DIGEST = 'test-conn-digest';
   const MATERIALIZED = 'MATERIALIZED_TBL';
 
   // Assert the invariant that matters — build-key == serve-key — through its
   // only observable consequence: a query routes to the materialized table.
-  // Build a manifest whose single entry is keyed by the source's build-time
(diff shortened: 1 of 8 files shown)


jswir:
Closing after review — the reviewers' analysis is correct and this is the wrong fix locus.

To answer the central question: we verified our production builder's behavior against live Snowflake query history, and it does CREATE with per-segment quoted, case-preserved identifiers — so the premise held for *our* builder. But the bare-builder regression argument stands: this repo's simple_builder (and malloy-cli build) create bare names, and unconditional serve-side quoting would turn their working bare/bare pairing into a broken bare/quoted one on Snowflake — introducing the exact error this PR set out to fix. #2813's contract ('tableName is canonical SQL, pasted verbatim; a bare name means let the engine fold') is coherent and should stand.

Resolution on our side: the producer will write the *quoted* canonical form into the manifest at bind time, using the same quoting module that quoted the CREATE — write/read symmetry enforced in one codebase, no compiler change needed. The existing manifest validation (requireCanonicalTablePathAnyDialect) already accepts quoted segments, so no upstream change is required.

May follow up with a docs-only PR making the 'quote it yourself for case preservation' contract explicit on BuildManifestEntry.tableName. Thanks for the sharp review — the counterfactual test against the in-repo builder was exactly the right challenge.

PR #3006, opened 2026-07-21
Support natural catalog aliases for shareable DuckDB connections

### LINKED ISSUE #2984: `shareable: true` is incompatible with catalog-qualified DuckDB views
## Summary

I ran into an issue when using the new `shareable: true` option (#2795) together with a DuckDB database generated by SQLMesh.

The problem appears to be related to how `shareable` attaches the database using a fixed alias (`malloy_db`) while existing views reference the original database catalog.

## Environment

* Malloy VS Code Extension: 0.2.1783014968
* SQLMesh: 0.235.4
* DuckDB: 1.5.4

## Configuration

```json
{
  "connections": {
    "duckdb": {
      "is": "duckdb",
      "databasePath": "./mydb.duckdb",
      "shareable": true
    }
  }
}
```

## Expected behavior

The same Malloy model should work regardless of whether `shareable` is enabled.

## Actual behavior

Compilation fails with:

```text
Binder Error: Catalog "mydb" does not exist!
```

Setting

```json
"shareable": false
```

immediately fixes the problem.

## Investigation

After some debugging, I found that SQLMesh creates views like:

```sql
CREATE VIEW marts.t1 AS
SELECT *
FROM mydb.sqlmesh__marts.marts__t1__3443228196;
```

The catalog name (`mydb`) is stored directly in the view definition.

According to the documentation and implementation, `shareable: true` attaches the database as:

```sql
ATTACH 'mydb.duckdb' AS malloy_db;
USE malloy_db.main;
```

As a result, the original catalog (`mydb`) no longer exists, so DuckDB cannot resolve the view.

## Verification

I also verified that this is DuckDB behavior rather than SQLMesh-specific behavior.

DuckDB preserves the SQL used to create a view:

```sql
CREATE VIEW test.v AS
SELECT * FROM test.t;
```

is stored as

```sql
SELECT * FROM test.t;
```

while

```sql
CREATE VIEW test.v2 AS
SELECT * FROM db.test.t;
```

is stored as

```sql
SELECT * FROM db.test.t;
```

So DuckDB does not rewrite or remove catalog qualifiers.

## Interesting observation

If I rename the database file from

```
mydb.duckdb
```

to

```
malloy_db.duckdb
```

everything works correctly.

This seems to confirm that the issue is caused by the catalog alias mismatch rather than the `shareable` mechanism itself.

## Possible improvement

Would it make sense for `shareable` to preserve the original database name instead of always attaching as `malloy_db`?

For example:

```sql
ATTACH 'mydb.duckdb' AS mydb;
```

Alternatively, making the attach alias configurable could improve compatibility with existing DuckDB databases containing catalog-qualified views.

Thanks for adding `shareable`—it's a very useful feature for sharing DuckDB databases between Malloy, SQLMesh, the DuckDB CLI, and other tools.

MianliWang:
## Summary

- add `shareableAttachAlias` with a backward-compatible `malloy_db` default, an `auto` mode that preserves DuckDB's natural catalog identity, and safely quoted explicit aliases
- make shareable DuckDB attachment handoffs lifecycle-safe and coordinate ownership by physical database file
- invalidate schema state across attachment generations and reuse scoped temporary search indexes within a generation
- document the lifecycle and compatibility contracts and add regression coverage

## Root cause

Shareable connections currently attach a DuckDB file using the fixed catalog alias `malloy_db`. Persisted views can retain references to the catalog under which they were created, for example `mydb.schema.table`. `USE malloy_db.main` changes the default catalog and schema, but cannot rewrite those stored catalog qualifiers, so the view fails with `Catalog "mydb" does not exist`.

## Compatibility

The existing `malloy_db` behavior remains the default. Configurations that need the file's natural DuckDB catalog identity can opt in with:

```json
{
  "shareable": true,
  "shareableAttachAlias": "auto"
}
```

Keeping the existing default avoids silently breaking persisted views that already refer to `malloy_db.*`.

## Validation

- `git diff --check`
- `tsc --build packages/malloy packages/malloy-db-duckdb`
- focused catalog identity, lifecycle, cache, concurrency, TEMP-table, and external-lock regression tests

The complete PR CI matrix is still expected to provide the final validation.

Fixes #2984


MianliWang:
The current CI failure appears to occur before any project code is installed, built, or tested.

In [run 29850858599](https://github.com/malloydata/malloy/actions/runs/29850858599):

- `check-permission` reports that the triggering account has read rather than write access, which is the expected gate for a fork PR;
- `pull_and_build` is stopped by `actions/checkout` while attempting to check out the fork head from a `pull_request_target` workflow;
- all downstream build and test jobs are consequently skipped;
- DCO passes.

I have left the implementation and workflow unchanged because this failure is not produced by the PR code, and a workflow change in this fork would not affect the base workflow used by the current `pull_request_target` run.

Could a maintainer please review the change and advise on the preferred way to run CI for this fork PR? This may require running the commits from an upstream-owned branch or adjusting the base workflow before rerunning it. Once the build and tests are able to start, I will follow up on any implementation failures they report.

PR #3007, opened 2026-07-21
feat: per-query database metadata (query tags, job labels) and response query ids

housejester:
## What

Adds optional per-query database metadata to the connector layer, in two directions:

- **`RunSQLOptions.queryMetadata`** — a connector-keyed bag of session metadata (query tags, job labels, client tags, session settings), typed loosely in core and narrowed by each connector.
- **`QueryRunStats.executionMetadata`** — the response-side mirror: the warehouse-assigned execution id of the result (Snowflake query id, BigQuery job id/location, Trino query id), carried on the existing `runStats` channel.

Together these let a caller attach warehouse-side metadata to a query — so operators can attribute cost and separate workload classes in their warehouse's history/billing views — and read the warehouse's own query id back out to join an application call to the query that produced it.

## Per connector

- **snowflake** — `QUERY_TAG` applied per statement (via `parameters.QUERY_TAG`, not the connection-level option, which the SDK clobbers for per-statement params); `getQueryId()` returned as the query id. The connection-level default rides the existing `queryOptions`, so it also covers runtime-internal statements (schema fetches).
- **bigquery** — job `labels` merged into every job (connection default + per-call override); `job.id`/`location` returned.
- **trino / presto** — client tags (`X-Trino-Client-Tags` / `X-Presto-Client-Tags`) and `source` at client construction; query id lifted from the result. Tags containing a comma or CR/LF are dropped so they can't corrupt or inject the header.
- **databricks** — query tags via the dedicated `SET QUERY_TAGS['k1'] = 'v1', QUERY_TAGS['k2'] = 'v2'` grammar, plus generic session settings, at session open (connection-layer). Verified against a live Databricks SQL warehouse.
- **postgres** — `SET application_name` and session GUCs at session open, on both the pooled and non-pooled paths (connection-layer).

No work in `db-mysql` / `db-duckdb`.

## Design notes

- **Additive and optional everywhere** — no behavior change when unset; connectors ignore metadata they don't recognize. The core field is deliberately loosely typed so core stays connector-agnostic; each connector narrows its own block.
- **Two threading mechanisms, matching each connector's existing shape:** statement/job-granular connectors (snowflake, bigquery) carry it on `RunSQLOptions` (like `rowLimit`); session-granular connectors (trino, databricks, postgres) take it as connection configuration applied at session open (like `setupSQL`, following the Postgres `ssl` block precedent).
- **The connector protects the integrity of what it emits** — `SET` keys must be bare identifiers, values are quote-escaped, header-unsafe client tags are dropped. Whether a tag/label/setting is semantically valid for the warehouse is the caller's responsibility; the connector applies what it is given and does not validate warehouse-specific constraints.
- **Session metadata never affects query results or data identity.** All of it — tags, labels, client tags, `application_name`, and session settings — is excluded from the connection digest, so changing it never re-keys the connection. (`setupSQL`, which can change query semantics, stays in the digest as before.)

## Tests

Offline unit tests per connector (`*_query_metadata.spec.ts`) — no live warehouse: they assert the metadata reaches the driver call (statement params / job labels / client headers), that the query id surfaces on `runStats`, plus escaping, digest exclusion, and the header/identifier guards.

Additionally smoke-tested end-to-end against a live instance of every affected engine — confirming the metadata actually attaches (read back from the engine's own history/settings) and the execution id comes back where one exists:
- **Snowflake** — `QUERY_TAG` read back from `QUERY_HISTORY`; `queryId` returned.
- **BigQuery** — job `labels` read back from `INFORMATION_SCHEMA.JOBS`; `jobId`/`location` returned.
- **Databricks** — `SET QUERY_TAGS` grammar + escaping accepted and read back.
- **Trino** — `source` + client tags read back from `system.runtime.queries` / query-info; `queryId` returned.
- **Postgres** — `application_name` + session GUC read back via `current_setting`.

Happy to split this per-connector if smaller PRs are preferred.

### DIFF
--- a/packages/malloy-db-bigquery/src/bigquery_connection.ts
+++ b/packages/malloy-db-bigquery/src/bigquery_connection.ts
@@ -25,6 +25,7 @@ import type {
   QueryRecord,
   QueryOptionsReader,
   QueryRunStats,
+  QueryExecutionMetadata,
   RunSQLOptions,
   StreamingConnection,
   TableSourceDef,
@@ -84,6 +85,30 @@ interface BigQueryConnectionOptions extends ConnectionConfig {
   client_email?: string;
   private_key?: string;
   setupSQL?: string;
+}
+
+/** The BigQuery block of `RunSQLOptions.queryMetadata`. */
+interface BigQueryQueryMetadata {
+  jobLabels?: Record<string, string>;
+}
+
+/** Read the BigQuery job labels out of a query's metadata, if present. */
+function bigQueryJobLabels(
+  options?: RunSQLOptions
+): Record<string, string> | undefined {
+  const block = options?.queryMetadata?.bigquery as
+    BigQueryQueryMetadata | undefined;
+  const labels = block?.jobLabels;
+  return labels && typeof labels === 'object' ? labels : undefined;
+}
+
+/** Merge two label maps (the second wins per key); undefined if both empty. */
+function mergeLabels(
+  base?: Record<string, string>,
+  override?: Record<string, string>
+): Record<string, string> | undefined {
+  if (!base && !override) return undefined;
+  return {...base, ...override};
 }
 
 interface SchemaInfo {
@@ -250,14 +275,18 @@ export class BigQueryConnection
 
   private async _runSQL(
     sqlCommand: string,
-    {rowLimit, abortSignal}: RunSQLOptions = {},
+    options: RunSQLOptions = {},
     rowIndex = 0
   ): Promise<{
     data: MalloyQueryData;
     schema: bigquery.ITableFieldSchema | undefined;
   }> {
+    const {rowLimit, abortSignal} = options;
     const defaultOptions = this.readQueryOptions();
     const pageSize = rowLimit ?? defaultOptions.rowLimit;
+    // Per-call labels; the connection-default labels are merged in
+    // createBigQueryJob so they also cover runtime-internal jobs.
+    const perCallLabels = bigQueryJobLabels(options);
 
     try {
       const queryResultsOptions: QueryResultsOptions = {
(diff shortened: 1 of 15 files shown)


mtoy-googly-moogly:
Generally looks good. I took a quick look and Claude had some issues. A few are real bugs; a couple are code-cleanliness that I'd want tightened before it lands.

1. Cross-chart X highlighting doesn't actually fire. The x_facet_values aggregate only keeps v (fields:['x'], ops:['values'], as:['v']), so datum.x is undefined on the x_highlight band. The outgoing brush ends up [undefined], and the band never lights from a sibling's brush either. Bar chart re-materializes x with fields:['x','x'], ops:['values','min'], as:['v','x']. Looks fine on a lone chart (self-hover matches, tooltip reads v[0].x) — which is why storybook missed it. One-line fix, but the PR notes say cross-chart X highlighting works, and it doesn't yet.
2. Axis-sizing math is duplicated. measureAxisMetrics() is essentially the primary-axis label-measuring block, but the primary axis still runs its own inline copy — so there are two, free to drift. And the comment claims "without duplicating the label-measuring math," which isn't true. Route the primary axis through the same helper so there's one copy and the comment is honest.
3. Null time-x shows "Invalid Date" in the tooltip. The title does new Date(first.x) with no null guard; line chart guards it (x === NULL_SYMBOL ? NULL_SYMBOL : …). Copy that.
4. limit is parsed but never applied — high-cardinality x renders every band, no cap, no "showing N of M" note. Bar/line cap it.
5. A measure listed both ways gets doubled — y=reach where reach also has # y pushes it twice → duplicate series + legend entry.
6. bar-on-both-axes overlaps — { y.chart=bar y2.chart=bar } draws both bar groups on top of each other. Reject it or document it as unsupported.
7. Legacy # combo_chart on a scalar gets the red tile but no clickable source location — combo_chart isn't in the nestOnly list (# viz=combo is fine).
8. hideReferences is computed but never read (dead field), and spark-size combos don't disable nice on the scales the way line does.

Also Claude says these words which me, not being a charting expert, can only parrot ...

One question, not a fix: what made you pick independent auto-scaled axes as the default? That's the reading people warn about — the crossover being an artifact of the two scales, not the data. Curious whether it was deliberate, and whether the default should lean somewhere harder to misread.

PR #3008, opened 2026-07-22
feat(render): add combo_chart dual-axis chart type

jswir:
## What

Adds a first-class **`# combo_chart`** dual-axis chart type to `malloy-render`: two measures on two independent y-axes (the classic bars + line), for comparing measures with different units or magnitudes — volume vs. rate, count vs. dollars.

Today `bar_chart`/`line_chart` accept multiple measures but fold them all onto **one shared y-scale**, which is useless when the measures differ wildly in scale. There is no secondary axis anywhere in the renderer. This adds one.

## API

`viz=combo` (legacy `# combo_chart`), built on the existing channel vocabulary:

```malloy
# combo_chart { x=network y=hh_reach y2=conversion_rate }   // bars(left) + line(right), the defaults
# combo_chart { y.chart=line y2.chart=bar }                 // swap: line-over-bar
# combo_chart { y=['reach','impressions'] y2=['cpa','cpm'] }// arrays per axis
# combo_chart { y2.line_width=3 y2.points=false }           // line styling
# combo_chart { y.min=0 y2.min=0 y2.max=100 }               // pin axis bounds
```

- **Axis = which channel**: `y` → left, `y2` → right (evidence.dev's model; maps onto Malloy's x/y/series channels).
- **Mark type per channel** via `y.chart` / `y2.chart` (`bar`|`line`). Smart defaults: `y`→bars, `y2`→line; a lone pair of measures auto-assigns to the two axes, so `# combo_chart` alone works.
- **Line styling** (line channels only): `y.line_width`/`y2.line_width` (stroke px, default 2); `y.points`/`y2.points` force dot visibility, unset = auto (dots hidden once a series has >1 point, matching `# line_chart`).
- **Axis pinning**: `y.min`/`y.max`/`y2.min`/`y2.max` pin either end of an axis (the other end stays data-driven). Pinning both axes to comparable ranges makes the bar/line crossover meaningful.
- Each axis **auto-scales independently**; only the left axis draws gridlines (the two tick sets don't align — standard dual-axis convention).

## Design notes

- **Self-contained generator, minimal shared extraction.** `createMeasureAxis` gains `orient`/`scaleName` (defaults unchanged → bar/line byte-identical); `chart-layout-settings` gains an optional secondary axis + right-padding reservation. No third copy of mark logic.
- **One place to register a viz type.** Routed the validator's allow-list and dispatch through the shared `VIZ_CHART_TYPES` constant, so a future chart type is valid without editing three lists.

## Update (post-review)

Second commit addresses review + real-render feedback:

- **Hover fixed.** The tooltip fired only in the gaps between bars, never on top of a bar/point — the data marks drew above the `x_highlight` target and swallowed pointer events. Bars/lines/points are now `interactive: false`, so the full-band highlight rect drives the tooltip over any mark.
- **Dual-axis scaling fixed for `size=spark` / `.independent`.** The y-scale domain fallback referenced `yLeft`/`yRight` fields that don't exist on the `values` dataset, collapsing the affected scale. It now reads the folded `left_values`/`right_values` `y` field, with `domainMin`/`domainMax` honoring any pin regardless of domain source.
- **Line styling + axis pinning** added (see API above); dots now auto-hide for multi-point lines like `# line_chart`.
- **Nits:** secondary axis now gets the same top-tick overflow correction as the primary; the embedded-tag chart check routes through `VIZ_CHART_TYPES`; invalid `y/y2.chart` warns instead of silently falling back; `extractFieldName` factored into one shared helper (was copy-pasted across bar/line/combo); mark-literal `as` casts replaced with typed consts.

## Scope (v1)

Out of scope, noted for follow-up: dimensional `series` split *combined* with dual-axis; 3+ axes; per-individual-measure mark types within one axis; measure-hover reference-line brushing on the two axes (cross-chart X highlighting works).

## Testing

- **Spec generation** (`generate-combo_chart-vega-spec.spec.ts`): dual scales, left/right axes, smart-default bar-left/line-right, mark swapping, single-gridline, `# label` titles, tooltip, hover non-interactivity, line `line_width`/`points` (incl. auto-hide), `y/y2.min/max` pinning, single-measure throw, invalid-`chart` fallback, settings round-trips.
- **Domain fallback** (`generate-combo_chart-vega-spec.fallback.spec.ts`, new): reproduces the `size=spark` / `.independent` cases the shared stub can't reach and asserts the fallback reads the folded datasets, not absent `values` fields.
- **Dispatch/settings** (`render-field-metadata.spec.ts`): `# combo_chart` & `# viz=combo` → chart, auto-assignment, explicit channels, invalid-`chart` fallback, and a no-"invalid viz type"-log guard.
- **Real Vega:** jest can't load Vega in the node env (spec tests mock it), so parse/run correctness is verified in Storybook (11 scenarios: defaults, swapped marks, explicit channels, large, styled line, forced points, independent axes, spark, pinned axes, dashboard tile, and the error case).

## Notes for reviewers

- The right-axis title currently uses Vega's default (rotated) placement; matching the left axis's horizontal title is a small follow-up.
- Measure-hover y-axis reference-line brushing is off on both axes in v1 (`showBrushes: false`); wiring it for two independent scales is deferred.
- A new viz type still has to be listed in `data_tree/utils`'s render-tag list; the validator and the embedded-tag chart check now route through `VIZ_CHART_TYPES`, but that one list isn't unified yet.

## Demo

[Loom walkthrough](https://www.loom.com/share/b6c084dfc243449ea6f2dc202f993667)


mtoy-googly-moogly:
Generally looks good. I took a quick look and Claude had some issues. A few are real bugs; a couple are code-cleanliness that I'd want tightened before it lands.

1. Cross-chart X highlighting doesn't actually fire. The x_facet_values aggregate only keeps v (fields:['x'], ops:['values'], as:['v']), so datum.x is undefined on the x_highlight band. The outgoing brush ends up [undefined], and the band never lights from a sibling's brush either. Bar chart re-materializes x with fields:['x','x'], ops:['values','min'], as:['v','x']. Looks fine on a lone chart (self-hover matches, tooltip reads v[0].x) — which is why storybook missed it. One-line fix, but the PR notes say cross-chart X highlighting works, and it doesn't yet.
2. Axis-sizing math is duplicated. measureAxisMetrics() is essentially the primary-axis label-measuring block, but the primary axis still runs its own inline copy — so there are two, free to drift. And the comment claims "without duplicating the label-measuring math," which isn't true. Route the primary axis through the same helper so there's one copy and the comment is honest.
3. Null time-x shows "Invalid Date" in the tooltip. The title does new Date(first.x) with no null guard; line chart guards it (x === NULL_SYMBOL ? NULL_SYMBOL : …). Copy that.
4. limit is parsed but never applied — high-cardinality x renders every band, no cap, no "showing N of M" note. Bar/line cap it.
5. A measure listed both ways gets doubled — y=reach where reach also has # y pushes it twice → duplicate series + legend entry.
6. bar-on-both-axes overlaps — { y.chart=bar y2.chart=bar } draws both bar groups on top of each other. Reject it or document it as unsupported.
7. Legacy # combo_chart on a scalar gets the red tile but no clickable source location — combo_chart isn't in the nestOnly list (# viz=combo is fine).
8. hideReferences is computed but never read (dead field), and spark-size combos don't disable nice on the scales the way line does.

Also Claude says these words which me, not being a charting expert, can only parrot ...

One question, not a fix: what made you pick independent auto-scaled axes as the default? That's the reading people warn about — the crossover being an artifact of the two scales, not the data. Curious whether it was deliberate, and whether the default should lean somewhere harder to misread.

jswir:
Thanks for the thorough pass — all eight addressed in `260dcd1bf`, plus a change for the axis-default question.

**Bugs**

1. **Cross-chart X highlight** — fixed. `x_facet_values` now re-materializes `x` (`fields:['x','x'], ops:['values','min'], as:['v','x']`), so `datum.x` is defined for both the outgoing brush and the sibling-lit `indexof(brushXIn, datum.x)`. You were right the PR note was wrong; it works now (verified across two combo tiles in a dashboard). Spec test asserts the aggregate.
3. **Null time-x → "Invalid Date"** — fixed. Guarded with `NULL_SYMBOL` before `new Date()`, matching line_chart. Test uses a genuine time x so it exercises the date branch.
4. **`limit` never applied** — fixed. A numeric `x.limit` caps the bands directly; unset auto-fits to plot width (budgeting per band by bar count, like bar chart). Dropped rows now set `isDataLimited` + a "Showing N of M" note.
5. **Doubled measure** — fixed. Channels are deduped after all sources (explicit + embedded + auto); a measure on both axes stays on the left. Test covers `y=m` alongside a `# y` tag.

**Cleanups**

2. **Duplicated axis math** — fixed. The primary axis now routes through the same `measureAxisMetrics()` helper as the secondary, so there's one copy; the false "without duplicating…" comment is gone. Bar/line snapshots unchanged confirm it's behavior-preserving.
7. **Legacy `# combo_chart` on a scalar** — fixed. `combo_chart` added to `nestOnly`, so it gets the source-located "requires a nested query" error instead of a bare red tile.
8. **Dead `hideReferences` + spark `nice`** — fixed. `hideReferences` removed from the combo settings/schema; spark now sets `nice:false` on both y-scales like line_chart.

**6. Bar-on-both-axes** — rather than reject it, they now render **side-by-side**: a single shared `xOffset` band scale over the union of both axes' bar measures gives each bar its own slot (each keeps its own y-scale). It collapses to the previous layout when only one axis draws bars, so nothing else changes. New `bars_both_axes` story.

**On the default (your not-a-fix question)** — deliberate, and I think it's the right default: independent auto-scale is the standard dual-axis convention (evidence.dev, Excel, Vega-Lite) and it's the only way to put measures of different magnitudes on one chart at all. Zero-basing both, or a shared scale, would either flatten a rate measure or defeat the point of a combo. But you're right the crossover can mislead, so I've added a visual mitigation: **each axis is now tinted to its mark's color** (title, ticks, axis line) so the two scales read as two separate rulers rather than one shared one — value labels stay neutral for legibility, and `color_axes=false` opts out. The docs now call out that the crossover is a scale artifact, and `y.min`/`y.max`/`y2.min`/`y2.max` (which pin either axis, incl. the left) remain the way to make an exact crossover meaningful.

189 render tests green; new tests cover 1/3/4/5/6/8 and the tint. Storybook has scenarios for all of the above.

mtoy-googly-moogly:
Thanks, Claude gives your changes 👍 

I need two things.

1) This PR is on the other side of a CI process change, so I need you to merge main onto the branch so CI can run.
2) The renderer doesn't really have CI, other than "it compiles and some unit tests pass". Claude is uncertain about these these changes enough that it is requesting a visual fidelity check.
  * Pull a new work tree from origin/main and build and run the renderer storybook
  * Build and run the renderer storybook in your tree
  * Check the following stories
  
  ```
  Story names below are Storybook's start-cased export names, so the sidebar may differ by a space here or there; the panel path is what to navigate by. Ordered by yield.

Malloy Next/Size Sheet pair — check these two first
- Bar Chart / Size Sheet
- Line Charts / Size Sheet

Every size preset on one screen. chartHeight drives ceil(height/40) tick count and whether the top-overflow correction fires, so if measureAxisMetrics drifted at all, one row of this sheet will disagree. Highest information per screenshot in the whole suite.

Wide y labels → left padding (padding.left)
- Bar Chart / Root Chart — topSellingBrands measures Sales $ as retail_price.sum() * 500, so the y labels are the widest in the story set. This is the baseline for horizontal drift.
- Bar Chart / Root Chart Preset Size — same data at # size=md.
- Line Charts / Dates — currency-formatted measures on a time x.

Y-axis title placement (the titleAlign spot)
- Bar Chart / With Titles — y-axis title competing with a chart title and subtitle; if the title moved, it shows here first.
- Bar Chart / With Viz Tag — same layout via # viz=bar, so it also confirms the dispatch change didn't reroute anything.

Ragged maxima → top clipping and gridline count
- Bar Chart / Lots Of Bars
- Bar Chart / Numeric Dimensions As Y
- Line Charts / Lots Of Lines
- Line Charts / Numeric Dimensions As Y

Look at whether the tallest bar/point still clears the top edge by the same margin, and count the gridlines.

The hasYAxis === false branch (spark)
- Bar Chart / Sparks, Bar Chart / Sparks Nested
- Line Charts / Sparks, Line Charts / Sparks Nested

This is the branch where yDomain is now computed separately rather than falling out of the shared scale.

Independent-domain path (yScale.domain: null)
- Bar Chart / Indepenent Axis (the typo is in the source)
- Line Charts / Indepenent Axis
- Line Chart Defaults / Default Independent Axes

The titlePadding / labelPadding cascade — the branchiest code, and almost nothing else reaches it
- Vega Config Override / Bar Chart
- Vega Config Override / Line Chart With New Config
- Vega Config Override / Bar Chart With New Config
- Vega Config Override / Legacy Line Chart
- Themes / Model Theme Override
- Themes / View Theme Override

Most stories leave axisY.titlePadding, axis.titlePadding, and title.offset undefined and never enter that cascade at all. These six are the only ones that do, which makes them disproportionately worth the look despite being the least glamorous.

Padding drift compounding at tile size
- Dashboard / Dashboard Flex
- Dashboard / Dashboard Dense Flex

A few pixels of left padding is invisible on a full-width chart and obvious when eight tiles have to line up.

That's 23 stories rather than 62. One story worth adding to the pass even though it isn't about layout: Bar Chart / Misconfigurations, which is the only existing story that exercises the validation paths in render-field-metadata.ts — the y-channel numeric check now runs for y2 on every chart type, not just combo, so a bar chart carrying a stray y2 tag would newly log an error there.
```

maintainable visual fidelity CI is on my list for renderer projects -- someday when we have more people.

**Comment by jswir** [comment]
Both done, plus a real bug your Misconfigurations note led me to.

## 1. main merged

`3ab32fc9`. CI runs now — `pull_and_build` was dying in 4s on `Refusing to check out fork
pull request code from a 'pull_request_target' workflow`, which the Core-job rework on main
fixes. `pull_and_build`, `main / main`, DCO, duckdb, duckdb-wasm, postgres, publisher and
consumer-canary all green.

Two reds, neither from this branch:

- `check-permission` — the fork-PR gate, needs a maintainer.
- `db-mysql` — a pre-existing race in `test/mysql/mysql_start.sh`, unrelated to this branch.
  **Could you re-run it? I don't have permission to.** It should pass on a retry.

I chased this one down rather than just calling it flaky. The wait loop greps for
`mysqld: ready for connections`, but mysql:8.4.2 prints that line twice, and the first one is
the temporary server the entrypoint uses to initialize:

```
/usr/sbin/mysqld: ready for connections. ... port: 0      <- init server, unix socket only
/usr/sbin/mysqld: ready for connections. ... port: 3306   <- the real server
```

`port: 0` means no TCP listener. The script matches it, leaves the wait loop, and the next
command is `docker exec ... mysql -P3306 -h127.0.0.1`, which needs TCP — hence
`ERROR 2003 (HY000) ... (111)`.

Measured locally over three container starts, the gap between those two lines is 1.8–2.2s, and
the script polls every 10s — so it fails only when a poll lands in that window, roughly one run
in five. The CI log matches exactly: container up at 16:02:27.2, one `...` (so a single
`sleep 10`), poll at 16:02:37.7 catching the `port: 0` line, failure at 16:02:38.3. A faster or
slower runner passes.

Three things confirm it isn't this branch: it dies inside `mysql_start.sh` before
`npm run ci-mysql` executes, the same script passed on main on Aug 20, and a renderer-only diff
can't affect how fast mysqld boots.

Worth a real fix, separately — the log scrape should be a TCP readiness check
(`mysqladmin ping --protocol=TCP`) in both wait loops, and while in there, that loop's timeout
is `counter -eq 120` with `sleep 10`, i.e. 20 minutes rather than the 2 the comment claims, with
a dead `break` after the `exit 1`. Happy to send that as its own PR off main if you want it.

## 2. Visual fidelity check — no differences

Two passes: I ran your list mechanically, and then went through every story by eye in a live
storybook.

**By eye:** all 27 of your stories plus the 13 combo stories. Nothing off — bar and line are
unchanged, and the combo stories render as intended.

**Mechanically:** built storybook from a clean `origin/main` worktree and from this branch,
screenshotted your list headless at 1280x900, and pixel-diffed. Your list enumerates 27 stories
rather than 23 (I counted the groups out), and `Bar Chart / Misconfigurations` is included.

**All 27 byte-identical.** Not "close" — the PNGs compare equal byte for byte.

Because a diff that finds nothing proves nothing on its own, I checked the harness can see:
changed `padding.right` from 8 to 9 in the main worktree, rebuilt, re-shot, re-diffed. 9 of the
27 lit up (`Indepenent Axis`, `With Viz Tag`, `Numeric Dimensions As Y`, both Vega Config
Override line charts, ...), from 192 px to 91,025 px. So the pipeline detects a one-pixel
padding change.

Worth knowing: the other 18 stayed byte-identical under that mutation — spark uses fixed
padding, and preset/fill charts absorb a spare pixel on the right. So screenshots alone would
miss some layout drift. That's why I also did the next thing.

### Numeric proof for the axis math

Screenshots sample; I wanted the layout function itself pinned. `getChartLayoutSettings` is
pure, so I bundled main's copy and this branch's copy side by side with real vega scale math
(jest can't — vega 6 is ESM-only) and one shared text-measurement stub, then compared
**80,640 input combinations**: every size preset including spark and fill, 14 domains, 10
chart heights, the whole `axisY.titlePadding` / `axis.titlePadding` / `title.offset` /
`labelPadding` cascade, independent-Y both ways.

Identical in every one. Same mutation check: changing `ceil(chartHeight / 40)` to `/ 41`
produces 4,928 differences, so the run isn't vacuous. The top-tick-overflow branch fires in
229 of 1,260 probe cases, so `measureAxisMetrics` is genuinely exercised — that was your
Size Sheet concern.

`measure-axis.ts` I argued rather than measured: `orient` defaults to `'left'` and `scaleName`
to `'yscale'`, the template literal reproduces the old `invert("yscale", ...)` string exactly,
and `titleAlign` is set only when `isRight`. The left axis every existing chart uses is
untouched — which the `With Titles` and `With Viz Tag` screenshots then confirm.

## 3. A real bug — your Misconfigurations note was well aimed

You were right to flag the y2 check widening, and it's worse than y2. To be clear on provenance:
I found this by reading that validation block after your note pointed at it, then confirming
with a test — not by spotting it on screen. The errors land in the logs panel, not the chart.

**The combo chart's per-axis validation fires on every chart type**, because it sits in the
shared `viz` block that all of them pass through, ungated. All four of these are actual output,
not predictions:

```
# bar_chart { y.min=100 y.max=0 }  -> error: Invalid y axis bounds on 'n': min (100) must be
                                     less than max (0). Fix: # combo_chart { y.min=0 y.max=100 }.
# viz=table { y.chart=bogus }      -> warn:  Unknown y.chart 'bogus'. Valid types: bar, line.
# bar_chart { y2=str }             -> error: Chart y2-channel field 'str' must be numeric...
```

A bar chart being told to write `# combo_chart` is the bad one. Bar and line have no
`y.min`/`y.max`/`y.chart` at all and no y2 channel, so a stray `y2` there is an unknown tag —
already reported by the `ownedPaths` check — not a channel type error.

Fixed in `d0b858b0` by gating on the viz type actually being combo. 9 tests pin that each case
no longer fires for bar/line/table and still fires for combo.

## 4. The layout test couldn't reach the branch it was testing

`chart-layout-settings.spec.ts` stubs vega with domain `[0,1]` and `.ticks() -> [0,1]`. The top
tick therefore always equals the domain max, so `topTickOverflowExtra` returns 0 every time and
the top-overflow correction never runs — the one branch the `measureAxisMetrics` extraction
actually moved.

Made the stub configurable and added 5 tests: no overflow, overflow borrowing from top padding
and pinning the tick count, the clamp at zero, and the dual-axis case where the larger of the
two corrections is applied exactly once. Both obvious mutations (dropping y2 from the `max`,
and double-counting the two) now fail the suite.

## 5. On dropping bar_chart/line_chart from RENDER_TAG_LIST

The change is fine but my commit message's reasoning was wrong. I said all chart tags normalize
to `viz=...` before that list is reached. They don't always — `convertLegacyToVizTag` returns
early when `viz` is already present, so the list *was* still reachable:

| tag | main | branch |
|---|---|---|
| `# viz=bogus bar_chart` | `chart` | `table` |
| `# viz bar_chart` | `chart` | `table` |

Both are malformed tags that already log an "Invalid viz type" error, and falling back to a
table beats rendering a chart with no plugin, so I think the new behavior is better — but it's
a behavior change, not a no-op, and it deserved to be described as one.

## Verification

212/212 render tests, `tsc --noEmit` clean, lint clean, and the three CI gates
(`ci-test-sanity-check.sh`, `ci-env-sanity-check.sh`, `ci-permission-sanity-check.js`) pass.
The three new combo spec files are correctly claimed by the new jest projects.

What is and isn't machine-checked: everything above proves I didn't break bar and line. The
combo chart's own appearance — colored axes, side-by-side bars, axis pinning — is human-verified
only, since there's no baseline to diff a new chart type against. The right-axis title is still
on Vega's rotated default rather than matching the left axis's horizontal placement; that's the
follow-up already noted in the PR body, and I'd rather do it separately than grow this diff.

Happy to record a walkthrough if you want to see the hover fix and cross-chart X highlighting
in motion — that's the part neither screenshots nor unit tests can show. Say the word.

mtoy-googly-moogly:
I am not the judge of what a good combo chart is, assuming that if you submitted it and it passed ci it is done.

PR #3011, opened 2026-07-23
feat: per-query database metadata (query tags) and response execution ids

housejester:
## Problem

Every query a Malloy connection issues arrives at the warehouse indistinguishable from any other. There's no way to attach cost-attribution, workload-classification, or tracing metadata to those queries so it shows up in the warehouse's own query history (Snowflake `QUERY_TAG` / `QUERY_HISTORY`, BigQuery job labels / `INFORMATION_SCHEMA.JOBS`, Trino `system.runtime.queries`, etc.). Malloy currently can't drive any of that.

## What this adds

One small, additive, optional field on the existing options channel:

- **`RunSQLOptions.queryMetadata?: Record<string, string>`** — a dialect-agnostic bag of string properties, supplied **per query**. Unset means no behavior change.

That's the whole interface. There is no response-side addition (see "No read-back", below).

## Applied per query, natively where possible

Because the metadata describes a single query, each connector applies it only through a mechanism that can attach **per query**:

| Connector | Mechanism |
|---|---|
| snowflake | per-statement `QUERY_TAG` (JSON of the bag) |
| bigquery | per-job `labels` (transformed to BigQuery's label grammar) |
| trino / presto | leading SQL comment |
| databricks | leading SQL comment |
| postgres | leading SQL comment |
| duckdb / mysql | leading SQL comment |

A connector with no per-query-native tag mechanism prepends the bag as a single leading comment via a shared helper:

```
-- team="finance" env="prod"
SELECT ...
```

**Session/connection-scoped tag mechanisms are deliberately not used** — Trino client tags are fixed at client construction, Databricks `SET QUERY_TAGS` is session state, Postgres `application_name` is a session GUC. Each applies to *every* query on the connection, so using it for per-query metadata would race any other query sharing that connection (or add a set/reset round-trip per query and assume exclusive checkout). The leading comment is reliably per-query, costs no extra round-trip, and is visible in the same places (`pg_stat_activity.query`, `system.runtime.queries`, query history).

## The contract

Core owns the door: `validateQueryMetadata` **throws** on an invalid bag rather than letting a connector silently mangle it or letting a backend reject the whole query. The contract is deliberately generic — permissive enough to be usable across every backend, strict enough that the bag can always be rendered safely (including into the comment form):

- property **names**: ASCII alphanumerics and underscore
- property **values**: printable ASCII, excluding the double-quote
- a small number of properties

A connector may still transform *within* those rules to satisfy its own native limits — **BigQuery is the only one that does**, lowercasing keys and substituting characters its label grammar disallows.

## No read-back — correlate on an id you supply

There is intentionally **no** response-side execution-id field. A caller that wants to correlate a Malloy result with the warehouse's own query record mints its own id (e.g. a `query_id` property), includes it in the bag, and logs it locally. The id travels out with the query, so the warehouse's history row is found by matching it — no value needs to be read back through Malloy, and nothing in core needs to enumerate per-dialect id shapes.

## Identity

Query metadata never affects query results or data identity: it is **excluded from the connection digest**, so changing it never re-keys a connection or re-addresses anything. (`setupSQL`, which can change query semantics, stays in the digest as before.)

## Testing

- Per-connector offline unit tests (no live warehouse): the metadata reaches the driver call / is rendered into the statement, the transforms and escaping are correct, an empty bag adds no prefix, and an invalid bag throws.
- A core spec for the contract and helpers.
- The underlying per-engine mechanisms — Snowflake `QUERY_TAG`, BigQuery label attachment, and the leading-comment path on Trino / Databricks / Postgres — were additionally exercised end-to-end against a live instance of each engine.


mtoy-googly-moogly:
Closed, commits moved to #3019 which is now merged

PR #3012, opened 2026-07-24
fix(persist): don't materialize an `extend`-derived reader of a persisted source

jswir:
## Problem

`#@ persist` is a Malloy annotation, so it propagates onto a source that `extend`s a persisted source. `Model.getBuildPlan()` enumerates build targets from `findPersistentDependencies`, gated on the folded `persistent` flag — which is `true` for an `extend`-derived reader (correctly, since it must read the pre-built table). As a result the reader was emitted as a **second build target**, materializing the base's table again under the same inherited `name=`.

Minimal repro (DuckDB dialect):

```malloy
##! experimental.persistence
#@ persist name="main.base_persist"
source: base_persist is duckdb.table('t') -> {
  group_by: astr
  aggregate: n is count()
}

// no #@ persist of its own — just adds a computed field
source: extended_reader is base_persist extend {
  dimension: astr_upper is upper(astr)
}
```

`getBuildPlan().sources` returned **both** `base_persist` and `extended_reader` (with the reader as the graph root and `base_persist` nested under `dependsOn`), so a consumer materializes one physical table twice.

Per the [persistence docs](https://docs.malloydata.dev/documentation/experiments/persistence), extending a persistent source should **read** the pre-built table, not re-materialize it (`#@ -persist` is the opt-out to recompute).

## Fix

Distinguish "uses a persisted table" from "declares the materialization":

- Add `persistDeclared` to `PersistableSourceProperties`, computed at define time (`DefineSource.execute`) from the source's **own** annotation only (via `checkPersistDeclaredOnOwn`), so an inherited `#@ persist` doesn't count and an own `#@ -persist` reports false.
- Gate build-target inclusion in `findPersistentDependencies` on `persistDeclared`. An inherited-only reader keeps `persistent === true` (so it still routes to the pre-built table) but is no longer a build target — it flattens to its real persisted dependency, exactly like a non-persistent source.

The folded `persistent` flag and the serve/read path are unchanged.

## Tests

`persist-source-sql.spec.ts`:
- an `extend` of a persisted source is a reader, not a build target
- an `extend` reader still routes to the base's materialized table
- a `#@ -persist` extension opts out (not a build target)
- an extension that declares its **own** persist name stays a distinct build target

Typecheck clean; existing persist/annotation/import/sql-block/runtime unit suites pass.

### DIFF
--- a/packages/malloy/src/api/foundation/persist-source-sql.spec.ts
+++ b/packages/malloy/src/api/foundation/persist-source-sql.spec.ts
@@ -95,3 +95,98 @@ describe('PersistSource build key matches serve-time manifest lookup', () => {
     expect(routedSQL(model, 'db_rollup', '_db_')).toContain(MATERIALIZED);
   });
 });
+
+describe('build targets exclude extend-derived readers', () => {
+  function buildTargets(model: Model): string[] {
+    return Object.values(model.getBuildPlan().sources)
+      .map(s => s.name)
+      .sort();
+  }
+
+  // The bug this guards: `#@ persist` is a Malloy annotation, so it propagates
+  // onto a source that `extend`s a persisted source. Such a reader must keep
+  // reading the pre-built table (docs: "the extension still reads from the same
+  // pre-built table"), but it must NOT be emitted as its own build target — it
+  // would materialize the base's table a second time under the inherited
+  // `name=`, colliding on one physical coordinate.
+  test('an extend of a persisted source is a reader, not a build target', () => {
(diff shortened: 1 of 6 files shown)


mtoy-googly-moogly:
So this does not actually fix the problem you identified, I am working on fix which addresses the problem that this PR reveals. A lot of the design of persistence was pushed out to the builder, and a tiny piece of what the builders need to do should probably be in the core, as there are multiple builders which do it wrong, including the sample ones I wrote.

PR #3015, opened 2026-07-24
update persistence documentation

mtoy-googly-moogly:
Persistence has no in-repo explanation. The annotation, the manifest, the compile-time substitution, and the builder contract are spread across four CONTEXT files, the API surface, and a sample nobody runs. `src/doc/persist/api.md` and `src/doc/persist/internal.md` are that explanation.

All of those places described a leveled build schedule that `getBuildPlan()` does not produce: it emits a single level holding the root sources, with the real ordering in each node's `dependsOn`. `scripts/simple_builder` believed its own header and walked only the levels, so it built roots and silently skipped every dependency — its `flattenBuildNodes()` was the correct walk and was never called.

The sample is cited as documentation, so it now has a guard: a `simple-builder` jest project, in `test-duckdb` and `ci-duckdb`, and `ci-test-sanity-check.sh` scans `scripts/` so nothing there is exempt again. The sample also builds and records the dialect's canonical table name rather than the raw `name=` — creating one and recording the other leaves the manifest naming a table nothing created.


housejester:
Thanks for writing these up!

One behavior I couldn't place against the docs. `api.md` introduces the two persistable types symmetrically:

> Only sources backed by a computation can be persisted: `query_source`
> (`source: x is y -> {...}`) and `sql_select` (`source: x is conn.sql("..."))`.

Substitution doesn't appear to be the same for them, though. Building both and then querying each one, a `query_source` reads its persisted table and an `sql_select` recomputes.

Here is one query touching both, so a single generated SQL shows the difference:

```
source: combined is t extend {
  join_one: q is qs on k = q.k      // qs: a query_source,  #@ persist name="qs_tbl"
  join_one: s is ss on k = s.k      // ss: an sql_select,   #@ persist name="ss_tbl"
}
```

```sql
FROM t as base
LEFT JOIN qs_tbl AS q_0                                              -- substituted
LEFT JOIN (SELECT k, SUM(amount) AS total FROM t GROUP BY k) AS s_0  -- inlined
```

Behaviorally, build both tables, then mutate the underlying data and re-run. A leg reading its built table (query_source) stays STALE (expected); a leg that recomputes (sql_select) goes FRESH (unexpected. at least to me it was).

```
after INSERT INTO t VALUES ('a', 1000):

  qs         STALE (read its table)   [{"k":"a","total":150}]  -> [{"k":"a","total":150}]
  ss         FRESH (recomputed)       [{"k":"a","total":150}]  -> [{"k":"a","total":1150}]
  viainterp  STALE (read its table)   [{"k":"a","total":150}]  -> [{"k":"a","total":150}]
```

`viainterp` is `duckdb.sql("""SELECT k, total FROM %{ ss }""")`. so ss's table *is* read, through interpolation. That path works exactly as documented, it just seems to be the only one that reaches it.

That matches the code: in `getStructSourceSQL` (`model/query_query.ts`), `case 'query_source'` does a manifest lookup gated on `persistent`, while `case 'sql_select'` returns `` `(${getCompiledSQL(...)})` `` with no lookup. and
`getCompiledSQL` returns `src.selectStr` verbatim when the source has no `selectSegments`. `expandPersistableSource`, the function that consults the manifest for a persistable source, is only reachable from `getCompiledSQL`'s `selectSegments` loop.

Is the asymmetry deliberate? If so, `Limitations` entry might be good. as written the two types read as interchangeable, and the failure is quiet: the build succeeds, the table exists, and a direct query is simply no faster. If not, mirroring
the `case 'query_source'` lookup into `case 'sql_select'` looks contained.

---
Separately, and not a docs issue: `ModelMaterializer.loadQuery(str, {buildManifest})` drops the `CompileQueryOptions` half of its `options`, so a `buildManifest` passed there is silently ignored. `loadFinalQuery`, `loadQueryByName` and `loadQueryByIndex` all merge `{...this.compileQueryOptions, ...options}`, while `loadQuery` calls `makeQueryMaterializer(fn)` with no second argument. Is that intended? I hit this when trying to see the sql get replaced and I (well, claude) picked up ModelMaterializer.loadQuery first, and it inlined there. It could be i'm wielding the tool wrong there though.

I can put up a gist for both of these if helpful.

PR #3021, opened 2026-07-27
fix(drill): filtered measures, boolean/and stable filters, and missing joins

lloydtabb:
Three defects that surface when drilling a result cell into a standalone query. Each is independently reproducible; they're together because they share the drill metadata path and one masks the next.

## 1. A filtered measure drills to a superset of its own rows

`QueryAtomicField.getFilterList()` was `return []` — a stub feeding a live call site. `getResultMetadata` calls it for measures and hands the result to `addDrillFiltersTag`, so a measure like `flight_count { where: carrier = 'WN' }` never emitted `drill_filters` at all. Drilling it produced the rows for the whole group.

Concretely, `total_babies { where: gender = 'F' }` for CA reads **15,340,742**, and the drill returned all **32,042,437** rows — the user is looking at a row list twice the size of the number they clicked, with nothing to indicate it.

It now walks the `filteredExpr` chain on the field's expression. Only a filter wrapping the *whole* expression is hoisted: in `x + count() {where: f}` the `filteredExpr` isn't at the top and `f` doesn't constrain the field.

## 2. `getStableFilter` only recognized two shapes

`= literal` and `~ filter-expression`. Anything else produced no stable form, so the strict reader (`stableDrillFilters`) treated the whole drill as unstable. Added:

- **bare booleans** — `where: is_domestic` is `is_domestic = true`, representable today via `LiteralValueWithBooleanLiteral`
- **`and`** — decomposes into independent clauses, recursively, so `a and b and c` flattens

`or` deliberately stays unstable: drill clauses are conjunctive, so splitting an `or` would silently widen the predicate.

This makes `FilterCondition.stableFilter` plural (`stableFilters`). The wire format is unchanged — `drill_filters` was already an array — so the renderer needed no changes.

## 3. A drill through a joined field dropped the join

The one that produces bad SQL rather than a wrong answer. When a drill path resolves through a view whose field is a reference across a join (`group_by: users.full_name` inside a view), the `DrillField` carried only `fieldDef.refSummary` — that field's usage *inside* `users`, never that `users` itself was used. The join path lived only in `compareField.path`, which nothing read for usage.

The builder then omitted the join while the filter still referenced its alias:

```sql
FROM order_items as base
 LEFT JOIN inventory_items AS inventory_items_0 ...
WHERE ((CONCAT(UPPER(SUBSTR(users_0."first_name",1,1)), ...))='Philip Labat')
```

`users_0` is never joined. This **compiles clean** — `problems: []` — and fails at execution with `Binder Error: Referenced table not found`.

It stayed hidden because `getDrillQueryMalloy` always appends `+ { select: * }`, and `select: *` happens to drag every join in. It surfaces as soon as the drill is composed with anything narrower, e.g. `source -> some_view + { drill: … }`.

Fixed by recording the reference path as field usage, the way `ExprIdReference` does for an ordinary `where:`.

## Also

`addDrillFiltersTag` now indexes by a running **output** position. It previously used the source index while `continue`-ing past skipped conditions, leaving holes in the tag array that read back as filters with no expression — which makes an otherwise-fine drill unstable. Fan-out from `and` would have made that fire routinely.

## Tests

Six new cases in `drill.spec.ts`: filtered measure contributes its filter; row drill unaffected by a sibling filtered measure; `and` decomposes; bare boolean is stable; `or` stays unstable and falls back to raw code; and drill-through-a-join.

That last one **runs** the query rather than matching text, because the defect produced valid-looking SQL. With the fix stashed it fails with `Binder Error: Referenced table "carriers_0" not found!`.

`malloy-render` 161 passing. `malloy-core` and `db-duckdb` show 11 pre-existing failures on this base — all in the in-flight `none` filter operator (`fnumber`/`ftemporal`/`filter-expressions`), verified identical on clean `main`.

### DIFF
--- a/packages/malloy-render/src/data_tree/drilling.ts
+++ b/packages/malloy-render/src/data_tree/drilling.ts
@@ -51,6 +51,16 @@ export function getStableDrillClauses(
   const result: Malloy.DrillOperation[] = [];
   while (current) {
     const {field} = current;
+    // The clicked leaf itself. A filtered measure (`total_sales {where: ...}`)
+    // carries its own drill_filters, and dropping them would drill to a SUPERSET
+    // of the rows behind the number that was clicked. Only the starting cell can
+    // be a leaf — every parent is an array or a record — so this runs at most
+    // once, and is disjoint from both branches below.
+    if (!field.isArray() && !field.isRecord()) {
+      const filters = field.stableDrillFilters;
+      if (filters === undefined) return undefined;
+      result.unshift(...filters.map(f => ({filter: f})));
+    }
     if (field.isArray()) {
       const filters = field.stableDrillFilters;
       if (filters === undefined) return undefined;
@@ -96,6 +106,12 @@ export function getDrillValues(cell: Cell): DrillValue[] {
   const result: DrillValue[] = [];
   while (current) {
     const field = current.field;
+    // The clicked leaf's own filters — see getStableDrillClauses. drillFilters
+    // falls back to the raw `code` when a filter has no stable form, so unlike
+    // the stable walk there is nothing to bail on here.
+    if (!field.isArray() && !field.isRecord()) {
+      result.unshift(...field.drillFilters.map(f => ({where: f})));
+    }
     if (field.isArray()) {
       const filters = field.drillFilters;
       result.unshift(...filters.map(f => ({where: f})));
--- a/packages/malloy-render/src/drill.spec.ts
+++ b/packages/malloy-render/src/drill.spec.ts
@@ -48,6 +48,136 @@ describe('drill query', () => {
     expect(row.getStableDrillQueryMalloy()).toEqual(expDrillQuery);
   });
 
+  // A filtered measure carries its own drill_filters. Drilling the CELL must
+  // include them, or the drill lands on a superset of the rows behind the
+  // number that was clicked.
+  test('filtered measure contributes its own filter', async () => {
+    const query = `
+      run: flights_base -> {
+        group_by: carriers.nickname
+        aggregate: flight_count
+        aggregate: wn_flights is flight_count { where: carrier = 'WN' }
+        limit: 1
+      }
+    `;
+    const result = await duckdb.loadModel(baseModel).loadQuery(query).run();
+    const table = getDataTree(API.util.wrapResult(result));
+    const cell = table.rows[0].column('wn_flights');
+    const expDrillQuery = `run: flights_base -> {
+  drill:
+    carriers.nickname = "Southwest",
+    carrier = "WN"
+} + { select: * }`;
+    expect(cell.getStableDrillQueryMalloy()).toEqual(expDrillQuery);
+  });
+
+  // The row-level drill must NOT pick up a sibling measure's filter — the leaf
+  // branch only fires for the clicked cell.
+  test('row drill is unaffected by a sibling filtered measure', async () => {
+    const query = `
+      run: flights_base -> {
+        group_by: carriers.nickname
+        aggregate: flight_count
+        aggregate: wn_flights is flight_count { where: carrier = 'WN' }
+        limit: 1
+      }
+    `;
+    const result = await duckdb.loadModel(baseModel).loadQuery(query).run();
+    const table = getDataTree(API.util.wrapResult(result));
+    const expDrillQuery =
+      'run: flights_base -> { drill: carriers.nickname = "Southwest" } + { select: * }';
+    expect(table.rows[0].getStableDrillQueryMalloy()).toEqual(expDrillQuery);
+  });
+
+  // A drill clause reaching through a JOIN must bring that join along, even when
+  // the view it lands on doesn't otherwise need it. Regression: the drill field's
+  // refSummary carried only the field's own usage, so the join was omitted and
+  // the SQL referenced an unjoined alias — it COMPILED, then failed at execution
+  // with a binder error. Hence this runs the query rather than matching text.
+  test('drill through a joined field pulls the join into the query', async () => {
+    const model = `
+      source: flights is flights_base extend {
+        view: by_origin is { group_by: \`Origin Code\`; aggregate: flight_count }
+        view: by_carrier is { group_by: carriers.nickname; aggregate: flight_count }
+      }
+    `;
+    const result = await duckdb
+      .loadModel(baseModel)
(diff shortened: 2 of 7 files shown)


lloydtabb:
Moving this to **draft**. An adversarial self-review turned up a live regression and two incomplete cases. Notes below so nobody re-derives them — @mtoy-googly-moogly this is the state of it.

## Verified solid

**Claim 3 (missing joins) is real and the fix works.** Isolated properly: disable *only* the two `compareFieldUsage` lines, keep everything else, **rebuild**, and the new test fails with `Binder Error: Referenced table "carriers_0" not found!`. Restore, 161 pass.

⚠️ **Trap for anyone verifying this:** `@malloydata/malloy`'s `main` is `dist/index.js`, so jest runs the **built** package, not source. Revert a compiler file without `cd packages/malloy && npm run build` and the old dist still answers — the change looks inert and the test passes for the wrong reason. Both my first "proof" and the review's counter-claim went wrong here in opposite directions.

However, the mechanism described in the code comment is **wrong** — `refSummary` feeds `addFieldUserFromFilter` → `compositeFieldUsers`, which drives composite-source input selection. The join emission happens elsewhere. Behavior is right, explanation needs rewriting.

## Blocking — regression introduced by this PR

**`getFilterList()` returns filters scoped to the *defining* source, not the query root.** `packages/malloy/src/model/query_node.ts:138-149`

The `FilterCondition` paths are relative to wherever the measure was declared. `to_stable` writes them as that field's `drill_filters`, and the renderer splices them into a drill rooted at the query's source. Nothing re-bases the path.

```malloy
source: carriers is duckdb.table('.../carriers.parquet') extend {
  primary_key: code
  measure: sw_count is count() { where: nickname = 'Southwest' }
}
source: flights_base is ... extend { join_one: carriers with carrier }
run: flights_base -> { group_by: `Origin Code`; aggregate: carriers.sw_count; limit: 1 }
```

Drilling that cell emits `drill: \`Origin Code\` = "TUL", nickname = "Southwest"` — should be `carriers.nickname`. It errors with *"Drill reference be a view name followed by a path to a field in that view"*.

**On `main` this cell drilled fine** (`getFilterList()` returned `[]`), so this is a new failure. And the error is the *lucky* outcome: if the root source has a same-named field — `name`, `state`, `status`, `id` are all common — it compiles and **silently filters the wrong column**. Both the stable and `code` paths are wrong, so there's no safe fallback.

## Blocking — incomplete

**Bare-boolean fallback fires on source parameters.** `filters.ts:89-97`. `ExprIdReference.drillExpression()` never checks `refType`, so a boolean parameter clears the type gate and emits `param = true` as a drill clause:

```malloy
source: fp2(only_wn::boolean is true) is flights_base extend {
  measure: pc is flight_count { where: only_wn }
}
```
→ `drill: \`Origin Code\` = "ATL", only_wn = true`, which won't compile. The same hole exists in the pre-existing `param = literal` branch, but `where: my_flag` is far more idiomatic, so this widens the exposure.

**Composed measures still drop the inner filter** — the exact bug this PR claims to fix. `query_node.ts:143-147`

```malloy
measure: wn     is flight_count { where: carrier = 'WN' }
measure: wn_100 is wn           { where: distance = 100 }
```
`wn_100.e` is `filteredExpr{ e: {node:'field', path:['wn']}, filterList:[distance=100] }`. The walk stops at the top-level unwrap and never sees `carrier='WN'`, so the drill is a superset — silently. It needs to follow `{node:'field'}` into the referenced field def.

## Worth knowing, not caused here

**The shape the renderer actually emits is broken for joined fields, independently of drill.** `getDrillQueryMalloy` produces `run: src -> { drill: … } + { select: * }`, and:

```
{drill; select:*}  one segment                 -> OK
{drill} + {select:*}                           -> FAIL  Binder Error
{drill} + {aggregate}                          -> FAIL
{where: carriers.nickname=...} + {select:*}    -> FAIL   <-- not drill-specific
```

Identical on `main`. It's a **refinement** bug (plain `where:` reproduces it), which means existing string-only assertions like `'run: flights_base -> { drill: carriers.nickname = "Southwest" } + { select: * }'` are asserting a query that cannot execute. Probably deserves its own issue.

**Capability regression on the stable path.** `aggregate: far is flight_count { where: distance > 1000 }` now makes `getStableDrillQueryMalloy()` return `undefined` for that cell. `>`, `<`, ranges, `or` and alternations in filtered measures are common, so a large class of previously-drillable cells now return nothing on the stable path. Correct-by-construction, but silent.

**`stableFilter` → `stableFilters` is a serialized-model wire break.** `FilterCondition` lives inside `PipeSegment` inside a serialized `ModelDef`, so anything caching compiled models across versions reads the old key as `undefined` and silently degrades every drill to unstable. Wants a compat read (`stableFilters ?? (stableFilter ? [stableFilter] : undefined)`) or an explicit note.

**PR description correction:** the `addDrillFiltersTag` rationale is wrong. `Tag.set` on sparse indices compacts — it does not leave readable holes. The running output index **is** required, but because `and` fan-out puts two stable filters at one source index and the first gets overwritten. Not because of holes.

## Checked, no problem found

- `and` decomposition is sound — `a and (b or c)` and `not (a and b)` both correctly stay unstable. (Side effect: `(a = 1) and (b = 2)` is also unstable, since parens break the `ExprEquality` match.)
- No missed consumers of the `stableFilter` rename in `packages/*/src`.
- No over-hoisting in `getFilterList()` — `count(){where:f} + 1`, `x{where:a} / y{where:b}` and `all(...)` all correctly drop.
- The `isAtomic(field)` branch in `drill.ts` correctly needs no usage recording — an inline `group_by: nick is carriers.nickname` already carries the right `refSummary`.
- 11 failures in `malloy-core` (`fnumber`/`ftemporal`/`filter-expressions`, all `none` operator) are pre-existing — verified identical on clean `main`.

## Plan

Claim 3 can stand once its comment is corrected. Claims 1 and 2 need the scope-rebasing fix and the parameter guard before this is safe to merge. Happy to split claim 3 out if it's more useful on its own.

codeCraft-Ritik (commented):
Fixing the drill filter metadata path resolves some very subtle bugs—especially joined fields dropping their join alias during narrow drill compositions (`Binder Error: Referenced table not found`). 

Also love the safety check keeping `OR` expressions unstable to prevent predicate widening. Excellent fixes and test coverage! 🚀

PR #3035, opened 2026-08-11
Bump @duckdb/node-api to 1.5.5-r.4

housejester:
Moves the DuckDB engine in `malloy-db-duckdb` from 1.5.3 to 1.5.5 (patch-level), and the node bindings from 1.5.3-r.2 to 1.5.5-r.4.

## Why

- DuckDB 1.5.4/1.5.5 are fix-only patch releases: several out-of-bounds-read security fixes, storage/checkpoint correctness fixes (`DROP COLUMN` metadata bookkeeping, row-group stats), and concurrency fixes (`TemporaryMemoryManager` deadlock, concurrent `ALTER`+`INSERT` crash). Release notes: [v1.5.4](https://github.com/duckdb/duckdb/releases/tag/v1.5.4), [v1.5.5](https://github.com/duckdb/duckdb/releases/tag/v1.5.5).
- Community extensions are built per DuckDB version, and the v1.5.3 builds are frozen at older extension code. Concretely, the `gcs` community extension's v1.5.3 build predates its `DirectoryExists` implementation, which DuckLake needs for `DATA_PATH` on GCS — so DuckLake-on-GCS via that extension only works on a 1.5.5 engine.
- The bindings delta (r.2 → r.4) is additive: engine tracking, list child-vector capacity fix, clearer MAP errors, N-API reference-lifetime fixes, and a new table-function API. No breaking TS API changes.

## Verification

- `npm run build`, `npm run typecheck`, and the full `npm run test-duckdb` suite (118 suites / 3,994 tests across malloy-core, render, db-duckdb, db-duckdb-core, simple-builder) pass unchanged.
- A downstream consumer (malloy-publisher) was built and tested against these exact packages via a local overlay: typecheck, unit suites, DuckDB extension bake (httpfs, aws, azure, postgres, ducklake, excel, bigquery, snowflake all load on 1.5.5), and its end-to-end scenario harness (real Postgres + DuckLake + booted server) — all green.

The lockfile diff touches only the `@duckdb/node-api` / `@duckdb/node-bindings` family.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3806,37 +3806,37 @@
       "license": "MIT"
     },
     "node_modules/@duckdb/node-api": {
-      "version": "1.5.3-r.2",
-      "resolved": "https://registry.npmjs.org/@duckdb/node-api/-/node-api-1.5.3-r.2.tgz",
-      "integrity": "sha512-xjn6W9n8Sn9PaYmzrB6ayidNLQHpqNKdaRz4gx3yVoLUkrsIebQC1M1jV3BR8KqXj5P0RX4wLs8Tg/fK2ZQmNA==",
+      "version": "1.5.5-r.4",
+      "resolved": "https://registry.npmjs.org/@duckdb/node-api/-/node-api-1.5.5-r.4.tgz",
+      "integrity": "sha512-8v0CZNo7aM6GQCNUHERGTrZWIfss8xKzZPF3ACNhPdMMrt7UjkNI5nRQ9FTFO7Yx8ghxYxVpdotmBBMQ5vXUOA==",
       "license": "MIT",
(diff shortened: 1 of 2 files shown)


mtoy-googly-moogly:
Thanks @housejester — taking this, with the DuckDB 1.5.5 engine and the DuckLake-on-GCS reasoning intact, in #3037. That PR carries your `@duckdb/node-api` change byte-for-byte and adds the matching `@duckdb/duckdb-wasm` dev45.0 → dev57.0 bump so both DuckDB paths stay on the same line, which is the only reason it's a separate PR rather than a review here. Your downstream verification against malloy-publisher is credited in the description and is what made this an easy call.

Closing in favor of #3037.

PR #3043, opened 2026-08-19
fix(lang): apply access modifiers to `*` expansion

seckenrode:
A field marked `private` or `internal` by an `include` block was rejected when a query named it, and returned anyway when the query wrote `*`:

    source: c is a include { *; private: ai }
    run: c -> { select: ai }   // error: 'ai' is private
    run: c -> { select: * }    // returned ai

Named references resolve through `StaticSpace.lookup()`, which compares the field's modifier against the access level of the space doing the reading. Star expansion instead walks `entries()`, the raw namespace map, which has no access check in it. Both `addWild()` implementations — `select:` in query-spaces and `index:` in index-field-space — did the same thing, so `index: *` leaked the field names *and* a sampled value range for each one.

The walk they duplicated now lives once, as `wildcardExpansion()` on `QueryOperationSpace`, and it does what `lookup()` does: start from the access level of the space the `*` was written in, narrow it at each join hop with `lessPermissiveAccessLevel`, and drop the entries that level disallows. A join named in the path before the `*` is checked too, so `select: private_join.*` is now an error. Fields are dropped silently rather than reported: an error naming the field would disclose the name the modifier exists to hide. Each `addWild()` keeps only the part that was ever different — how it names and type-filters what comes back.

One error code moves. The access check runs before the check for whether the path part can be expanded at all, so an inaccessible scalar now reports `field-not-accessible` where it reported `invalid-wildcard-source`. Reporting that a field is not expandable confirms it is there.

Unchanged: `*` written inside the statement that declares the modifiers still expands to everything, because the labels are not on the field defs until that statement completes. That is the same rule that lets the declaring statement's `extend` block name a private field.

### DIFF
--- a/packages/malloy/src/lang/CONTEXT.md
+++ b/packages/malloy/src/lang/CONTEXT.md
@@ -194,6 +194,19 @@ The `restrictedMode` flag flows: `ParseOptions.restrictedMode` → `MalloyTransl
 
 API-level details: [`../api/CONTEXT.md`](../api/CONTEXT.md) and the JSDoc on `ModelMaterializer.loadRestrictedQuery`.
 
+## Access modifiers
+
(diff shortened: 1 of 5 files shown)


mtoy-googly-moogly:
Closed, went with #3051 instead

PR #3065, opened 2026-09-03
fix(trino): move to the renamed @trinodb/trino-js-client

mosabua:
The Trino client was republished under the `trinodb` scope. Versions up to 0.2.9 shipped as the unscoped `trino-client`; the scoped `@trinodb/trino-js-client` starts at 0.3.0. The unscoped package is no longer maintained.

I am a Trino maintainer, and we are working through the downstream consumers of the client as part of that move.

## This also clears the axios hold

`docs/dependency-management` tracks this connector as a standing advisory hold:

> `high      axios, trino-client          → Trino connector maintenance`

The lockfile carried a nested copy pinned by the old client. After the bump it is gone and the top level moves up:

```
before:  node_modules/axios                            1.18.1
         node_modules/trino-client/node_modules/axios  1.13.2
after:   node_modules/axios                            1.19.0
```

axios 1.19.0 closes the prototype pollution fix from 1.15.1, the CR/LF header injection vector, the NO_PROXY loopback bypass from 1.16.0, and raises the form-data floor past GHSA-hmw2-7cc7-3qxx.

## Risk

Low. The declaration files of `trino-client@0.2.9` and `@trinodb/trino-js-client@0.3.2` are identical byte for byte:

```
$ npm pack trino-client@0.2.9 && tar xzf trino-client-0.2.9.tgz
$ diff -u package/dist/index.d.ts node_modules/@trinodb/trino-js-client/dist/index.d.ts
$ echo $?
0
```

So this is a rename of the dependency and its import specifiers, with no API change. Same twenty exports, same shapes.

0.3.2 contains no source changes relative to 0.3.1 either — it is a dependency-maintenance release we cut specifically so that downstream renames would carry the axios fixes rather than only a new name.

## Changes

- `packages/malloy-db-trino` — dependency and the imports in `trino_connection.ts`, `trino_query_metadata.spec.ts`, and the commented-out import in `test.nonspec.ts`
- `packages/malloy-render-validator` — two comments that name the package by its old identifier
- `.github/dependabot.yml` — the `connectors` group pattern becomes `@trinodb/*`, otherwise the old `trino-client` pattern stops matching and this connector silently drops out of the group
- `package-lock.json`

## Verification

I did not run the Trino dialect CI, which needs a live Trino. Beyond the declaration comparison above, I verified 0.3.2 against a real coordinator using a harness built from this connector's own call patterns — `Trino.create`, `BasicAuth`, the `for await` drain in `runSQL`, error handling off the result object, and session header round-tripping. All checks pass. That harness now runs in the client's CI (trinodb/trino-js-client#970).

Commits are signed off per CONTRIBUTING.md.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -107,7 +107,7 @@ updates:
           - "@google-cloud/*"
           - "@databricks/*"
           - "@prestodb/*"
-          - "trino-client"
+          - "@trinodb/*"
           - "snowflake-sdk"
           - "@motherduck/*"
           - "mysql2"
--- a/package-lock.json
+++ b/package-lock.json
@@ -11587,6 +11587,15 @@
       "integrity": "sha512-C5Mc6rdnsaJDjO3UpGW/CQTHtCKaYlScZTly4JIu97Jxo/odCiH0ITnDXSJPTOrEKk/ycSZ0AOgTmkDtkOsvIA==",
       "license": "MIT"
     },
+    "node_modules/@trinodb/trino-js-client": {
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@trinodb/trino-js-client/-/trino-js-client-0.3.2.tgz",
+      "integrity": "sha512-TNFOXkk2wp/SSgNtRdfArKZSR78ebYR7zxTobu+o01C+KJO7ELWcti90DeCw4WJDh0YbjZGT0XuXV48n7TEyoQ==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "axios": "1.19.0"
+      }
+    },
     "node_modules/@ts-graphviz/adapter": {
       "version": "2.0.6",
       "resolved": "https://registry.npmjs.org/@ts-graphviz/adapter/-/adapter-2.0.6.tgz",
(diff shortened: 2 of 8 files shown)


mtoy-googly-moogly:
@mosabua Thanks for this, there's some fuss about the axios upgrade that I am having claude explain to me, hope to get this merged in a day or so.

mtoy-googly-moogly:
Thanks for this, and for doing the sweep from the client side — knowing you were
working through the consumers deliberately made this easy to trust.

It's landing as #3066, which carries your commit unchanged. Two things had to ride
along with it, and neither is a problem with this PR.

Moving to `@trinodb/trino-js-client` pulls axios forward, and our generated
publisher OpenAPI client doesn't compile against 1.19 — `request()` resolves its
response through a conditional on a type axios doesn't export, so the inferred
return type has no name. That break is ours and it was latent; any axios bump
would have hit it, and our own Dependabot would have walked into it next pass.

And since the client pins axios exactly, we now pin axios to match (`1.19.0`) and
hold the connector at `0.3.2` so the two can't drift apart — a release like 0.3.2,
whose only change is the axios pin, would otherwise leave us with a second nested
copy. We've written trinodb/trino-js-client#956 into our dependency ledger as the
thing that retires both pins, so when axios gives way to native `fetch` we drop
them rather than carry them.

For what it's worth, your risk assessment checked out and then some: the 0.3.2
tarball is byte-identical to `trino-client@0.2.9` apart from the package name and
the axios pin — same file list, no diff anywhere in `dist`. Full dialect CI is
green on #3066, Trino included.

Closing in favour of #3066, with you as co-author on the merge. malloy should be
tickable off trinodb/trino-js-client#979 once it lands.

PR #3066, opened 2026-09-04
fix(trino): move to the renamed @trinodb/trino-js-client

mtoy-googly-moogly:
Thanks to @mosabua for #3065 — the rename here is his commit.

We couldn't take it on its own. Moving to `@trinodb/trino-js-client` pulls axios
forward, which breaks our generated publisher client, and the client pins axios
exactly, so we have to pin axios to match: `1.19.0`, with the connector held at
`0.3.2` so the two can't drift apart. trinodb/trino-js-client#956 drops axios for
native `fetch` and retires both pins.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -107,7 +107,7 @@ updates:
           - "@google-cloud/*"
           - "@databricks/*"
           - "@prestodb/*"
-          - "trino-client"
+          - "@trinodb/*"
           - "snowflake-sdk"
           - "@motherduck/*"
           - "mysql2"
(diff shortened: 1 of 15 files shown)


.github/dependabot.yml
```diff
@@ -56,6 +56,13 @@ updates:
       # (pg/pg-cursor/pg-query-stream move together); the real fix is issue #2928.
       - dependency-name: "pg"
       - dependency-name: "pg-query-stream"
+      # axios exact-pinned (1.19.0) at the root to match the exact pin inside
+      # @trinodb/trino-js-client. Any other version resolves a second, nested copy
+      # under the connector instead of one hoisted axios. The pin is not ours to
+      # keep: trinodb/trino-js-client#956 drops axios for native fetch in their
```
mosabua, line 62: Just to be clear .. we are not 100% set on dropping axios since there are potentially issues. But either way .. we will let you know

Issue #2928, opened 2026-06-22
Postgres connector: pg / pg-cursor / pg-query-stream must move as a coupled set


mtoy-googly-moogly:
The connectors-group PR #2923 bumped `pg` 8.7.3→8.22.0 and `pg-query-stream` 4.2.3→4.16.0, and **every dialect test job failed** — not just postgres — with:

```
Cannot find module 'pg/lib/result.js' from 'node_modules/pg-cursor/index.js'
```

## Root cause
`pg-query-stream` pulls `pg-cursor`, and **`pg-cursor` deep-imports `pg`'s internal `lib/result.js`**. The bump created a version skew between `pg` and the installed `pg-cursor`, so the internal path no longer resolves. `pg` / `pg-cursor` / `pg-query-stream` are a **coupled set** — they must move to mutually-compatible versions *together*, like the gts/eslint cluster. Bumping `pg` alone (or `pg` + `pg-query-stream` with a lagging `pg-cursor`) breaks.

Why it took down *every* dialect, not just postgres: the shared test harness loads the postgres connection path, so the broken `pg-cursor` require fails module resolution at load time for all test projects (duckdb, snowflake, trino included).

## Stopgap in place
`pg` and `pg-query-stream` are held at their working versions — exact-pinned in `packages/malloy-db-postgres/package.json` and `ignore`d in `.github/dependabot.yml`, recorded in `.github/dependabot-pins.md`. This issue tracks doing it right and removing the holds.

## Scope
- Bump `pg` + `pg-cursor` + `pg-query-stream` together to mutually-compatible versions; verify `db-postgres` and that the shared harness still loads.
- Remove the exact pins and the `ignore` entries; delete the ledger row.

mtoy-googly-moogly:
**Bundler note for whoever does this bump:** the exact-pin to `pg@8.7.3` has a downstream-bundling side effect worth fixing when the set moves.

pg `8.7.3`'s `lib/native/client.js` does a **bare** `require('pg-native')`. esbuild (and other bundlers) treat that as a required import and **hard-error** when pg-native isn't installed (it's an optional native client). Newer pg wraps the same require in a `try/catch` — with a literal comment *"to avoid upstream bundlers from complaining"* — which esbuild treats as **optional** (warning, not error). The guard is absent in 8.7.3 and present by 8.22.0 (exact introduction version not pinned down).

[...]

Issue #2918, opened 2026-06-22
Stop reaching into vscode-textmate internals in generateMonarchGrammar — fix it properly


mtoy-googly-moogly:
`packages/malloy-syntax-highlight/scripts/generateMonarchGrammar.ts` works around the absence of a public API by importing **internal** paths of `vscode-textmate` and depending on the shape of its private rule types:

- `import … from 'vscode-textmate/release/theme'`
- `import … from 'vscode-textmate/release/rawGrammar'`
- relies on `TextMateBeginEndRule` exposing `beginCaptures` / `endCaptures` / `captures` / `name`

None of that is part of `vscode-textmate`'s public surface. When it went `9.0.0 → 9.3.2` — an **in-range minor**, semver entirely honored — it stopped exporting those subpaths and reshaped the internal types, and our syntax-highlight codegen build broke. We broke ourselves by reaching past the front door.

## The principle
Don't work around a missing API by reaching into a dependency's secret implementation details. Either use the package's public API, or own the small set of types/data we need (vendor or define them) so an internal refactor upstream can't break our build. A pin is a stopgap, not the fix.

## Stopgap currently in place
`vscode-textmate` is pinned at `9.0.0` (Dependabot `ignore` + recorded in `.github/dependabot-pins.md`) so it can't ride the monthly group and re-break the build. This issue tracks doing it right and removing the pin.

## Scope
- Rework `generateMonarchGrammar.ts` to consume only the public `vscode-textmate` API, or vendor the minimal grammar/theme types it needs.
- Then unpin `vscode-textmate` (remove the `ignore` and the ledger entry).
- Relates to the textmate-grammar-rebuild work.

PR #3082, opened 2026-09-09
Move every pin that can move, and make the rest checkable

mtoy-googly-moogly:
Move every dependency that can move, and make each one that can't say why in a form the next person can check.

Three of the five "held because it breaks this repo" entries turned out to rest on a wrong diagnosis, so they came out:

- **postgres (#2928)** was never a pg/pg-cursor version skew. pg 8.8 added an `exports` map, and jest resolves subpaths through its own `resolve.exports` — 1.x, which cannot match subpath patterns. Node resolved `pg/lib/result.js` the whole time. Jest 29.7 (already in range) brought 2.x and the hold evaporated.
- **vscode-textmate (#2918)** deep-imported type modules that 9.3.2 doesn't export. The generator reads a TextMate grammar `.json`, so it now declares that file's shape itself — a contract that survives whatever upstream does with its internals. Generated `malloy.monarch.ts` is byte-identical.
- **vega** was half-released already: vega moved to 6, vega-lite stayed on 5 and dragged its own vega 5 along, which is where the three high advisories lived. vega-lite 6 peers on vega ^6, so there is now one vega and no advisory. No renderer code changed.

Also taken: jest 30, vite 8, peggy 5, jasmine-core 7 (which retires the `@types/jasmine` tracking ignore), vscode-oniguruma 2, every in-range minor and patch, and the two advisories a direct dep could clear (mysql2, toml). Six dependencies nothing imports are deleted.

Prod `npm audit` goes 27 → 18, high 14 → 6. Every remaining high and critical reconciles to the snowflake or databricks hold.

Four things did **not** move, and the ledger now says why in terms that can be re-tested rather than believed:

- **BigQuery** — the entry said jest could never load gaxios 7's ESM-only dynamic import. It can: name the importer *and* its target and the whole auth path runs live queries on bigquery 9. What actually blocks the bump is `projectId` — the connector reads it synchronously in a constructor and the v8/v9 SDK resolves it asynchronously. The hold waits on us, not on vitest.
- **Node** — the 24.17 keep-alive regression is fixed in 24.18.0, but the BigQuery suite on any post-24.16 Node takes an intermittent jest-worker SIGSEGV: 0 in 16 runs on 24.16.0, 2 in 8 on 24.18.0, 2 in 16 on 24.21.0, same tree. `precheck` is clean on all three. Cause not established; the measurement is. This one now costs us a security release, so it wants an owner.
- **snowflake / databricks** — boundaries re-verified rather than assumed. `snowflake-sdk@3.3.0` bundles nine `.node` binaries with an empty `optionalDependencies`; `@databricks/sql@2.1.0` uses the other shape, eight per-platform kernel packages. The ledger now carries the two commands that show this, since no metadata field does.
- **apache-arrow** — new hold. `@duckdb/duckdb-wasm` and `@motherduck/wasm-client` both require `^17.0.0`, so raising ours leaves the tree `invalid` and our typings a major ahead of the values duckdb-wasm returns, with the build green throughout.

On the open-PR limit hiding things: it does, and turning security auto-updates off is what gave it teeth. GitHub exempts security PRs from the cap — but there are none here, so the cap governs everything Dependabot produces, and full means silent. Slots go to whatever is oldest, so an un-ignored major nobody intends to take squats one forever; storybook alone held four of ten. Storybook, jsdom, lerna and gaxios get ignore rows with argued reasons, which frees the queue.

`@types/node` is left alone — #3002 already syncs it to the runtime major and adds a check so the two can't drift again.

Verified from a clean tree: `npm ci`, `npm run build`, `npm run build-duckdb-db`, `npm run precheck` (4176 passing of 4270, unchanged), `npm run lint`. The one snapshot change is jest 30 rewriting its own header URL.

https://claude.ai/code/session_01NuuXJ8g133X9wL6MBiyDUf


mtoy-googly-moogly:
Closing unmerged. The branch `dependency-sweep` stays pushed if any of it is wanted later.

What's in it: 17 dependencies moved, 6 unused ones removed, 3 holds released (pg #2928, vscode-textmate #2918, vega), prod `npm audit` 27→18 with highs 14→6, and two CI bugs jest 30 exposed — a bare `--reporters` in `ci-core` and `ci-test-sanity-check.sh` globbing spec files out of `node_modules`. CI was green across every dialect plus consumer-canary.

Not landing because the pin-ledger half of it is the wrong shape: three of five documented holds turned out to have the wrong cause recorded, so the document was costing more than it returned.

PR #3097, opened 2026-09-22
feat(sqlserver): SQL Server dialect and connection, experimental

housejester:
A native SQL Server dialect (`sqlserver`) and connection package (`@malloydata/db-sqlserver`), experimental behind `##! experimental.dialect.sqlserver`. SQL Server 2017 is the floor; the same SQL runs on 2017, 2019, 2022 and Azure SQL Database. Every query runs on the server and only result rows move.

Malloy's cross-database suite (`db-all`, 811 tests) gives **681 pass / 19 fail / 110 skip**, identical on a SQL Server 2017 and a 2022 container. All 19 failures are one class, arrays and records built inside a query, and are the open question below. The generated SQL of every other dialect is byte-identical to `main`: a 48-query set covering each emitter path was compiled on both trees for postgres, standardsql, snowflake, trino, presto, mysql, databricks and duckdb, with no differing line. `test-duckdb` 4298 / 0, `ci-postgres` 810 / 0.

## Decision points

These are the places the core changes shape, each a flag with a default so no dialect outside this repo changes behaviour. I would like a read on each before this moves out of experimental.

1. **A boolean is a condition or a value.** `booleanType: 'none'` now means the compiler tracks which form a boolean expression is in (`BooleanForm`) and converts at the boundary through four `Dialect` hooks (`sqlConditionAsValue`, `sqlValueAsCondition`, `sqlNot`, `sqlNullIsTrue`, plus `sqlBooleanValue` beside `sqlBoolean`). For a dialect with booleans every hook is the identity or the string it wrote before. The translator's `CASE WHEN c THEN 1 ELSE 0 END` wrap from #2348 is removed, since it was a second seam and read NULL as 0; `sql_boolean` marks its node as a condition.
2. **`orderByStage: 'final'` and `FinalStageOrdering`.** T-SQL refuses `ORDER BY` in a CTE without `TOP`, so the final stage receives the query's order and limit by output name and writes them itself. Postgres keeps `'last'` and its SQL is unchanged.
3. **`groupByClause: 'expression'`.** T-SQL has no ordinal GROUP BY and refuses a constant in one, so a constant dimension is left out and a stage of only constants writes `GROUP BY ()`. Whether a dimension is constant follows field references (`QueryField.readsColumn`), so one defined through another constant counts. An empty input therefore yields one row rather than none for such a query on this dialect.
4. **`limitClause: 'top'`** (the shape asked for in #2317), **`subqueryOrderByRequiresLimit`**, **`likeExtraWildcards`** (`[` is a LIKE wildcard), **`supportsConstantWindowOrder`** (T-SQL refuses a constant in a window's ORDER BY; false leaves constant terms out and orders by `(SELECT NULL)` when none remain), and a spread that wraps each argument in a prefix and suffix (`greatest`/`least` as the first ordered row of a `VALUES` list). Trino's `concat` template named a `CAST` prefix and suffix the compiler never applied; the template now says what it emits, so Trino's SQL is unchanged, and whether those arguments should be cast is a separate question.
5. **One `FOR JSON PATH` document per result row**, since the driver has no nested types. A `bigint` above 2^53 and the 17th digit of a float do not survive; the README says so.
6. **Nests as JSON text built by concatenation**, each sorted in its own `OPENJSON` subquery because T-SQL allows one `WITHIN GROUP` ordering per SELECT. A multi-stage pipeline inside a nest and `limit:` on a nested `select:` are refused at translation.
7. **`sqlDistinctKey()`** as `ROW_NUMBER() OVER (ORDER BY (SELECT NULL))`: `NEWID()` is re-evaluated per joined row and double-counts a symmetric aggregate.
8. **`supportsRegexpMatch = false`** makes `~ r'...'` a located translation error, and the regex functions carry error templates. #2317's review said `sqlRegexpMatch` should be implementable; SQL Server has no regular expressions before 2025, so on the 2017 floor there is nothing to implement. `string_agg_distinct`, `ltrim`/`rtrim` with a character set (2022) and `byte_length` (UTF-8 collations, 2019) are refused the same way.
9. **NULL sorts last** in both directions at the top level and inside nests, as on DuckDB; SQL Server's default is first.
10. **`uniqueidentifier` reads as a string** (its only Malloy-facing operations are equality, grouping and display). **`datetimeoffset` reads as a timestamp** whose clock functions run at the stored offset; this is the timestamp-context question and I would rather have the maintainers' answer than guess. `text`/`ntext`, `time`, binary, `xml` and the rest are `sql native`.
11. **Time zones** through `AT TIME ZONE` and a CLDR IANA→Windows table with every bcp47 alias. An unmapped name is still a compile-time throw rather than a located error; a `dialect.hasTimezone(name)` translator hook would fix that and is left for a follow-up.
12. **Materialized tables** go to a `scratchSchema` (`tempdb.dbo` by default), named for the query and the principal, filled whole or dropped under an application lock. `sample:` takes the first n rows: `TABLESAMPLE` returns pages.
13. **CI** is a `workflow_dispatch` workflow that starts the container and loads the parquet through Malloy's own DuckDB connection, like the Databricks precedent; wiring it into `run-tests.yaml` is the maintainers' call.

Pre-existing and deliberately untouched: `noCTE … dialect.name !== 'postgres'` in `query_model_impl.ts`.

## Open question: arrays and records built in a query

The 19 remaining failures are all one shape: an array or record constructed inside a query, as a literal (`[1, 2, 3]`, `mk_array`, `mk_record`) or by a SQL block returning `JSON_ARRAY(...)`. SQL Server has no typed array or record value before 2025's `json` type, so the value comes back as a string. The existing flags (`supportsArraysInData`, `compoundObjectInSchema`) cover compound values read from a table and already skip that part of the suite; nothing covers compound values *constructed* in a query.

How would you like this handled? The shape I would propose is a sibling flag, default true, read at two seams: the translator refuses an array or record literal and `mk_array`/`mk_record` with a located error when it is false (as `nesting-unsupported` does), and the suite gates the 19 on it. The SQL-block cases cannot be refused, since the schema says string, so they would gate only. I have left the 19 red rather than name-gate them, so the choice is visible here.

Five further cases the engine cannot run are gated in `databases/all` with the engine fact stated at the gate: `coalesce` of only the NULL constant and `nullif` with a NULL first argument (Malloy's bare null has no type to cast it with), and three `time.spec` tests that nest only literals under an ungrouped query (the sorting subquery has no outer column to correlate on). The README lists every refusal and known failure in one place.

## Connection

`mssql` 12 over tedious, pure JavaScript. SQL login, NTLM, and Microsoft Entra ID (`azure-default`, service principal, MSI, access token; unit-tested, not verified against a live Azure SQL instance). `server` accepts `host,port`; `instanceName`, `readOnlyIntent`, `multiSubnetFailover`, `hostNameInCertificate`, `applicationName`, `scratchSchema`, `requestTimeoutMs`, pool bounds, or an ADO.NET `connectionString`. The digest carries the principal and never a secret. Schema methods return an error string for a missing table or a bad SQL block.

## Testing

```
bash test/mssql/mssql_start.sh          # SQL Server 2022 on 1433; MSSQL_IMAGE / MSSQL_PORT pick another
export MSSQL_HOST=localhost MSSQL_PORT=1433 MSSQL_USER=sa MSSQL_PASSWORD=... MSSQL_DATABASE=malloytest MSSQL_TRUST_SERVER_CERTIFICATE=true
MALLOY_DATABASE=sqlserver npx jest --selectProjects db-sqlserver   # 66 / 66
npm run ci-sqlserver                                                # 681 / 19 / 110
```

`test/src/databases/sqlserver/sqlserver.spec.ts` builds the dialect up in the order a query grows (literal, table, filter, order, group, boolean, nest, zone, join) and pins each fix in this PR; `packages/malloy-db-sqlserver/src/sqlserver_connection.spec.ts` covers connectivity, the type map against the live server, error strings, the materialized table and the digest; `sqlserver_config.unit.spec.ts` covers the driver configuration without a database. The `BooleanForm` hooks and `likeExtraWildcards` have DB-free pins in `dialect.spec.ts`.


housejester:
Superseded by #3099, the small first cut we agreed on: connection, schema and result readers, the trivial half of the dialect, experimental. This branch stays as the source for the follow-up PRs (nesting via the result schema, booleans as values, CROSS APPLY dimensions, NULL ordering, datetimeoffset as timestamptz, a 2022 subclass, located refusals).

